PR #4, opened 2022-09-23
react_todo-app-add-and-delete

yurii-kozlov:
Demo link - https://yurii-kozlov.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,114 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  FormEvent,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { createTodo, getTodos, removeTodo } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,223 +1,169 @@
...
+
+    if (!title.trim()) {
+      setError(Error.ADDING);
+
+      return;
+    }
+
+    setIsAdding(true);
+
+    await createTodo(userId, title)
```
TarasHoliuk, line 74: you don't need `await` (and `async` above) here if used `then`
with `await` it should look in the following way:
```
const createdTodo = await createTodo(userId, title);
setTodos([...todos, todo]);
```
with `try-catch` block :arrow_up:

TarasHoliuk (changes requested):
Good start! Continue working and provide all other features

Bohdan-Maksymiuk (changes requested):
fix demo link) https://yurii-kozlov.github.io/react_todo-app-add-and-delete/

https://user-images.githubusercontent.com/94289038/193906701-991f7a79-4b67-41b7-a7a3-4c1d32b31cde.mov

* show the loader on the added todo

PR #9, opened 2022-09-28
Solution

oleg-zolkiewskyi:
[DEMO LINK](https://oleg-zolkiewskyi
.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -1,8 +1,22 @@
 import { Todo } from '../types/Todo';
 import { client } from '../utils/fetchClient';
 
-export const getTodos = (userId: number) => {
+export const getTodos = (userId: number | undefined) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const createTodo = (userId: number | undefined, query: string) => {
+  return client.post<Todo>(`/todos?userId=${userId}`, {
+    userId,
+    completed: false,
+    title: query,
+  });
+};
+
+export const deleteTodo = (todoID: number) => {
+  return client.delete(`/todos/${todoID}`);
+};
+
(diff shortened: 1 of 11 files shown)


src/api/todos.ts
```diff
@@ -1,8 +1,22 @@
...
+    completed: false,
+    title: query,
+  });
+};
+
+export const deleteTodo = (todoID: number) => {
+  return client.delete(`/todos/${todoID}`);
+};
+
+export const updateTodo = (todoId: number, object: unknown): Promise<Todo> => {
```
TarasHoliuk, line 20:
```suggestion
export const updateTodo = (todoId: number, data: Partial<Todo>): Promise<Todo> => {
```
please, fix this type everywhere

src/api/todos.ts
```diff
@@ -1,8 +1,22 @@
 import { Todo } from '../types/Todo';
 import { client } from '../utils/fetchClient';
 
-export const getTodos = (userId: number) => {
+export const getTodos = (userId: number | undefined) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const createTodo = (userId: number | undefined, query: string) => {
```
TarasHoliuk, line 8: There is probably no case to get `undefined` as `userId` and we'll never send a post request with `undefined`, but it's better to prevent it, even theoretically. You should check if userId exists before calling this function (if not - do nothing)

TarasHoliuk (changes requested):
Amazing :fire: But left 2 small comments, check it

Bohdan-Maksymiuk (changes requested):
Good job, bit this is not implemented
* manually add a temp todo with id: 0 after the list while waiting for the response (don't add it to the array);
* show the loader on the added todo (see the styles of the 5th todo Redux);
* also loader on deleting todo

PR #17, opened 2022-10-02
Todo-App-Add-Delete solution

NataliyaAn:
[DEMO LINK](https://NataliyaAn.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,120 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useRef, useState, useMemo,
+} from 'react';
+import classNames from 'classnames';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Error } from './components/Error';
+import { Filter } from './components/Filter';
+import { getTodos, createTodo, deleteTodo } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortFilter, setSortFilter] = useState('all');
+  const [visibelTodos, setVisibelTodos] = useState<Todo[]>([...todos]);
+  const [completedTodos, setCompletedTodos]
+  = useState<Todo[]>([...todos].filter(todo => todo.completed));
+
+  const [isError, setError] = useState(false);
+  const [messageError, setMessageError] = useState('');
+  const [newTodoTitle, setTitle] = useState('');
+  const [activeTodoId, setActiveTodoId] = useState(0);
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
+
+    if (user) {
+      getTodos(user.id)
+        .then(response => {
+          setTodos(response);
+        })
+        .catch(() => {
+          setError(true);
+          setMessageError('Todos from server were not gotten');
+        });
+    }
   }, []);
+
+  useMemo(() => {
+    setVisibelTodos(() => (
+      todos.filter(todo => {
+        switch (sortFilter) {
+          case 'active':
+            return !todo.completed;
+
+          case 'completed':
+            return todo.completed;
+
+          default:
+            return true;
+        }
+      })));
+  }, [todos, sortFilter]);
+
+  useMemo(() => {
+    setCompletedTodos(() => todos.filter(todo => todo.completed));
+  }, [todos]);
+
+  const handleSubmit = async (event: React.FormEvent) => {
+    event.preventDefault();
+    if (!user) {
+      return;
+    }
+
+    if (newTodoTitle.trim().length === 0) {
+      setError(true);
+      setMessageError('Title can\'t be empty');
+
+      return;
+    }
+
+    setVisibelTodos(state => [...state, {
+      id: 0,
+      userId: user.id,
+      title: newTodoTitle,
+      completed: false,
+    }]);
+
+    try {
+      const isAdding = await createTodo({
+        userId: user.id,
+        title: newTodoTitle,
+        completed: false,
+      });
+
+      setTodos(state => [...state, isAdding]);
+    } catch (errorFromServer) {
+      setError(true);
+      setMessageError('Unable to add a todo');
+    }
+
+    setTitle('');
+  };
+
+  const handleRemoveTodo = async (removeTodoID: number) => {
+    setActiveTodoId(removeTodoID);
+
+    try {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,223 +1,197 @@
...
+      });
+
+      setTodos(state => [...state, isAdding]);
+    } catch (errorFromServer) {
+      setError(true);
+      setMessageError('Unable to add a todo');
+    }
+
+    setTitle('');
+  };
```
VolodymyrChuyko, line 98: image

the temp todo should be removed in any case. use finally to solve this issue

src/App.tsx
```diff
@@ -1,223 +1,197 @@
...
+
+  const handleRemoveTodo = async (removeTodoID: number) => {
+    setActiveTodoId(removeTodoID);
+
+    try {
+      await deleteTodo(removeTodoID);
+
+      setTodos(state => state.filter(todo => todo.id !== removeTodoID));
```
VolodymyrChuyko, line 106:
```suggestion
      setTodos(prevTodos => prevTodos.filter(todo => todo.id !== removeTodoID));
```
A variable name should describe what is stored in it.

src/App.tsx
```diff
@@ -1,223 +1,197 @@
...
+      setError(true);
+      setMessageError('Unable to delete a todo');
+    } finally {
+      setActiveTodoId(0);
+    }
+  };
+
+  const handlerRemoveComleted = () => {
+    completedTodos.forEach((todoComleted) => handleRemoveTodo(todoComleted.id));
+  };
```
VolodymyrChuyko, line 117: on remove several todos only one todo is shown with a loader.
Save an array of ids in activeTodoId to solve this issue

src/App.tsx
```diff
@@ -1,223 +1,197 @@
...
+          <form onSubmit={handleSubmit}>
             <input
               data-cy="NewTodoField"
               type="text"
               ref={newTodoField}
               className="todoapp__new-todo"
               placeholder="What needs to be done?"
+              value={newTodoTitle}
+              onChange={() => setTitle(newTodoField.current?.value || '')}
             />
```
VolodymyrChuyko, line 145: disable the input until receiving the response from the API (use isAdding variable);
isAdding should have a boolean type

src/components/Filter.tsx
```diff
@@ -0,0 +1,62 @@
...
+        data-cy="FilterLinkAll"
+        href="#/"
+        className={classNames(
+          'filter__link',
+          { selected: sortFilter === 'all' },
+        )}
+        onClick={() => (
+          sortFilter !== 'all'
+          && setSortFilter('all')
+        )}
```
VolodymyrChuyko, line 25: It would be better to make a handler instead of using inline function

src/components/Filter.tsx
```diff
@@ -0,0 +1,62 @@
...
+        data-cy="FilterLinkActive"
+        href="#/active"
+        className={classNames(
+          'filter__link',
+          { selected: sortFilter === 'active' },
+        )}
+        onClick={() => (
+          sortFilter !== 'active'
+          && setSortFilter('active')
+        )}
```
VolodymyrChuyko, line 40: make a handler

src/components/Filter.tsx
```diff
@@ -0,0 +1,62 @@
...
+        data-cy="FilterLinkCompleted"
+        href="#/completed"
+        className={classNames(
+          'filter__link',
+          { selected: sortFilter === 'completed' },
+        )}
+        onClick={() => (
+          sortFilter !== 'completed'
+          && setSortFilter('completed')
+        )}
```
VolodymyrChuyko, line 55: make a handler

VolodymyrChuyko (changes requested):
Good job, just pay attention to comments

src/App.tsx
NataliyaAn: good point, i fixed it

src/App.tsx
NataliyaAn: good point, i fixed it

src/App.tsx
NataliyaAn: done

src/App.tsx
NataliyaAn: i add new todo with loader until get answer from server. I think it is the right behavior

src/components/Filter.tsx
NataliyaAn: done

src/components/Filter.tsx
NataliyaAn: done

src/components/Filter.tsx
NataliyaAn: done

src/App.tsx
VolodymyrChuyko: You should also disable input. This is a task requirement

src/components/Filter.tsx
VolodymyrChuyko: I mean, you'd better make something like 
```
onClick={handleClick}
```
in this case,  onClick will fire the same function (handleClick) every time.
If you use an inline function, a new function will be created every time the onClick event takes place

src/App.tsx
```diff
@@ -1,223 +1,207 @@
...
+          setMessageError('Todos from server were not gotten');
+        });
+    }
   }, []);
 
+  useEffect(() => {
+    setVisibelTodos(() => (
+      todos.filter(todo => {
+        switch (sortFilter) {
+          case 'active':
```
ArturErmolenko, line 48: Do not rely on the unknown string, make constants for this.
```
const FILTERS = {
  all: ‘all’,
  completed: ‘completed’,
  active: ‘active’,
};
```

src/App.tsx
```diff
@@ -1,223 +1,207 @@
...
+import { getTodos, createTodo, deleteTodo } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortFilter, setSortFilter] = useState('all');
```
ArturErmolenko, line 17: create enum for filter

src/api/todos.ts
```diff
@@ -5,4 +5,14 @@ export const getTodos = (userId: number) => {
...
+// eslint-disable-next-line @typescript-eslint/naming-convention
+type createTypeTodo = Omit<Todo, 'id' >;
+
+export const createTodo
+  = async ({ userId, title, completed }: createTypeTodo) => {
+    return client.post<Todo>('/todos', { userId, title, completed });
+  };
+
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
```
ArturErmolenko, line 17: add type for server result (like on post method)

ArturErmolenko (changes requested):
Almost done 🔥 
just small improvements

src/App.tsx
NataliyaAn: done

src/App.tsx
NataliyaAn: done

PR #31, opened 2022-10-04
react todo app add and delete solution

anyakrainiuk:
[DEMO LINK](https://anyakrainiuk.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,27 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
+import { TodoList } from './components/TodoList';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { ErrorMessage } from './components/ErrorMessage';
+import { Todo } from './types/Todo';
+import { getTodos, createTodos, deleteTodo } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { FilterBy } from './types/filterBy';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[] | []>([]);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [filterBy, setFilterBy] = useState('all');
+  const [title, setTitle] = useState('');
+  const [isAdding, setAdding] = useState(false);
 
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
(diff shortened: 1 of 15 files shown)


src/components/ErrorMessage.tsx
```diff
@@ -0,0 +1,40 @@
...
+        },
+      )}
+
+    >
+      <button
+        aria-label="a problem"
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => setHiddenError(prev => !prev)}
```
Mejeezz, line 35: You can make a separate function above return  `  const handleErrorMessage = () => {
    setHiddenError(prev => !prev);
  };`  and then put here only a name of the function

src/api/todos.ts
```diff
@@ -1,8 +1,20 @@
 import { Todo } from '../types/Todo';
 import { client } from '../utils/fetchClient';
 
-export const getTodos = (userId: number) => {
+export const getTodos = (userId: number | undefined) => {
```
Mejeezz, line 4: Probably we have to use `null `instead of `undefined`

src/App.tsx
```diff
@@ -14,210 +30,109 @@ export const App: React.FC = () => {
     }
   }, []);
 
+  useEffect(() => {
+    async function todosFromServer(userId: number | undefined) {
```
Mejeezz, line 34: `null` instead of `undefined`

src/App.tsx
```diff
@@ -14,210 +30,109 @@ export const App: React.FC = () => {
...
+    });
+  }, [todos, filterBy]);
+
+  const handleTodos = async (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!title) {
+      setErrorMessage('Title can`t be empty');
+
+      return;
```
Mejeezz, line 75: *Optional 
its possible to write as `return setErrorMessage('Title can`t be empty');`

src/App.tsx
```diff
@@ -14,210 +30,109 @@ export const App: React.FC = () => {
...
+
+    try {
+      const newTodo = await createTodos(user?.id, title);
+
+      setTodos([...todos, newTodo]);
+    } catch {
+      setErrorMessage('Can\'t be added');
+    }
+
+    setTitle('');
```
Mejeezz, line 88: to previous coment: if you will change it, so you need some default return, for example `return setTitle('');` and switch position with `setAdding(false);`

Mejeezz (changes requested):

Mejeezz (approved):

src/App.tsx
AndriiRut: @Mejeezz what created by setState() functions returns? => yourAnswer
``Why do @anyakrainiuk should return ${yourAnswer} here?``

src/components/Auth/TodoItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <button
+        type="button"
+        className="todo__remove"
+        data-cy="TodoDeleteButton"
+        onClick={() => removeTodo(todo?.id || -1)}
+      >
+        ×
+      </button>
+
+      {(isAdding && todo?.id === 0) && (
```
AndriiRut, line 46:
```suggestion
      {isLoading && (
```
const isLoading = isAdding && todo?.id === 0

src/components/Filter.tsx
```diff
@@ -0,0 +1,66 @@
...
+  setFilterBy: (param: string) => void;
+  filteredTodos: Todo[];
+};
+
+export const Footer: React.FC<Props> = ({
+  filterBy,
+  setFilterBy,
+  filteredTodos,
+}) => {
+  return (
```
AndriiRut, line 16:
```suggestion
}) => (
```

AndriiRut (changes requested):
fix loader on deletion

https://user-images.githubusercontent.com/99652389/194621230-23a5899d-5d4f-49c3-a34c-596ea486116b.mov

dellachaise (approved):
Good job 👍 

Pay attention to all comments and add changes according to it to the last part pf the project.

I could add title only with spaces. Fix it.
Screenshot 2022-10-10 at 15 53 09

PR #46, opened 2022-10-05
solution

oksana-kyryienko:
https://oksana-kyryienko.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,146 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  FormEvent,
+  useContext,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import { createTodo, deleteTodo, getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorNotifaction } from
+  './components/ErrorNotification/ErrorNotifaction';
+import { FilterType } from './types/FilterType';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer/Footer';
+import { TodoList } from './components/TodoList/TodoList';
+import { ErrorText } from './types/Error';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
-  const newTodoField = useRef<HTMLInputElement>(null);
+  const [title, setTitle] = useState('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<FilterType>(FilterType.All);
+  const [errorText, setErrorText] = useState<string>('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [selectedId, setSelectedId] = useState<number[]>([]);
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
-    if (newTodoField.current) {
-      newTodoField.current.focus();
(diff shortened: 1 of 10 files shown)


David-Shahmatov (approved):
🚀

src/App.tsx
```diff
@@ -1,223 +1,146 @@
...
+  './components/ErrorNotification/ErrorNotifaction';
+import { FilterType } from './types/FilterType';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer/Footer';
+import { TodoList } from './components/TodoList/TodoList';
+import { ErrorText } from './types/Error';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
dellachaise, line 21: Delete `eslint-disable`.

src/App.tsx
```diff
@@ -1,223 +1,146 @@
...
+
+  const deleteFinishedTodos = useCallback(() => {
+    setSelectedId(finishedTodos.map(({ id }) => id));
+    Promise.all(finishedTodos.map(({ id }) => removeTodo(id)))
+      .then(() => setTodos((prevTodos) => prevTodos
+        .filter(({ completed }) => !completed)))
+      .catch(() => {
+        setErrorText(ErrorText.Delete);
+        setSelectedId([]);
+      });
```
dellachaise, line 105: Better use `async await` here instead of `then`
[more details how to do it](https://betterprogramming.pub/how-to-use-async-await-with-map-in-js-5059043564e0)

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+};
+
+export const Footer: React.FC<Props> = ({
+  filterType,
+  filteredType,
+  todos,
+  deleteCompleted,
+}) => {
+  const unFinishedTodos = useMemo(() => (
```
dellachaise, line 20:
```suggestion
  const unfinishedTodos = useMemo(() => (
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+      {todosCompletedLength > 0 && (
+        <button
+          data-cy="ClearCompletedButton"
+          type="button"
+          className="todoapp__clear-completed"
+          onClick={deleteCompleted}
+        >
+          Clear completed
+        </button>
```
dellachaise, line 81: A place for this button should always be reserved.(check how it looks with no completed on working version and how on your demo link).

src/components/TodoList/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+          removeTodo(todo.id);
+        }}
+      >
+        ×
+      </button>
+      <div
+        data-cy="TodoLoader"
+        className={classNames(
+          'modal overlay',
+          { 'is-active': selectedId.includes(todo.id) },
```
dellachaise, line 50: Better create a separate variable for `selectedId.includes(todo.id)` and use it here.

src/types/Error.ts
```diff
@@ -0,0 +1,6 @@
+export enum ErrorText {
```
dellachaise, line 1: 👍

dellachaise (changes requested):
You did a good job 👍 

Just need to add some improvments.
1. Red border should be only on the chosen filter, not for all.
2. Place for `Clear complited` button should be always available.
Your app now looks this way 👇 
Screenshot 2022-10-15 at 19 54 01
Need to be this way 👇 
Screenshot 2022-10-15 at 19 54 08
3. I could add a todo only with spaces
Screenshot 2022-10-16 at 00 53 50

PR #64, opened 2022-10-08
todo add and delete

Alej4ndro1:
• [DEMO LINK](https://alej4ndro1.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,153 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  FormEvent,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,223 +1,153 @@
...
+    }
+
+    setIsAdding(true);
+
+    try {
+      const postTodo = await createTodo(title, user.id);
+
+      setTodos((prevTodos) => [...prevTodos, postTodo]);
+    } catch {
+      setErrorText('123');
```
AndriiRut, line 83: interesting code of error 🤔 
Do users have to install or uninstall something from your page so they get such an error code?

AndriiRut: or maybe they are denied because of pure permissions?
Make your error test more informative! use code only for consoles aca for other devs. and for user you have to show the reason in text message)

AndriiRut (changes requested):
https://user-images.githubusercontent.com/99652389/196919314-20ef0767-1294-40a0-b8d8-e341c93e1877.mov

PR #66, opened 2022-10-08
React Todo App Add and Delete

rboiar:
- [DEMO LINK](https://rboiar.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoList.tsx
+++ b/src/components/TodoList.tsx
@@ -1,0 +1,135 @@
+import classNames from 'classnames';
+import { useState } from 'react';
+import { Todo } from '../types/Todo';
+import { FilterBy } from './TodoFilter';
+
+type Props = {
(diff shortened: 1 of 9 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,135 @@
...
+  });
+
+  const [deletingId, setDeletingId] = useState<number | null>(null);
+
+  const handleTodoDelete = (todoId: number) => {
+    onDelete(todoId);
+    setDeletingId(todoId);
+    setClosingError(false);
+
+    setTimeout(() => setDeletingId(null), 1000);
```
AndriiRut, line 44:
```suggestion
```
what if server can't be reached?
`try {} catch {} finally {}`

AndriiRut (changes requested):

PR #79, opened 2022-11-04
Implemented new functionality for App: adding and deleting todos

anastasiia-nahaiska:
[DEMO LINK](https://anastasiia-nahaiska.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,188 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Todo } from './types/Todo';
+import { ErrorType } from './types/ErrorType';
+import { FilterBy } from './types/FilterBy';
+import { getVisibletodos } from './utils/getVisibleTodos';
+import { NewTodoField } from './components/NewTodoField';
+import { TodosCountInfo } from './components/TodosCountInfo/TodosCountInfo';
+import { ClearCompletedTodos } from './components/ClearCompletedTodos';
+import { TodosFilter } from './components/TodosFilter';
+import { TodosList } from './components/TodosList';
+import { ErrorNotification } from './components/ErrorNotification';
+import { TempTodo } from './components/TempTodo';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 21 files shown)


src/App.tsx
```diff
@@ -1,223 +1,188 @@
...
+
+      try {
+        await addTodo(newTodo);
+      } catch {
+        setError(ErrorType.ADD);
+      }
     }
+
+    setNewTodoTitle('');
+    setIsAdding(false);
```
AntonVaida, line 89: When adding a new todo, after the animation of loading to the server, the list  is gapping, maybe you should reload todo list before, you setting false in setIsAdding

AntonVaida (approved):

src/App.tsx
anastasiia-nahaiska: Good point, thanks!))

src/components/ClearCompletedTodos/ClearCompletedTodos.tsx
```diff
@@ -0,0 +1,24 @@
...
+    data-cy="ClearCompletedButton"
+    type="button"
+    className="todoapp__clear-completed"
+    style={{ width: '100px' }}
+    onClick={onDeleteAllCompleted}
+    disabled={!hasCompleted}
+  >
+    {hasCompleted
+      ? 'Clear completed'
+      : ''}
```
Denchepornyuk, line 22: it's better to not render button if `hasCompleted` is false, not only text inside the button

src/components/NewTodoField/NewTodoField.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const NewTodoField: React.FC<Props> = ({
+  onError,
+  onAddTodo,
+  isAdding,
+  hasTodos,
+  newTodoField,
+  newTodoTitle,
+  onTitleChange,
+}) => {
+  return (
```
Denchepornyuk, line 22: don't use return here, avoid return if you can

src/components/TodosCountInfo/TodosCountInfo.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[]
+};
+
+export const TodosCountInfo: React.FC<Props> = React.memo(({ todos }) => {
+  const leftTodos = todos.filter(({ completed }) => !completed);
```
Denchepornyuk, line 9: use useMemo for the leftTodos variable

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #80, opened 2022-11-04
Adding and deleting the todo is implemented

jojnajbude:
[DEMO LINK](https://jojnajbude.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,143 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+
+import { getTodos, addTodo, deleteTodo } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodosList } from './components/TodosList';
+import { FilterBy } from './types/FilterBy';
+import { Footer } from './components/Footer';
+import { TodoAdd } from './components/TodoAdd';
+import { TodoAddCard } from './components/TodoAddCard';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
+  // Add new Todo
+  const [todosList, setTodoList] = useState<Todo[]>([]);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+
+  // Errors
+  const [titleError, setTitleError] = useState(false);
+  const [addingError, setAddingError] = useState(false);
+  const [deleteError, setDeleteError] = useState(false);
+  const [updateError, setUpdateError] = useState(false);
+
+  // states
+  const [isAdding, setIsAdding] = useState(false);
+  const [deleteAll, setDeleteAll] = useState(false);
+
+  // Filter By
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.All);
+
+  const leftTodos = todosList.filter(todo => todo.completed === false).length;
+  const completedTodos = todosList.length - leftTodos;
+
+  const filteredTodos = todosList.filter(todo => {
+    switch (filterBy) {
+      case FilterBy.Active:
+        return todo.completed === false;
+      case FilterBy.Completed:
+        return todo.completed === true;
+      case FilterBy.All:
+      default:
+        return true;
+    }
+  });
+
+  const getTodosList = async () => {
+    if (user) {
+      setTodoList(await getTodos(user.id));
+    }
+  };
+
+  const handlerFormSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    setIsAdding(true);
+
+    const normilizedTodoTitle = newTodoTitle
+      .trim()
+      .split(' ')
+      .filter(words => words !== '')
+      .join(' ');
+
+    if (!newTodoTitle) {
+      setTitleError(true);
+      setTimeout(() => {
+        setTitleError(false);
+      }, 3000);
+      setNewTodoTitle('');
+      setIsAdding(false);
+
+      return;
+    }
+
+    if (user && !titleError) {
+      try {
+        const newTodo = await addTodo(user.id, normilizedTodoTitle);
+
+        setTodoList(currentList => [...currentList, newTodo]);
+      } catch {
+        setAddingError(true);
+        setTimeout(() => {
+          setAddingError(false);
+        }, 3000);
+      }
+    }
+
+    setNewTodoTitle('');
+    setIsAdding(false);
+  };
+
+  const handlerInputTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTodoTitle(event.target.value);
+    setTitleError(false);
+  };
+
+  const ClearError = () => {
+    setDeleteError(false);
+    setUpdateError(false);
+    setTitleError(false);
+  };
+
+  const deleteAllCompleted = async () => {
+    setDeleteAll(true);
+    try {
+      await Promise.all(todosList.map(todo => {
+        if (todo.completed) {
+          return deleteTodo(todo.id);
+        }
+
+        return null;
+      }));
+
+      setTodoList(currentList => currentList
+        .filter(todo => !todo.completed));
+    } catch {
+      setDeleteError(true);
+      setTimeout(() => {
+        setDeleteError(false);
+      }, 3000);
+    }
+
+    setDeleteAll(false);
+  };
+
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -26,198 +152,192 @@ export const App: React.FC = () => {
...
+              id: 0,
+              title: newTodoTitle,
+              completed: false,
+              userId: user ? user.id : 0,
+            }}
+            isLoading={isAdding}
+          />
+        )}
+
+        {/* <div data-cy="Todo" className="todo">
```
grafit-off, line 186: You can remove these comments

grafit-off (approved):
LGTM ⭐

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+import { TodoAddCard } from './components/TodoAddCard';
+
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
+  // Add new Todo
+  const [todosList, setTodoList] = useState<Todo[]>([]);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
```
OleksiiBlezniuk, line 25: should we store this state in App?

I mean it looks more like state of adding todo form

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+
+  // Errors
+  const [titleError, setTitleError] = useState(false);
+  const [addingError, setAddingError] = useState(false);
+  const [deleteError, setDeleteError] = useState(false);
+  const [updateError, setUpdateError] = useState(false);
```
OleksiiBlezniuk, line 31: instead of 4 errors state, you can create enum with error type and store object with type (and message if needs)

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+
+  // states
+  const [isAdding, setIsAdding] = useState(false);
+  const [deleteAll, setDeleteAll] = useState(false);
+
+  // Filter By
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.All);
+
+  const leftTodos = todosList.filter(todo => todo.completed === false).length;
```
OleksiiBlezniuk, line 40: let's use useMemo hook for this action

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+  const completedTodos = todosList.length - leftTodos;
+
+  const filteredTodos = todosList.filter(todo => {
```
OleksiiBlezniuk, line 43: useMemo

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+      default:
+        return true;
+    }
+  });
+
+  const getTodosList = async () => {
+    if (user) {
+      setTodoList(await getTodos(user.id));
+    }
+  };
```
OleksiiBlezniuk, line 59: 1 - use useCallback for functions
2 - what if there will be an error?

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+
+  const handlerFormSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
```
OleksiiBlezniuk, line 61: useCallback

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+    event.preventDefault();
+
+    setIsAdding(true);
```
OleksiiBlezniuk, line 64: shouldn't it be a `loading` state? do you really need two states for adding and deleting loadings?

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+
+    const normilizedTodoTitle = newTodoTitle
+      .trim()
+      .split(' ')
+      .filter(words => words !== '')
+      .join(' ');
```
OleksiiBlezniuk, line 70: Would be nice to move this function somewhere outside component (better to `helpers.ts` file)

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+  const handlerInputTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTodoTitle(event.target.value);
+    setTitleError(false);
+  };
+
+  const ClearError = () => {
+    setDeleteError(false);
+    setUpdateError(false);
+    setTitleError(false);
+  };
```
OleksiiBlezniuk, line 109:
```suggestion
  const clearError = () => {
    setDeleteError(false);
    setUpdateError(false);
    setTitleError(false);
  };
```

OleksiiBlezniuk: useCallback

src/App.tsx
```diff
@@ -1,17 +1,143 @@
...
+      await Promise.all(todosList.map(todo => {
+        if (todo.completed) {
+          return deleteTodo(todo.id);
+        }
+
+        return null;
+      }));
+
+      setTodoList(currentList => currentList
+        .filter(todo => !todo.completed));
```
OleksiiBlezniuk, line 123: useCallback

src/App.tsx
```diff
@@ -26,198 +152,70 @@ export const App: React.FC = () => {
...
+
+        {todosList.length > 0
+          && (
+            <Footer
+              leftTodos={leftTodos}
+              filterBy={filterBy}
+              setFilterBy={setFilterBy}
+              completedTodos={completedTodos}
+              deleteAllCompleted={deleteAllCompleted}
+            />
```
OleksiiBlezniuk, line 194: It looks more like some kind of panel with filters then Footer

src/App.tsx
```diff
@@ -26,198 +152,70 @@ export const App: React.FC = () => {
...
+          )}
+      </div>
 
+      {(titleError || deleteError || updateError || addingError) && (
+        <div
+          data-cy="ErrorNotification"
+          className="notification is-danger is-light has-text-weight-normal"
+        >
```
OleksiiBlezniuk, line 202: Let's move this error message to separate component

src/App.tsx
```diff
@@ -26,198 +152,70 @@ export const App: React.FC = () => {
...
-        <br />
-        Unable to update a todo
-      </div>
+          {titleError && ('Title can\'t be empty')}
+
+          {addingError && <br /> && ('Unable to add a todo')}
+
+          {deleteError && <br /> && 'Unable to delete a todo'}
+
+          {updateError && <br /> && 'Unable to update a todo'}
```
OleksiiBlezniuk, line 216: I think better to create object with error data with error type and message

Then we'll be able to show message here instead of conditionally show this text here

src/components/Filter.tsx
```diff
@@ -0,0 +1,58 @@
...
+        className={classNames(
+          'filter__link',
+          {
+            selected: filterBy === FilterBy.All,
+          },
+        )}
+        onClick={() => setFilterBy(FilterBy.All)}
+      >
+        All
+      </a>
```
OleksiiBlezniuk, line 28: Or you could make array from FilterBy values and map it here 🙂

src/components/Footer.tsx
```diff
@@ -0,0 +1,44 @@
...
+        setFilterBy={setFilterBy}
+      />
+
+      <button
+        data-cy="ClearCompletedButton"
+        type="button"
+        className="todoapp__clear-completed"
+        style={{ opacity: completedTodos === 0 ? 0 : 100 }}
+        disabled={completedTodos === 0}
+        onClick={() => deleteAllCompleted()}
```
OleksiiBlezniuk, line 38:
```suggestion
        onClick={deleteAllCompleted}
```

src/App.tsx
OleksiiBlezniuk: And let's handle form submit inside form component
Just create here callback for todo adding and validate/normalize form values inside form, no need to write all logic inside App component

src/components/TodoAddCard.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todo: Todo,
+  isLoading: boolean,
+};
+
+export const TodoAddCard: React.FC<Props> = ({
```
OleksiiBlezniuk, line 10: Is this a component for mocked todo to show it while it's adding?

If yes, than why can't we use usual TodoCard for this purpose? Or create component with mocked data and don't pass here mocked todo

src/App.tsx
```diff
@@ -1,17 +1,115 @@
...
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+
+  const [todosList, setTodoList] = useState<Todo[]>([]);
+
+  // Errors
+  const [errors, setErrors] = useState<ErrorsType[]>([]);
```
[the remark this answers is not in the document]
jojnajbude: Don't agree about object with error state, 'cause on my mind it possible to have many error at one time, and it's give more visibility for user, and logic with array of errors not too complexity.

OleksiiBlezniuk (approved):

PR #81, opened 2022-11-05
Solution

MihailKostov:
[DEMO LINK](https://MihailKostov.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,191 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+  FormEvent,
+  useCallback,
+  useMemo,
+} from 'react';
+
+import {
+  postTodo,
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,223 +1,191 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
vlad-demchuk, line 1: do you need it?

src/App.tsx
```diff
@@ -1,223 +1,191 @@
...
+    } catch (error) {
+      setIsError(ErrorMessage.delete);
+    } finally {
+      setProcessing(current => current.filter(id => id !== todoId));
+    }
+  }, []);
+
+  const clearCompleted = useCallback(() => todos.forEach(todo => {
+    if (todo.completed) {
+      handleDelete(todo.id);
```
vlad-demchuk, line 144: i'd better use PromiseAll for deleting instead of iterating through the whole array and doing a request each time. But it's okay now.

vlad-demchuk (approved):

PR #98, opened 2022-11-08
 react_todo-app-add-and-delete

xiimiik:
[DEMO LINK](https://xiimiik.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,6 +12,7 @@ import { AuthContext } from './components/Auth/AuthContext';
 import { ErrorNotification } from './components/ErrorNotification';
 import { Footer } from './components/Footer';
 import { Header } from './components/Header';
+import { Loader } from './components/Loader';
 import { TodoList } from './components/TodoList';
 import { FilterType } from './types/FilterType';
 import { Todo } from './types/Todo';
@@ -140,25 +141,27 @@ export const App: React.FC = () => {
           isAdding={isAdding}
         />
 
-        {todos.length > 0 && (
-          <>
-            <TodoList
-              todos={filtredTodos}
-              removeTodo={removeTodoFromServer}
-              isAdding={isAdding}
-              tempTodo={tempTodo}
-              todoIdsToRemove={todoIdsToRemove}
-            />
+        {todos.length > 0 ? (
+          (
+            <>
+              <TodoList
+                todos={filtredTodos}
+                removeTodo={removeTodoFromServer}
+                isAdding={isAdding}
+                tempTodo={tempTodo}
+                todoIdsToRemove={todoIdsToRemove}
+              />
 
-            <Footer
-              filterType={filterType}
-              setFilterType={setFilterType}
-              todos={filtredTodos}
-              completedTodos={completedTodos.length}
-              onRemove={removeAllCompletedTodos}
-            />
-          </>
-        )}
+              <Footer
+                filterType={filterType}
+                setFilterType={setFilterType}
+                todos={filtredTodos}
+                completedTodos={completedTodos.length}
+                onRemove={removeAllCompletedTodos}
+              />
+            </>
+          )
+        ) : <Loader />}
       </div>
 
       <ErrorNotification
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,100 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
+import { getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList';
+import { FilterType } from './types/FilterType';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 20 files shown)


vladyslav-harkusha (approved):
GJ

src/App.tsx
```diff
@@ -1,223 +1,175 @@
...
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [todoIdsToRemove, setTodoIdsToRemove] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo>({
+    id: 0,
+    userId: 0,
+    title: '',
+    completed: false,
+  });
```
OleksiiBlezniuk, line 35: Do you really need to save it in state? You already have user and title in state, this state can be removed

src/App.tsx
```diff
@@ -1,223 +1,175 @@
...
+      if (user) {
+        setIsAdding(true);
+
+        const currTodo = {
+          title,
+          userId: user.id,
+          completed: false,
+        };
+
+        setTempTodo(prev => ({ ...prev, currTodo }));
```
OleksiiBlezniuk, line 61: can we handle this in list component? I mean we can pass needed values to list and we have `isAdding` state to define should we render this temporary todo or not

src/App.tsx
```diff
@@ -1,223 +1,175 @@
...
+
+        await addTodo(currTodo);
+        await getTodosFromServer();
```
OleksiiBlezniuk, line 64: No need to send this request, you can get new todo from `addTodo` method, and then add it to the list, why to send additional request every time you adding new todo

src/App.tsx
```diff
@@ -1,223 +1,175 @@
...
+    }
+  }, []);
+
+  const removeTodoFromServer = useCallback(async (todoId: number) => {
+    try {
+      if (user) {
+        setTodoIdsToRemove(currIds => [...currIds, todoId]);
+
+        await deleteTodo(todoId);
+        await getTodosFromServer();
```
OleksiiBlezniuk, line 79: same here, if todo removed successfully, just filter existing todos, why to sent additional request?

src/App.tsx
```diff
@@ -1,223 +1,175 @@
...
+
+  const completedTodos = useMemo(() => (
+    todos.filter(({ completed }) => completed)
+  ), [todos]);
+
+  const removeAllCompletedTodos = useCallback(async () => {
+    try {
+      await Promise.all(completedTodos.map(({ id }) => (
+        removeTodoFromServer(id)
+      )));
```
OleksiiBlezniuk, line 95:
```suggestion
      await Promise.all(completedTodos.map(async ({ id }) => (
        removeTodoFromServer(id)
      )));
```

src/App.tsx
```diff
@@ -1,223 +1,175 @@
...
+          )
+        ) : <Loader />}
       </div>
 
-      <div
-        data-cy="ErrorNotification"
-        className="notification is-danger is-light has-text-weight-normal"
+      <ErrorNotification
+        hasError={hasError}
+        onClose={() => setHasError(false)}
```
OleksiiBlezniuk, line 169: better to save this function to constant using useCallback, now this function is created on every render and trigger ErrorNotification component render

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="todosCounter">
+        {`${uncompletedCount} items left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        {filterTypeList.map(status => (
+          <a
+            data-cy="FilterLinkAll"
+            href="#/"
```
OleksiiBlezniuk, line 36: href and data-cy should be different for every status

OleksiiBlezniuk, line 36: also you may save this values to object and read it by key

```
const filterHrefByType = {
  [FilterType.All]: '#/',
  // other values
};

/* code */

href={filterHrefByType[status]}
```

OleksiiBlezniuk: But still it's not critical, but you should have proper attribute values for this links

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,60 @@
...
+            key={status}
+            className={classNames('filter__link', {
+              selected: filterType === status,
+            })}
+            onClick={() => setFilterType(status)}
```
OleksiiBlezniuk, line 41: useCallback

OleksiiBlezniuk (changes requested):
Something wrong with temporary todo, I can't see its title 


https://user-images.githubusercontent.com/84275496/200793184-6a5ff909-8b12-4ec0-b7fc-2f442845c2a9.mov

src/App.tsx
```diff
@@ -1,223 +1,186 @@
...
+    title: '',
+    completed: false,
+  });
+
+  const getTodosFromServer = useCallback(async () => {
+    try {
+      if (user) {
+        const todosFromServer = await getTodos(user.id);
+
+        setTodos(todosFromServer);
```
OleksiiBlezniuk, line 42: well, I was wrong about additional request in this case, better to make new request after adding/deleting/updating todos because in this case we don't have any need to save resources but we need to have one source of truth (which is database on server)

sorry for that 🥲

OleksiiBlezniuk: please return it back in next task part

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+  onRemove,
+  todosLength,
+}) => {
+  const uncompletedCount = todosLength - completedTodos;
+
+  const filterHrefByType = useMemo(() => ({
+    [FilterType.All]: '#/',
+    [FilterType.Active]: '#/active',
+    [FilterType.Completed]: '#/completed',
+  }), []);
```
OleksiiBlezniuk, line 26: this object shouldn't be in component, move it to constants file

OleksiiBlezniuk (approved):

PR #99, opened 2022-11-09
Develop

Oleksandra-Shevchenko:
[DEMO LINK](https://Oleksandra-Shevchenko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -56,13 +56,13 @@ export const App: React.FC = () => {
     }
   });
 
-  const addNewTodo = async (title: string) => {
+  const addNewTodo = async (newTitle: string) => {
     if (user) {
       setIsAdded(true);
 
       const newTodo = {
         userId: user.id,
-        title,
+        title: newTitle,
         completed: false,
       };
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -67,14 +67,13 @@ export const App: React.FC = () => {
       };
 
       try {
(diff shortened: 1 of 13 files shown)


iryna-borniak (approved):
LGTM ⭐

src/App.tsx
```diff
@@ -1,223 +1,142 @@
...
+    }
+  };
+
+  const deleteSingleTodo = async (todoId: number) => {
+    if (user) {
+      try {
+        await deleteTodos(todoId);
+        setTodos(currentTodos => (
+          currentTodos.filter(todo => todo.id !== todoId)
+        ));
```
Anna-Tkachenko, line 99: let's call here the method for getting available todos from server and will set them in state instead of filter them manually

src/components/Errors.tsx
```diff
@@ -0,0 +1,44 @@
...
+    {errors === ErrorType.DELETE && (
+      'Unable to delete a todo'
+    )}
+    <br />
+    {errors === ErrorType.UPDATE && (
+      'Unable to update a todo'
+    )}
+    {errors === ErrorType.EMPTYTITLE && (
+      'Title can not be empty'
+    )}
```
Anna-Tkachenko, line 42: will be better to define error message before the return statement depending on error type via switch / case

Anna-Tkachenko (approved):

PR #109, opened 2022-11-09
Develop

pelahiia:
[DEMO LINK](https://pelahiia.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,156 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useRef, useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorMessage } from './components/ErrorMessage';
+import { TodoList } from './components/TodoList';
+import { FilterTypes } from './types/FilterTypes';
+import { Footer } from './components/Footer';
+import { TodoField } from './components/TodoField';
+
+import { getTodos, newTodo, deleteTodo } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filterType, setFilterType] = useState<FilterTypes>(FilterTypes.all);
+  const [isAdding, setIsAdding] = useState(false);
+  const [deletedTodo, setDeletedTodo] = useState(0);
+  const [deletedTodos, setDeletedTodos] = useState<number[]>([]);
+  const [todo, setTodo] = useState('');
+
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
+  const filteredTodos = todos.filter(todo => {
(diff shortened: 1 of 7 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,55 @@
...
+            className={classNames('todo', {
+              completed: todo.completed,
+            })}
+          >
+            <label className="todo__status-label">
+              <input
+                data-cy="TodoStatus"
+                type="checkbox"
+                className="todo__status"
+                defaultChecked={todo.completed}
```
vlad-demchuk, line 27: use destructuring instead of repeating `todo.`

vlad-demchuk (changes requested):
i can't add a todo after an error
https://user-images.githubusercontent.com/102411130/201296005-7ab6c372-b054-452e-9246-7e6963da97a6.mp4

here is always should be active todos
https://user-images.githubusercontent.com/102411130/201296297-f6cb039e-82bc-4b53-836a-c812fb220e1d.mp4

add loader while todo is deleting

src/App.tsx
```diff
@@ -1,223 +1,159 @@
...
     }
+
+    const loadTodos = async () => {
+      try {
+        const loadedTodos = await getTodos(user?.id || 0);
+
+        setTodos(loadedTodos);
+      } catch {
+        setError('load');
+        setTimeout(() => setError(''), 3000);
```
TarasHoliuk, line 52: To avoid doing it by hand every time, let's `useEffect` work instead of us. You can write useEffect with `error` in the dependency array. And check: if the error is not empty, set it empty after 3000 ms as you do.

src/App.tsx
```diff
@@ -1,223 +1,159 @@
...
+
+import { getTodos, newTodo, deleteTodo } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filterType, setFilterType] = useState<FilterTypes>(FilterTypes.all);
+  const [isAdding, setIsAdding] = useState(false);
+  const [deletedIds, setdeletedIds] = useState<number[]>([]);
```
TarasHoliuk, line 20:
```suggestion
  const [deletedIds, setDeletedIds] = useState<number[]>([]);
```

src/types/ErrorType.ts
```diff
@@ -0,0 +1,7 @@
+export enum ErrorTypes {
+  LOAD = 'load',
+  ADD = 'add',
+  LENGTH = 'length',
+  DELETE = 'delete',
+  DELETEALL = 'deleteAll',
```
vlad-demchuk, line 6:
```suggestion
  DELETE_ALL = 'deleteAll',
```

src/App.tsx
```diff
@@ -1,223 +1,164 @@
...
-              data-cy="NewTodoField"
-              type="text"
-              ref={newTodoField}
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
+        <TodoField
+          todos={todos}
+          newTodo={newTodoField}
+          isAdding={isAdding}
+          addNewTodo={() => addNewTodo()}
```
vlad-demchuk, line 135: you don't have arguments, so it's not necessary to  call it inside another function

vlad-demchuk (approved):

PR #127, opened 2022-11-14
create second part of todo task

AndriyZhabynets:
[DEMO LINK](https://AndriyZhabynets.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoList/Todo.tsx
+++ b/src/components/TodoList/Todo.tsx
@@ -1,0 +1,57 @@
+import classnames from 'classnames';
+import React from 'react';
+import { Loader } from '../Loader/Loader';
+import { Todo } from '../../types/Todo';
+
+type Props = {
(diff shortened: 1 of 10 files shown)


src/components/TodoList/Todo.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  isAdding: boolean,
+  todo: Todo,
+  deleteTodo:(param: number) => void,
+  loadingTodoIds: number[],
+};
+
+export const Todos: React.FC<Props> = ({
```
glugovsky, line 13: Would be better to name the component **TodoItem**, because it is a single todo and to avoid confusing it with the type. Here, in the file name and in TodoList

glugovsky (approved):
👍

PR #131, opened 2022-11-15
Solution

AMahalias:
[DEMO LINK](https://AMahalias.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,13 +45,13 @@ Remove all the completed todos after `Clear completed` button click:
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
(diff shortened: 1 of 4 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export const Header: React.FC<Props> = ({
+  inputValue,
+  newTodoField,
+  isExist,
+  inputDisabled,
+  handleInputChange,
+  handleAddTodo,
+}) => {
```
qoa11a, line 20: I don't see where you use this component. Please add and commit all files.

qoa11a (changes requested):

src/components/Header/Header.tsx
AMahalias: Excuse me, I don't understand this comment, because there is this component in app.tsx and application can't work without it. Or is it something wrong?

Anastasiia-Svintsova (changes requested):
- I shouldn't be able to add an empty todo
Animation

polosanya (approved):
Good job!

PR #134, opened 2022-11-17
part2

zhenija:
[DEMO LINK](https://zhenija.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,160 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useCallback,
+  useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import { getTodos, addTodo, removeTodo } from './api/todos';
+import { Errors } from './components/Errors';
+import { Footer } from './components/Footer';
+import { FilterMethods } from './types/FilterMethods';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,223 +1,160 @@
...
+
+      setTodos(todosFromServer);
+    } catch (Error) {
+      setErrorMessage('Loading error!');
+
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, []);
```
qoa11a, line 43: There's no need in useCallback as you don't pass it to child components or in any deps array.

https://beta.reactjs.org/apis/react/useCallback

src/App.tsx
```diff
@@ -1,223 +1,160 @@
...
+      } catch (error) {
+        setErrorMessage('Unable to add a todo');
+        setTitle('');
+        setIsAdding(false);
+
+        setTimeout(() => {
+          setErrorMessage('');
+        }, 3000);
+      }
+    }
```
qoa11a, line 91: Let's avoid unnecessary nesting.
```suggestion
      if (!user) return;
      try {
        setIsAdding(true);

        const todoToAdd = {
          title: newTitle,
          userId: user.id,
          completed: false,
        };

        const newTodo = await addTodo(todoToAdd);

        setIsAdding(false);

        setTodos(currentTodos => [...currentTodos, newTodo]);
      } catch (error) {
        setErrorMessage('Unable to add a todo');
        setTitle('');
        setIsAdding(false);

        setTimeout(() => {
          setErrorMessage('');
        }, 3000);
      }
    
```

qoa11a (changes requested):
https://www.loom.com/share/1b6e74978e2a47c983be9088523b5288
I can't unfocus from the input. Let's fix it.

polosanya (approved):
Good job!

PR #136, opened 2022-11-21
solution

bohdan-tron:
[DEMO LINK](https://bohdan-tron.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -66,7 +66,7 @@ export const App: React.FC = () => {
     }
 
     fetchTodos();
-  }, [todos]);
+  }, []);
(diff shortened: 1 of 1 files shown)


qoa11a (changes requested):
https://www.loom.com/share/e0f9dc2b25d94a568fe841f4563f3aed

I can't unfocus from input. Let's work on it. Also experience seems very laggy. First todo is added, than it disappears and adds again. And my input isn't cleared until it's added second time.

liliia-livandovskaya (approved):
Looks good to me 👍

PR #156, opened 2022-12-14
Solution with styles

taraselnik:
[DEMO LINK](https://taraselnik.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,102 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext, useEffect, useRef, useState,
+} from 'react';
+import { getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Errors } from './types/Errors';
+import { Filter } from './types/Filter';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isError, setIsError] = useState(true);
+  const [selectedFilter, setSelectedFilter] = useState<string>(Filter.ALL);
+  const [typeError, setTypeError] = useState<string>('');
+  const [newTitleTodo, setNewTitleTodo] = useState('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [isDeletedComplete, setisDeletedComplete]
+  = useState<boolean>(false);
+
+  const filteredTodos = [...todos];
+
+  const fetchData = useCallback(async () => {
+    if (user) {
+      try {
+        const inData = await getTodos(user.id);
+
+        setTodos(inData);
+      } catch (inError) {
+        setIsError(false);
+        setTypeError(Errors.ErrGET);
+      }
+    }
+  }, []);
 
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
-  }, []);
+  });
+
+  useEffect(() => {
+    fetchData();
+  }, [selectedFilter]);
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          <button
-            data-cy="ToggleAllButton"
-            type="button"
-            className="todoapp__toggle-all active"
-          />
-
-          <form>
-            <input
-              data-cy="NewTodoField"
-              type="text"
-              ref={newTodoField}
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
-            />
-          </form>
-        </header>
-
-        <section className="todoapp__main" data-cy="TodoList">
-          <div data-cy="Todo" className="todo completed">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-                defaultChecked
-              />
-            </label>
-
-            <span data-cy="TodoTitle" className="todo__title">HTML</span>
-            <button
-              type="button"
-              className="todo__remove"
-              data-cy="TodoDeleteButton"
-            >
-              ×
-            </button>
-
-            <div data-cy="TodoLoader" className="modal overlay">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-
-          <div data-cy="Todo" className="todo">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
-
-            <span data-cy="TodoTitle" className="todo__title">CSS</span>
-
-            <button
-              type="button"
-              className="todo__remove"
-              data-cy="TodoDeleteButton"
-            >
-              ×
-            </button>
-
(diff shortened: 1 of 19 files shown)


src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,53 @@
...
+  typeError,
+}) => {
+  useEffect(() => {
+    onSetIsError(isError);
+    if (isError) {
+      setTimeout(() => {
+        onSetIsError(true);
+      }, 3000);
+    }
+  }, [isError]);
```
iryna-lelyk, line 23: Your errror don't disappear in 3 seconds. Maybe you need to clearTimeout  if it was set.
https://prnt.sc/49poofCDtBVt

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,53 @@
...
+        && <>{Errors.ErrBlankTitle}</>}
+      <br />
+      {typeError === `${Errors.ErrADD}`
+        && <>{Errors.ErrADD}</>}
+      <br />
+      {typeError === `${Errors.ErrDEL}`
+        && <>{Errors.ErrDEL}</>}
+      <br />
+      {typeError === `${Errors.ErrUPD}`
+        && <>{Errors.ErrUPD}</>}
```
iryna-lelyk, line 50: Here you have very complicated logic. Maybe you can paste only your `typeError` with neccessary message. Ad show it only if there is an error
```suggestion
     {!isError && typeError}
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,121 @@
...
+    e.preventDefault();
+    onSetFilterGlobal(title);
+  };
+
+  const compTodos = todos.filter(todo => todo.completed);
+
+  console.log(compTodos);
+  const clearCompletedTodos = () => {
+    compTodos.forEach(async (a) => {
+      console.log(a);
```
iryna-lelyk, line 44: Remove console.logs from your final commit

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,121 @@
...
+        await removeTodo(a.id);
+      } catch (inError) {
+        console.log('ERROR DELETE', inError);
+        onSetIsError(false);
+        onSetTypeError(Errors.ErrDEL);
+      }
+
+      onSetisDeletedComplete(false);
+      onSetFilterGlobal(Filter.ALL);
+      toLoad();
```
iryna-lelyk, line 56: Better to move in your try construction, after removeTodo, because if you don't edlete completed todos, you don't need to reload data from server

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,121 @@
...
+      try {
+        onSetisDeletedComplete(true);
+        await removeTodo(a.id);
+      } catch (inError) {
+        console.log('ERROR DELETE', inError);
+        onSetIsError(false);
+        onSetTypeError(Errors.ErrDEL);
+      }
+
+      onSetisDeletedComplete(false);
```
iryna-lelyk, line 54: And this better to move in finally construction. After your request removeTodo will be resolved

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,121 @@
...
+  ) => {
+    e.preventDefault();
+    onSetFilterGlobal(title);
+  };
+
+  const compTodos = todos.filter(todo => todo.completed);
+
+  console.log(compTodos);
+  const clearCompletedTodos = () => {
+    compTodos.forEach(async (a) => {
```
iryna-lelyk, line 43: What does `a` mean? Make your variables clearer

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+        ×
+      </button>
+
+      <div
+        data-cy="TodoLoader"
+        className={classNames(
+          'modal overlay',
+          {
+            'is-active': isAdding
+            || isDeleted || (isDeletedComplete && todo.completed),
```
iryna-lelyk, line 73: Maybe it will be easier if you have had 1 variable, for example `isLoading` and you will be able to use it when you add or delete todo

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+        {/* <div data-cy="Todo" className="todo completed">
+        <label className="todo__status-label">
+          <input
+            data-cy="TodoStatus"
+            type="checkbox"
+            className="todo__status"
+            defaultChecked
+          />
+        </label>
```
iryna-lelyk, line 93: Remove unnecessary comments from your final commit

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,121 @@
...
+      }
+
+      onSetisDeletedComplete(false);
+      onSetFilterGlobal(Filter.ALL);
+      toLoad();
+    });
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
```
iryna-lelyk, line 61: Maybe it will be better to separate components in different way. For example, you don't need to remove all foter from App.tsx. You could make only separate component for **Filter** (this nav tag). And you will pass less props into this component. Also you don't need to move all header in separarte component. You can move only **form for creating new todo**. And this also will help you to don't pass unncessary props. 

And optional recomendation. You can **filter your todos in App.tsx**. And then pass in TodoList already filtered todos. And in your Filter component pass filter status and method to change it.

iryna-lelyk (approved):
You have done hard work. Just add a couple corrections)

src/App.tsx
```diff
@@ -1,223 +1,102 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext, useEffect, useRef, useState,
```
mykola-koval, line 4: you promised to fix this
```suggestion
  useCallback,
  useContext, 
  useEffect, 
  useRef, 
  useState,
```

mykola-koval (changes requested):

src/components/ErrorNotification/ErrorNotification.tsx
taraselnik: hahh)) you are right

src/components/Footer/Footer.tsx
taraselnik: wil doo ofter all reviews )

src/components/Footer/Footer.tsx
taraselnik: it's good mark! but not in try cause await is async and flow might go through await and any way i will. maybe in catch?? i'll try

src/components/Footer/Footer.tsx
taraselnik: i tried but did't see difference

src/components/Footer/Footer.tsx
taraselnik: okay)

src/components/TodoItem/TodoItem.tsx
taraselnik: nope for now. i have refactored a bit in 3 part

src/components/TodoList/TodoList.tsx
taraselnik: of course! thanks a lot

src/components/Footer/Footer.tsx
taraselnik: i thought a lot about, and i'm sure it better to keep App as clear as i can. and i tried to imagine that every component have to do separate programer and they need to agree comunication beetwen. create a base structural plan of the APP and after move on. on tech side for this job is not enough context i know (i didn't use it) and maybe something else to imrove working with states.

src/App.tsx
taraselnik: ok. i will

taraselnik:
I' ve fixed a lot in 3d part. amount of components and managing are a discussion. i think it could be a better way to avoid regular renderings. the hottest part is group deleting. Ill try to solve it in future

denys-danyliuk (approved):

PR #179, opened 2022-12-30
Solution

Ivanenko1402:
[DEMO LINK](https://Ivanenko1402.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Filter/Filter.tsx
+++ b/src/components/Filter/Filter.tsx
@@ -18,7 +18,7 @@ export const Filter: React.FC<Props> = ({
         href="#/"
         className={
           classNames('filter__link', {
-            selected: filterBy === 'all',
+            selected: filterBy === Filters.All,
           })
         }
         onClick={() => setFilterBy(Filters.All)}
@@ -31,7 +31,7 @@ export const Filter: React.FC<Props> = ({
         href="#/active"
         className={
           classNames('filter__link', {
-            selected: filterBy === 'active',
+            selected: filterBy === Filters.Active,
           })
         }
         onClick={() => setFilterBy(Filters.Active)}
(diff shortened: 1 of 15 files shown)


src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,55 @@
...
+  setFilterBy,
+}) => {
+  return (
+    <nav className="filter" data-cy="Filter">
+      <a
+        data-cy="FilterLinkAll"
+        href="#/"
+        className={
+          classNames('filter__link', {
+            selected: filterBy === 'all',
```
Anastasiia-Svintsova, line 21: Seems that you have an enum for `all` value, so use it

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,55 @@
...
+      >
+        All
+      </a>
+
+      <a
+        data-cy="FilterLinkActive"
+        href="#/active"
+        className={
+          classNames('filter__link', {
+            selected: filterBy === 'active',
```
Anastasiia-Svintsova, line 34: Same here

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,55 @@
...
+        onClick={() => setFilterBy(Filters.Active)}
+      >
+        Active
+      </a>
+      <a
+        data-cy="FilterLinkCompleted"
+        href="#/completed"
+        className={
+          classNames('filter__link', {
+            selected: filterBy === 'completed',
```
Anastasiia-Svintsova, line 46: And here

Anastasiia-Svintsova (changes requested):
- The loaded todo should disappear after the error. And a user should be able to add a new todo after this.

https://user-images.githubusercontent.com/99403198/210066345-d15d4f65-1d9d-4ed8-944d-7ecb8c795d18.mov

- Same for removing

https://user-images.githubusercontent.com/99403198/210066397-0df0318d-63a4-47d4-bd47-6b6353d14412.mov

- Before adding the todo, check if the field is not empty

https://user-images.githubusercontent.com/99403198/210066700-e2d78ada-5bbd-4b8d-801c-b6438d9d33c5.mov

Anastasiia-Svintsova (approved):

PR #190, opened 2023-01-10
Done

p-omogit-e:
[DEMO LINK](https://p-omogit-e.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,18 +11,26 @@ import { FilterList } from './components/FilterList';
 import { getTodos, pushTodo } from './api/todos';
 import { Todo } from './types/Todo';
 import { ErrorNotification } from './components/ErrorNotification';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
   const [todoList, setTodoList] = useState<Todo[]>([]);
-  const [emptyFieldError, setEmptyError] = useState(false);
+  const [visibleTodoList, setVisibleList] = useState<Todo[]>([]);
   const [newTitle, setNewTitle] = useState('');
+
+  const [isAdding, setIsAdding] = useState(false);
+
+  const [emptyFieldError, setEmptyError] = useState(false);
+  const [failedAddError, setAddError] = useState(false);
+  const [failedDeleteError, setDeleteError] = useState(false);
 
   const findTodos = async () => {
     const loadedTodos = await getTodos(user?.id) || null;
 
     setTodoList(loadedTodos);
+    setVisibleList(loadedTodos);
   };
 
   useEffect(() => {
@@ -33,8 +41,9 @@ export const App: React.FC = () => {
     findTodos();
   }, []);
 
-  const cancelEmptyError = () => {
+  const cancelErrors = () => {
     setEmptyError(false);
+    setAddError(false);
   };
 
   const findNewTodoId = (): number => {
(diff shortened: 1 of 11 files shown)


src/api/todos.ts
```diff
@@ -1,8 +1,24 @@
...
-export const getTodos = (userId: number) => {
+export const getTodos = (userId: number | undefined) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const pushTodo = (
+  id: number,
+  title: string,
+  userId: number | undefined,
```
tetiana-ivanchenko, line 11:
```suggestion
  userId?: number,
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[];
+  onDelete: (id: number) => void;
+};
+
+export const TodoList: React.FC<Props> = ({ todos, onDelete }) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
tetiana-ivanchenko, line 13: remove eslint-disable

tetiana-ivanchenko (changes requested):
add demo link

src/App.tsx
```diff
@@ -1,19 +1,134 @@
...
+  const [visibleTodoList, setVisibleList] = useState<Todo[]>([]);
+  const [newTitle, setNewTitle] = useState('');
+
+  const [isAdding, setIsAdding] = useState(false);
+
+  const [emptyFieldError, setEmptyError] = useState(false);
+  const [failedAddError, setAddError] = useState(false);
+  const [failedDeleteError, setDeleteError] = useState(false);
+
+  const findTodos = async () => {
```
vlad-demchuk, line 29:
```suggestion
  const fetchTodos = async () => {
```

vlad-demchuk: it will be better than `find`

src/App.tsx
```diff
@@ -1,19 +1,134 @@
...
+    const loadedTodos = await getTodos(user?.id) || null;
```
vlad-demchuk, line 30: loadedTodos shouldn't be `null`. use try/catch construction to catch errors

src/App.tsx
```diff
@@ -1,19 +1,134 @@
...
+      }
+
+      return sum;
+    });
+
+    return sum;
+  };
+
+  const filterTodos = async (filterBy: Filter) => {
+    setVisibleList(todoList);
```
vlad-demchuk, line 98: you don't need `setVisibleList` here. if you do it in cases it will be the same. just use default case correctly

src/App.tsx
```diff
@@ -26,117 +141,51 @@ export const App: React.FC = () => {
...
+              } else {
+                try {
+                  setIsAdding(true);
+                  pushTodos();
+                } catch {
+                  setAddError(true);
+                } finally {
+                  setIsAdding(false);
+                  setNewTitle('');
+                  setTimeout(cancelErrors, 3000);
```
vlad-demchuk, line 158: create a separate handle function

vlad-demchuk (changes requested):
demo link isn't working

src/components/FilterList/FilterLIst.tsx
```diff
@@ -0,0 +1,80 @@
...
+          }}
+        >
+          {Filter.all}
+        </a>
+
+        <a
+          data-cy="FilterLinkActive"
+          href="#/active"
+          className={cn('filter__link',
+            { selected: filterBy === Filter.active })}
```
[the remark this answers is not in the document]
p-omogit-e: image
image

vlad-demchuk (approved):

PR #191, opened 2023-01-11
Develop

JohnRust12:
[DEMO LINK](https://JohnRust12.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -242,7 +242,7 @@ export const App: React.FC = () => {
       <div className="todoapp__content">
         <Header
           todos={todos}
-          heandleSubmit={handleSubmit}
+          heandlSubmit={handleSubmit}
           title={newTodo?.title || ''}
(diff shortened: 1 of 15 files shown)


qoa11a (changes requested):
Good job!
One fix: if I check two todo items consequently and fast only one get checked on frontend side (query to backend works fine though, as when we refresh the page two items are checked)

JohnRust12:
> Good job! One fix: if I check two todo items consequently and fast only one get checked on frontend side (query to backend works fine though, as when we refresh the page two items are checked)

Nobody knows why this issue happens.

qoa11a (approved):

JohnRust12:
> Good job! One fix: if I check two todo items consequently and fast only one get checked on frontend side (query to backend works fine though, as when we refresh the page two items are checked)

D0ne

PR #197, opened 2023-01-16
possibility of adding todos was implemented

vadiimvooo:
[DEMO LINK](https://vadiimvooo.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,7 +6,7 @@ import React,
   useRef,
   useState,
 } from 'react';
-import { addTodo, getTodos } from './api/todos';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
 import { ErrorMessages } from './components/ErrorMessages';
 import { Footer } from './components/Footer';
@@ -26,6 +26,7 @@ export const App: React.FC = () => {
   const [title, setTitle] = useState('');
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [isAdding, setIsAdding] = useState(false);
+  const [deletingDataIds, setDeletingDataIds] = useState<number[]>([]);
 
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
@@ -96,6 +97,29 @@ export const App: React.FC = () => {
     }
   };
 
+  const handleDeleteButton = (todoId: number) => {
+    setDeletingDataIds(prev => [...prev, todoId]);
+
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos(deletedTodo => {
+          return deletedTodo.filter(todo => todo.id !== todoId);
+        });
+      })
+      .catch(() => setError('Unable to delete a todo'))
+      .finally(() => {
+        setDeletingDataIds([]);
+      });
+  };
+
+  const handleClearButton = () => {
+    todos.forEach(todo => {
+      if (todo.completed) {
+        handleDeleteButton(todo.id);
+      }
+    });
+  };
+
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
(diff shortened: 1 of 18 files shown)


src/App.tsx
```diff
@@ -1,223 +1,165 @@
...
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isAdding, setIsAdding] = useState(false);
+  const [deletingDataIds, setDeletingDataIds] = useState<number[]>([]);
 
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
```
TetKopot, line 35: you could move this part to Header component

src/App.tsx
```diff
@@ -1,223 +1,165 @@
...
+      case 'Completed':
+        return todo.completed;
+
+      case 'Active':
+        return !todo.completed;
+
+      default:
+        return true;
+    }
+  });
```
TetKopot, line 59: move this part to some helper file

TetKopot (changes requested):
JG! Just a little changes needed

TetKopot (approved):

src/App.tsx
```diff
@@ -1,223 +1,150 @@
...
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>('All');
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isAdding, setIsAdding] = useState(false);
+  const [deletingDataIds, setDeletingDataIds] = useState<number[]>([]);
```
ivan-kadykalo, line 30:
```suggestion
  const [todoIdsForDelete, setTodoIdsForDelete] = useState<number[]>([]);
```

src/App.tsx
```diff
@@ -1,223 +1,150 @@
...
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
-    if (newTodoField.current) {
-      newTodoField.current.focus();
+    if (user) {
+      getTodos(user.id)
+        .then(setTodos)
+        .catch(() => {
+          setError('Can\'t load todos');
```
ivan-kadykalo, line 37: It's a good practice to use Enum for error messages

src/App.tsx
```diff
@@ -1,223 +1,150 @@
...
+        });
     }
   }, []);
 
+  const remainingTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
+  const filteredTodos = filterTodos(todos, filterStatus);
+
+  const handleSubmitButton = (
+    e: React.FormEvent<HTMLFormElement>,
```
ivan-kadykalo, line 47:
```suggestion
    event: React.FormEvent<HTMLFormElement>,
```

src/App.tsx
```diff
@@ -1,223 +1,150 @@
...
-          </button>
-        </footer>
-      </div>
+        <Header
+          newTodoField={newTodoField}
+          title={title}
+          onChange={setTitle}
+          onSubmit={handleSubmitButton}
+          isAdding={isAdding}
+        />
```
ivan-kadykalo, line 119: I think it will be more understandable if you will save function names in props.

We can use some generic names for props functions (like onChange, onSubmit) if it is a generic pure component, and we know that it will be used several times throw the project, in different ways, depending on its props.
But in this situation I believe would be better to keep the names.

Please check other components, and think about when we really need to rename function names in props and when we can use generic names.

```suggestion
        <Header
          newTodoField={newTodoField}
          title={title}
          setTitle={setTitle}
          handleSubmitButton={handleSubmitButton}
          isAdding={isAdding}
        />
```

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):
As the next task is approved, this one will be too)) But keep in mind that comments in the future.

PR #202, opened 2023-01-21
react_todo-app-add-and-delete

kolya-movchan:
[DEMO LINK](https://kolya-movchan.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,122 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import classNames from 'classnames';
+import React, {
+  useContext, useEffect, useRef, useState,
+} from 'react';
+
+import { getTodos, postTodos, deleteTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Filter } from './components/React/Filter';
+import { NewTodo } from './components/React/NewTodo';
+import { TodoList } from './components/React/TodoList';
+import { Todo } from './types/Todo';
+
+enum Errors {
+  Server = 'Unable to load data from the Server',
+  Post = 'Unable to add the Todo',
+  Add = 'Title can\'t be empty',
+  Delete = 'Unable to delete a todo',
+  Update = 'Unable to update a todo',
+}
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
-  useEffect(() => {
-    // focus the element with `ref={newTodoField}`
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('all');
+  const [error, setError] = useState('');
+  const [input, setInput] = useState('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [todoOnload, setTodoOnLoad] = useState<Todo | null>(null);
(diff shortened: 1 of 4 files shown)


Virass (changes requested):
I can't mark any tasks as completed. Pls fix it

Virass (changes requested):
I can't mark any tasks as completed. Pls fix it

src/App.tsx
```diff
@@ -1,17 +1,122 @@
...
+  const [todoIdsOnRemove, setTodoIdsOnRemove] = useState<number[]>([]);
+  const [isCompleted] = useState(false);
+
+  const focusInput = () => {
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
+  };
+
+  const downloadDataAPI = () => {
```
vlad-demchuk, line 42: use better variable names. fetch or request will be good here

src/App.tsx
```diff
@@ -1,17 +1,122 @@
...
+  };
+
+  const handleTodoDeleteButton = (todoId: number) => {
+    deleteDataAPI(todoId);
+    setTodoIdsOnRemove([...todoIdsOnRemove, todoId]);
+  };
+
+  const handleClearButton = () => {
+    todos
+      .filter(todo => todo.completed)
```
vlad-demchuk, line 70: you should use `setTodos`. it won't re-render data

src/App.tsx
```diff
@@ -1,17 +1,122 @@
...
+        completed: false,
+      };
+
+      setIsAdding(true);
+      setTodoOnLoad(newTodo);
+
+      postTodos(newTodo)
+        .then(() => setIsAdding(false))
+        .catch(() => setError(Errors.Post))
+        .finally(() => downloadDataAPI());
```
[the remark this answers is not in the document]
vlad-demchuk: anyway, I don't see that you add todo to `todos` state. I assume that is the cause your app isn't working correctly.

vlad-demchuk (changes requested):
It isn't working now
Jan-24-2023 10-31-17

Check the working example to convince that isn't an api issue, so try to fix this behavior.
If you have any questions, just feel free to ask in fe_chat
[the working example](https://mate-academy.github.io/react_todo-app-with-api/)

src/App.tsx
kolya-movchan: image

src/App.tsx
```diff
@@ -1,18 +1,120 @@
...
+enum Errors {
+  Server = 'Unable to load data from the Server',
+  Post = 'Unable to add the Todo',
+  Add = 'Title can\'t be empty',
+  Delete = 'Unable to delete a todo',
+  Update = 'Unable to update a todo',
+}
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
TarasHoliuk, line 23: you can delete it now

TarasHoliuk (changes requested):

src/components/React/Filter.tsx
```diff
@@ -0,0 +1,81 @@
...
+            Completed
+          </a>
+        </nav>
+
+        <button
+          data-cy="ClearCompletedButton"
+          type="button"
+          className="todoapp__clear-completed"
+          onClick={() => onCompletedClear()}
+          style={!isCompleted
```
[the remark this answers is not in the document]
kolya-movchan: It is supposed to be 'false' in the state by default, then the status has to be changed to true when there is at least one completed todo. It will be implemented in the 3rd part of the task

tetiana-ivanchenko (changes requested):
image

nazarmatsevych (approved):
👍

PR #213, opened 2023-01-25
solution

Fynkzia:
[DEMO LINK](https://Fynkzia.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,146 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
-import { AuthContext } from './components/Auth/AuthContext';
+import React, {
+  useContext,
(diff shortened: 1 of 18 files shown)


src/App.tsx
```diff
@@ -1,223 +1,183 @@
...
+      switch (filterType) {
+        case FilterType.Active:
+          return !todo.completed;
+
+        case FilterType.Completed:
+          return todo.completed;
+
+        default:
+          throw new Error('Invalid type');
+      }
```
4ebupel, line 64: It's better to make  a filter inside the switch, 'cause now swich is triggered on every element of an array

4ebupel (approved):
One small mistake, but otherwise - GJ🔥

ivan-kadykalo (approved):
Auto approve, all comments will be in the next task

PR #216, opened 2023-01-25
in progress

KrisPlanka:
[DEMO LINK](https://KrisPlanka.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,87 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { getTodos } from './api/todos';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
+import { Todo } from './types/Todo';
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,223 +1,153 @@
...
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState(Filter.All);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [title, setTitle] = useState('');
```
batstolya, line 29: maybe I would move this state to the header, but I don’t think it’s required

batstolya (approved):
good job🔥

src/App.tsx
ivan-kadykalo: If you are using this value in App.tsx, it should be here, if no, then move it to the header.

src/App.tsx
```diff
@@ -1,223 +1,153 @@
...
+    filteredTodosFunction(todos, status)
+  ), [todos, status]);
+
+  const createTodo = useCallback(
+    async (event: React.FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
+
+      try {
+        if (!title.trim()) {
+          setErrorMessage('Title can\'t be empty');
```
ivan-kadykalo, line 66: Good practice to use ENUM for errors, so please fix it in the next task
```suggestion
          setErrorMessage(TodoErrors.EmptyTitle);
```

ivan-kadykalo (approved):
Good job 🔥

src/App.tsx
KrisPlanka: Sure, thank you!

PR #217, opened 2023-01-25
commit message

11Olya112:


src/App.tsx
```diff
@@ -1,223 +1,184 @@
...
+    .filter((todo) => todo.id !== todoId));
+
+  const removeTodo = (todoId: number | undefined) => {
+    deleteTodo(todoId)
+      .then((todoID) => removTodo(todoID))
+      .catch(() => setisDelete(true));
+  };
+
+  const addNewTodo = (todo: Todo) => setTodos([...todos, todo]);
+  const handleKeyDown = (event: any) => {
```
polosanya, line 43: Same here. Please fix all cases

src/App.tsx
```diff
@@ -1,223 +1,184 @@
...
+    if (event.key === 'Enter' && value !== '') {
+      createTodo(value, user?.id)
+        .then((todo) => addNewTodo(todo))
+        .catch(() => setisError(true));
+      setValue('');
+    } else if (event.key === 'Enter' && value === '') {
```
polosanya, line 49: You can nest value check inside Enter to make your code a bit cleaner)

polosanya (changes requested):
Let's improve your code a bit) 
Try not to use `any` in your code

src/App.tsx
```diff
@@ -1,223 +1,186 @@
...
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [value, setValue] = useState('');
+
+  const [currentFilter, setCurrentFilter] = useState(Filters.All);
+  const [isError, setisError] = useState(false);
+  const [isDelete, setisDelete] = useState(false);
+  const [isHidden, setisHidden] = useState(false);
+  const [isEmpty, setisEmpty] = useState(false);
+  const [isAdding, setisAdding] = useState(false);
```
agniya-i, line 25:
```suggestion
  const [isError, setIsError] = useState(false);
  const [isDelete, setIsDelete] = useState(false);
  const [isHidden, setIsHidden] = useState(false);
  const [isEmpty, setIsEmpty] = useState(false);
  const [isAdding, setIsAdding] = useState(false);
```

src/App.tsx
```diff
@@ -1,223 +1,186 @@
...
+  const completedTodo = todos.filter((todo) => todo.completed);
+  const clearCompleted = () => {
+    completedTodo.filter((todo) => deleteTodo(todo.id)
+      .then()
+      .catch(() => setisDelete(true)));
+  };
+
+  const removTodo = (todoId: unknown) => setTodos(todos
```
agniya-i, line 33:
```suggestion
  const updateTodos = (todoId: number) => setTodos(todos
```
it's a little bit confusing to have `removTodo` and `removeTodo`, you can change to updateTodos,  for example, or use `setTodos` right in `.then`
also, I can suppose that that `todoId` have number type

agniya-i (changes requested):

src/App.tsx
11Olya112: Yes, but the todoId was written like unkown type in fetchClient file, so I can't fix it

agniya-i (approved):

PR #230, opened 2023-01-26
React todo_app add-and-delete done

DiDemkiv:
[DEMO LINK](https://DiDemkiv.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -17,18 +17,31 @@ export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [sortType, setSortType] = useState<FilterType>(FilterType.ALL);
   const [errorText, setErrorText] = useState('');
+  const [tempNewTodo, setTempNewTodo] = useState<Todo | null>(null);
 
   const showErrorBanner = (errorMsg: string) => {
     setErrorText(errorMsg);
     setTimeout(() => setErrorText(''), 3000);
   };
 
-  useEffect(() => {
+  const loadTodos = () => {
     if (user) {
       getTodos(user.id)
-        .then(setTodos).catch(() => showErrorBanner('Cant load user todos'));
+        .then(setTodos)
+        .catch(() => showErrorBanner('Cant load user todos'));
     }
+  };
 
+  const removeTodo = (todoId: number) => setTodos(
+    todos.filter(todo => todo.id !== todoId),
+  );
+
+  const removeTodos = (todoIds: number[]) => setTodos(
+    todos.filter(todo => !todoIds.includes(todo.id)),
+  );
+
+  useEffect(() => {
+    loadTodos();
     // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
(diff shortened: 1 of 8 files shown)


Anastasia4sik (changes requested):
You should fix this code a little bit, cause I can`t add todo, the website just reloads and doesn`t show me anything

Anastasia4sik (approved):

src/App.tsx
```diff
@@ -1,223 +1,107 @@
...
 
   useEffect(() => {
+    loadTodos();
     // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
   }, []);
 
+  const filteredTodos: Todo[] = useMemo(() => todos.filter(todo => {
```
solaryasha, line 51: It would be best if you don't filter todos if `sortType` equals `all`

src/App.tsx
```diff
@@ -1,223 +1,107 @@
...
-            <div data-cy="TodoLoader" className="modal overlay">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
+        <Header
+          newTodoField={newTodoField}
+          todos={filteredTodos}
+          showErrorBanner={showErrorBanner}
+          user={user}
```
solaryasha, line 72: Why do you need to pass `user` inside header?

src/App.tsx
```diff
@@ -1,223 +1,107 @@
...
+          setTempNewTodo={setTempNewTodo}
+          setTodos={setTodos}
```
solaryasha, line 74: Why do you need to pass these functions to the Header component?
It would be great to create a function which will add todo and pass this function to the Header

src/App.tsx
```diff
@@ -1,223 +1,107 @@
...
+        />
 
-          <div data-cy="Todo" className="todo">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
+        { filteredTodos.length || sortType !== 'all'
```
solaryasha, line 77: Let's create a variable for the expression.

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+export type Props = {
+  setErrorText: (arg: string) => void,
+  errorText: string
+};
+
+export const ErrorNotification: React.FC<Props> = ({
```
solaryasha, line 8: What about `React.memo` ?

src/components/Footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const todosToDelete = todos.filter(todo => todo.completed);
+
+    await Promise.all(
+      todosToDelete.map(
+        todo => deleteTodo(todo.id),
+      ),
+    );
+
+    removeTodos(todosToDelete.map(todo => todo.id));
+  };
```
solaryasha, line 27: Why do we need to pass an array with all todos ?
Let's create function clearCompletedTodos and pass it to the Footer

src/components/Footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </a>
+      </nav>
+
+      { todos.find(todo => todo.completed)
+      && (
+        <button
+          data-cy="ClearCompletedButton"
+          type="button"
+          className="todoapp__clear-completed"
+          onClick={() => clearCompletedTodos()}
```
solaryasha, line 78:
```suggestion
          onClick={clearCompletedTodos}
```

src/components/Header.tsx
```diff
@@ -0,0 +1,87 @@
...
+          .then((createdTodo) => {
+            setTodos([...todos, createdTodo]);
+          })
+          .catch(() => showErrorBanner('Unable to add a todo'))
+          .finally(() => {
+            setTempNewTodo(null);
+            setTitle('');
+            setDisableInput(false);
+          });
+      }}
```
solaryasha, line 72: Why need a separate function for this, please don't use such inline functions(it's harder to support such code)

src/components/TodoRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+        className="todo__remove"
+        data-cy="TodoDeleteButton"
+        onClick={() => {
+          setDeletionProcess(true);
+          deleteTodo(todo.id)
+            .then(() => removeTodo(todo.id))
+            .catch(() => showErrorBanner('Unable to delete a todo'))
+            .finally(() => setDeletionProcess(false));
+        }}
+      >
```
solaryasha, line 46: It's should be separate function

solaryasha (changes requested):

PR #241, opened 2023-01-29
Develop

panianchuk:
https://panianchuk.github.io/react_todo-app-add-and-delete/#/


src/App.tsx
```diff
@@ -1,223 +1,162 @@
...
+  const [error, setError] = useState('');
+  const [isHiddenErrorNote, setIsHiddenErrorNote] = useState(true);
+  const [isAdding, setIsAdding] = useState(false);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const visibleTodos = filterTotos(todos, typeFilter);
+  const activeTodosCount = filterTotos(todos, FilterTypes.Active).length;
+  const completedTodosCount = filterTotos(todos, FilterTypes.Completed).length;
+  const allCompleted = todos.length === completedTodosCount;
+
+  function hiddenErrorNote() {
```
Volodymyr-Melnyk-Fron10der, line 26:
```suggestion
  function showErrorNote() {
```

src/App.tsx
```diff
@@ -1,223 +1,162 @@
...
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const user = useContext(AuthContext);
+  const user = useAuthContext();
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [typeFilter, setTypeFilter] = useState(FilterTypes.All);
+  const [error, setError] = useState('');
+  const [isHiddenErrorNote, setIsHiddenErrorNote] = useState(true);
```
Volodymyr-Melnyk-Fron10der, line 18: It is difficult logic. you have hidden error by default, so it's better to name variable isErrorNoteShown and set it as false. And only when you need to show the error you do setErrorNoteShown(true) - so now it really do what it suppose to do

src/api/filter.ts
```diff
@@ -0,0 +1,12 @@
...
+export function filterTotos(todosArr: Todo[], type: string) {
+  switch (type) {
+    case FilterTypes.Completed:
+      return todosArr.filter((todo) => todo.completed);
+    case FilterTypes.Active:
+      return todosArr.filter((todo) => !todo.completed);
+    default:
+      return todosArr;
+  }
+}
```
Volodymyr-Melnyk-Fron10der, line 12: Move it into `helpers `folder

src/App.tsx
```diff
@@ -1,223 +1,162 @@
...
+  const [isAdding, setIsAdding] = useState(false);
```
Volodymyr-Melnyk-Fron10der, line 19: isLoading / isCreating would sound such better

src/App.tsx
```diff
@@ -1,223 +1,162 @@
...
   const newTodoField = useRef<HTMLInputElement>(null);
+  const visibleTodos = filterTotos(todos, typeFilter);
+  const activeTodosCount = filterTotos(todos, FilterTypes.Active).length;
+  const completedTodosCount = filterTotos(todos, FilterTypes.Completed).length;
+  const allCompleted = todos.length === completedTodosCount;
```
Volodymyr-Melnyk-Fron10der, line 24: use useMemo for better performance

src/App.tsx
```diff
@@ -1,223 +1,162 @@
...
+
+    if (user) {
+      getTodos(user.id)
+        .then((res) => setTodos(res))
+        .catch(() => {
+          setError(ErrorTypes.Loading);
+          hiddenErrorNote();
+        })
+        .finally(() => setIsAdding(false));
+    }
```
Volodymyr-Melnyk-Fron10der, line 48: If there is no user your `isAdding ` will be always true, fix it

src/App.tsx
```diff
@@ -1,223 +1,162 @@
...
+      .catch(() => {
+        setError(ErrorTypes.Remove);
+        setTimeout(() => {
+          setIsHiddenErrorNote(true);
+        }, 3000);
+      });
+  };
+
+  const deleteCompletedTodosHandler = () => {
+    todos.map((todo) => (todo.completed ? deleteTodoHandler(todo.id) : todo));
```
Volodymyr-Melnyk-Fron10der, line 98: Why map? Just use forEach

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, I left comments +

1. move your useAuthContext into `src -> hooks`. Why it is in components folder now?)
2. Choose one method of organazing components: all are just the files | all are folders with files inside
image

src/components/useAuthContext.ts
```diff
@@ -0,0 +1,6 @@
+import { useContext } from 'react';
+import { AuthContext } from '../hooks/AuthContext';
+
+export const useAuthContext = () => {
+  return useContext(AuthContext);
+};
```
DarkMistyRoom, line 6: better import it from the context folder, to avoid confusion. it isn't a component

DarkMistyRoom (approved):

PR #242, opened 2023-01-29
Solution

superpooperxxx:
[DEMO LINK](https://superpooperxxx.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,7 +12,7 @@ import { ErrorsAlerts } from './components/ErrorsAlerts';
 import { TodoAppHeader } from './components/TodoAppHeader/TodoAppHeader';
 import { TodosFooter } from './components/TodosFooter';
 import { TodosList } from './components/TodosList';
-import { getTodos, createTodo } from './api/todos';
+import { getTodos, createTodo, deleteTodo } from './api/todos';
 import { Todo } from './types/Todo';
 import { Filter } from './types/Filter';
 import { filterTodos } from './utils/filterTodos';
@@ -29,7 +29,6 @@ export const App: React.FC = () => {
   const userId = user ? user.id : 0;
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
@@ -87,6 +86,16 @@ export const App: React.FC = () => {
       });
   }, []);
 
+  const removeTodo = (todoId: number) => {
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos(currentTodos => {
+          return currentTodos.filter(todo => todo.id !== todoId);
+        });
+      })
+      .catch(() => setErrorMessage('Unable to delete a todo'));
+  };
+
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
@@ -103,6 +112,7 @@ export const App: React.FC = () => {
             <TodosList
               todos={filteredTodos}
               tempTodo={temporaryTodo}
+              removeTodo={removeTodo}
             />
             <TodosFooter
               activeTodosNum={activeTodosNumber}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,90 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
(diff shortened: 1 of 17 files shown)


OlggaJul (approved):
GJ

src/App.tsx
```diff
@@ -1,223 +1,146 @@
...
+  const [complitedFilter, setComplitedFilter] = useState(Filter.All);
+  const [isAdding, setIsAdding] = useState(false);
+  const [temporaryTodo, setTemporaryTodo] = useState<Todo | null>(null);
+  const userId = user ? user.id : 0;
 
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
```
solaryasha, line 34: You should move this logic inside `Header` component

src/App.tsx
```diff
@@ -1,223 +1,146 @@
...
 
+  const filteredTodos = useMemo(() => (
+    filterTodos(todos, complitedFilter)
+  ), [todos, complitedFilter]);
+
+  const activeTodosNumber = useMemo(() => (
+    todos.filter(todo => !todo.completed).length
+  ), [todos]);
+
+  const someComplited = useMemo(() => {
```
solaryasha, line 53: What do we store inside this variable ?

src/components/ErrorsAlerts/ErrorsAlerts.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import cn from 'classnames';
+
+type Props = {
+  errorMessage: string;
+  closeErrors: (message: string) => void;
+};
+
+export const ErrorsAlerts: React.FC<Props> = ({
```
solaryasha, line 9: What about the `React.memo` ?

solaryasha (changes requested):

src/App.tsx
```diff
@@ -1,223 +1,130 @@
...
-            </a>
-            <a
-              data-cy="FilterLinkCompleted"
-              href="#/completed"
-              className="filter__link"
-            >
-              Completed
-            </a>
-          </nav>
+  const removeTodo = (todoId: number) => {
```
solaryasha, line 79: What about the `useCallback` ? You are passing this function to the another component.

src/App.tsx
```diff
@@ -1,223 +1,130 @@
...
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos(currentTodos => {
+          return currentTodos.filter(todo => todo.id !== todoId);
+        });
+      })
+      .catch(() => setErrorMessage('Unable to delete a todo'));
+  };
+
+  const clearCompleted = () => {
```
solaryasha, line 89: What about the `useCallback` ? You are passing this function to the another component.

src/components/TodoAppHeader/TodoAppHeader.tsx
```diff
@@ -0,0 +1,54 @@
...
+  useEffect,
+  useRef,
+} from 'react';
+
+type Props = {
+  addNewTodo: (title: string) => void;
+  isAdding: boolean;
+};
+
+export const TodoAppHeader: React.FC<Props> = ({
```
solaryasha, line 12: Why don't you wrap all components with `React.memo`  ?

solaryasha (changes requested):

src/App.tsx
superpooperxxx: I did in 3rd part

src/App.tsx
superpooperxxx: I did in 3rd part

src/components/TodoAppHeader/TodoAppHeader.tsx
superpooperxxx: As Ihor told us it is a bad practice to wrap everything in React.memo, as sometimes render happens quicker than checking whether we need or not to rerender component

solaryasha: So what is the condition to use it then? Never or just when do you want it?

superpooperxxx: When you know exactly that the most of the time some component will rerender in vain because of its parent, but makes sense to wrap Header

ivan-kadykalo (approved):
Approved, will check the rest in the third part.

PR #258, opened 2023-02-02
Solution

vitaliidox:
[DEMO LINK](https://vitaliidox.github.io/react_todo-app-add-and-delete/)


src/components/TodosItem/TodosItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodosItem: React.FC<Props> = ({ todo }) => {
+  const [query, setQuery] = useState('');
+  const [isLoad, setisLoad] = useState(false);
+
+  const providerValue = useContext(TodosContext);
```
TarasHoliuk, line 17: That's nice you created context to optimize your app, but it seems redundant here. Because you pass almost all things from `providerValue` to `TodoList` and other components but use it only here
So I guess better to pass values you need from `TodoList`

TarasHoliuk (changes requested):

AlisaVlasova (approved):

PR #261, opened 2023-02-03
Not done yet.

IoghiStudio:
[DEMO LINK](https://Ioghistudio.github.io/react_todo-app-add-and-delete/)


src/components/App/App.tsx
```diff
@@ -0,0 +1,194 @@
...
+        userId: user.id,
+        title: todoTitle,
+        completed: false,
+      };
+
+      setTodoTitle('');
+
+      Api.addTodo(newTodo)
+        .then(response => {
+          setTempTodo({
```
UmizDemud, line 66: When you add the temp todo, dont give it the title. It is just a placeholder, so better not show the result until we have a response from server

```ts
setTempTodo({
  id: 0,
  userId: user.id,
  title: '',
  completed: false,
});

File: src/components/App/App.tsx [hunk 2]
```diff
@@ -0,0 +1,194 @@
...
+          });
+
+          console.log(response);
+          getTodosFromServer();
+        })
+        .catch(() => {
+          setError('Unable to add a todo');
+          setTimeout(() => setError(''), 3000);
+        })
+        .finally(() => {
```
UmizDemud, line 80: And here set it back to null.

```ts
setTempTodo(null);

UmizDemud (changes requested):
I like your code, in general.

There is one mistake and I wrote where it is, so when you solved it I can approve right away.
That mistake is when we add a todo, it shows twice, cause temp todo is never set back to null. 

Good naming conventions

meforoma (changes requested):
- purge redundant commented code
- currently todos are doubled upon adding - check out yourself
let's fix these

PR #264, opened 2023-02-05
Develop

lukacsja:
[DEMO LINK](https://lukacsja.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,18 +24,23 @@ export const App: React.FC = () => {
   // const [hasChangesInTodos, setHasChangesInTodos] = useState(0);
   const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todos);
   const [selectedFilter, setSelectedFilter] = useState<
-  'all' | 'active' | 'completed'
-  >('all');
-
-  // HEADER
+    'all' | 'active' | 'completed'>('all');
   const [todoTitleToAdd, setTodoTitleToAdd] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [idsOfLoadingTodos, setIdsOfLoadingTodos] = useState<number[]>([]);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
+
   const onNewTitleAdd = (event: ChangeEvent<HTMLInputElement>) => {
     setTodoTitleToAdd(event.target.value);
   };
 
   const onSubmit = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
-    if (todoTitleToAdd.length) {
+    if (todoTitleToAdd) {
+      setIsInputDisabled(true);
       const todoToAdd: Todo = {
         id: 0,
         userId: USER_ID,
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,24 +1,182 @@
...
+import { getTodos, addTodo, removeTodo } from './api/todos';
+
+const USER_ID = 6160;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
```
naziamoff, line 23: I guess you can create a separate enum for the error type and make this state typeof this enum. And change it in the `catch`. Will make more sense this way

src/App.tsx
```diff
@@ -1,24 +1,182 @@
...
+  const [isEditing] = useState(false);
+  // const [hasChangesInTodos, setHasChangesInTodos] = useState(0);
```
naziamoff, line 25:
```suggestion
```

src/App.tsx
```diff
@@ -1,24 +1,182 @@
...
+
+const USER_ID = 6160;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [isEditing] = useState(false);
```
naziamoff, line 24: seems this variable is always falsy in your code. Do you really need it? What for?

src/App.tsx
```diff
@@ -1,24 +1,182 @@
...
+  // const [hasChangesInTodos, setHasChangesInTodos] = useState(0);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todos);
+  const [selectedFilter, setSelectedFilter] = useState<Filters>(Filters.All);
```
naziamoff, line 27: acutally your name of the variable not bad as well, so treat this as a suggestion, up to you
```suggestion
  const [filterBy, setFilterBy] = useState<Filters>(Filters.All);
```

src/App.tsx
```diff
@@ -1,24 +1,182 @@
...
+  const [todoTitleToAdd, setTodoTitleToAdd] = useState('');
```
naziamoff, line 28: it is not clear for me what it does by its name. Any ideas on how to improve its name?
Isn't this a newTodoTitle?

src/App.tsx
```diff
@@ -1,24 +1,182 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [idsOfLoadingTodos, setIdsOfLoadingTodos] = useState<number[]>([]);
```
naziamoff, line 30: what is this state for? Probably I can't get it because of the name

src/App.tsx
```diff
@@ -1,24 +1,182 @@
...
+          setIsInputDisabled(false);
+        })
+        .catch(() => {
+          setError('Unable to add a todo');
+        });
+      setTodoTitleToAdd('');
+    } else {
+      setError('Title can\'t be empty');
+    }
+  };
```
naziamoff, line 67: looks really messy. You can validate the title at the very beginning for make it look a bit clearer
```suggestion
  const onSubmit = (event: FormEvent<HTMLFormElement>) => {
    event.preventDefault();
    
    if (!todoTitle.length) {
      setError('Title can\'t be empty');
    }
    
      setIsInputDisabled(true);
      
      const todoToAdd: Todo = {
        id: 0,
        userId: USER_ID,
        title: todoTitleToAdd,
        completed: false,
      };

      setTempTodo(todoToAdd);

      addTodo(todoToAdd)
        .then((newTodo) => {
          setTodos([...todos, newTodo]);
          setFilteredTodos([...todos, newTodo]);
          setTempTodo(null);
          setIsInputDisabled(false);
        })
        .catch(() => {
          setError('Unable to add a todo');
        });
        
      setTodoTitleToAdd('');
  };
```

src/App.tsx
```diff
@@ -1,24 +1,182 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [isEditing] = useState(false);
+  // const [hasChangesInTodos, setHasChangesInTodos] = useState(0);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todos);
+  const [selectedFilter, setSelectedFilter] = useState<Filters>(Filters.All);
+  const [todoTitleToAdd, setTodoTitleToAdd] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [idsOfLoadingTodos, setIdsOfLoadingTodos] = useState<number[]>([]);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
```
naziamoff, line 31: It is better to make names more descriptive, I should understand why are you disabling input
```suggestion
  const [areTodosLoading, setAreTodosLoading] = useState(false);
```

naziamoff (changes requested):
needs polishing 👀

src/App.tsx
lukacsja: its created for later usage, forgot to remove from the 2nd part

src/App.tsx
lukacsja: ye i had hard times adding a name for this. 
its for storing the ID-s of the todos which I want to move to a loading state on clicking the 'Clear completed' button.
or a single ID of a todo for the loading when removing it from the list.

src/App.tsx
lukacsja: thank you for your review, working on them

src/App.tsx
naziamoff: oh, kk. You can just create a separate boolean variable like `shouldRenderThirdPartOfTheTask` to make it clear to the reviewers. Or to write a comment in the PR description

lukacsja: yes, agree

src/App.tsx
naziamoff: do you really need to store them in the state? It does not seem to affect the page behavior

lukacsja: image
i push them into the array, then clear the array.
I think i need them. unless i figure out a better way to track them for the multiple loading states.

lukacsja: image

naziamoff (approved):
🔥

PR #266, opened 2023-02-07
Develop

Grayfox147:
[DEMO LINK](https://Grayfox147.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,7 +2,7 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 
 import classNames from 'classnames';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import {
   getTodos,
   postTodo,
@@ -21,8 +21,8 @@ export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [filter, setFilter] = useState('all');
   const [error, setError] = useState('');
-  const [isItError, setIsItError] = useState(false);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todosBeingTransform, setTodosBeingTransform] = useState<number[]>([]);
 
   useEffect(() => {
     getTodos(USER_ID)
@@ -31,12 +31,11 @@ export const App: React.FC = () => {
         setError('');
       })
       .catch(() => {
-        setIsItError(true);
         setError(Errors.LOADING);
       })
       .finally(() => {
         setTimeout(() => {
-          setIsItError(false);
+          setError('');
         }, 3000);
       });
   }, []);
@@ -44,10 +43,9 @@ export const App: React.FC = () => {
   const handleAddTodo = (todoData: Omit<Todo, 'id'>) => {
     if (!todoData.title) {
       setError(Errors.TITLE);
-      setIsItError(true);
 
       setTimeout(() => {
-        setIsItError(false);
+        setError('');
       }, 3000);
 
       return;
@@ -60,46 +58,52 @@ export const App: React.FC = () => {
         setTimeout(() => {
           setTodos([...todos, newTodo]);
           setTempTodo(null);
-        }, 1000)
+        }, 500)
       ))
       .catch(() => {
         setError(Errors.ADDING);
-        setIsItError(true);
 
         setTimeout(() => {
-          setIsItError(false);
+          setError('');
         }, 3000);
       });
   };
 
   const handleRemoveTodo = (todoId: number) => {
+    setTodosBeingTransform(current => [...current, todoId]);
     removeTodo(todoId)
       .then(() => {
-        const filteredTodos = todos.filter((todo) => todo.id !== todoId);
-
-        setTodos(filteredTodos);
+        setTodos(
+          current => current.filter((todo) => todo.id !== todoId),
+        );
       })
       .catch(() => {
         setError(Errors.REMOVING);
-        setIsItError(true);
 
         setTimeout(() => {
-          setIsItError(false);
+          setError('');
         }, 3000);
+      })
+      .finally(() => {
+        setTodosBeingTransform(
+          todosBeingTransform.filter((id) => id !== todoId),
+        );
       });
   };
 
-  const visibleTodos = todos
-    .filter((todo) => {
-      switch (filter) {
-        case Filter.ACTIVE:
-          return !todo.completed;
-        case Filter.COMPLETED:
-          return todo.completed;
-        default:
-          return true;
-      }
-    });
+  const visibleTodos = useMemo(() => {
+    return todos
+      .filter((todo) => {
+        switch (filter) {
+          case Filter.ACTIVE:
+            return !todo.completed;
+          case Filter.COMPLETED:
+            return todo.completed;
+          default:
+            return true;
+        }
+      });
+  }, [filter, todos]);
 
   if (!USER_ID) {
     return <UserWarning />;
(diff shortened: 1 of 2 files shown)


naziamoff (changes requested):
demo's not working

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
-const USER_ID = 0;
+const USER_ID = 6101;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('all');
+  const [error, setError] = useState('');
+  const [showError, setShowError] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [reload, setReload] = useState(false);
```
naziamoff, line 25: you store boolean value there, but by its name I can't guess it is a boolean

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6101;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('all');
```
naziamoff, line 21: not really good idea to use strings this way, it is really easy to make a typo and it will not work as expected. Let's create a enum for filterBy

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+        setError('Unable to load todos');
+      })
+      .finally(() => {
+        setTimeout(() => {
+          setShowError(false);
+        }, 3000);
+      });
+  }, [reload]);
+
+  const handleAdd = (todo: Todo) => {
```
naziamoff, line 44: handle add what?

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6101;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('all');
+  const [error, setError] = useState('');
+  const [showError, setShowError] = useState(false);
```
naziamoff, line 23: boolean value, but is named like a function

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((data) => {
+        setTodos(data);
+        setError('');
+      })
+      .catch(() => {
+        setShowError(true);
+        setError('Unable to load todos');
```
naziamoff, line 35: use enum here which I mentioned above

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+  };
+
+  const handleClear = (todoIds: number[]) => {
+    todoIds.map((id) => {
+      return removeTodo(id)
+        .then(() => {
+          const filteredTodos = todos.filter((todo) => todo.id !== id);
+
+          setTodos(filteredTodos);
+        })
```
naziamoff, line 99: why not try here?

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+  const visibleTodos = todos
+    .filter((todo) => {
+      switch (filter) {
+        case 'active':
+          return !todo.completed;
+        case 'completed':
+          return todo.completed;
+        default:
+          return true;
+      }
```
naziamoff, line 126: enum

src/App.tsx
```diff
@@ -1,24 +1,220 @@
...
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6101;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('all');
+  const [error, setError] = useState('');
```
naziamoff, line 22: same here, create enum for errortype

src/components/Footer.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        onClick={() => {
+          const selectedTodosIds = todos
+            .filter((todo) => todo.completed)
+            .map((todo) => todo.id);
+
+          onSetClearHandler(selectedTodosIds);
+        }}
```
naziamoff, line 72: should be a separate helper function

src/components/Form.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <input
+        type="text"
+        className={className}
+        placeholder={placeholder}
+        value={title}
+        onChange={(event) => {
+          setTitle(event.target.value);
+        }}
+        disabled={tempDisable}
+
```
naziamoff, line 57:
```suggestion
```

naziamoff (changes requested):

src/components/Form.tsx
Grayfox147: I am not sure why my demo page is not working, and I am not sure how to fix it, I did all of the requested changes, the only one I did not was the implementation of the try inside the handleClearTodos, cause I am not sure how to do it properly.

src/components/TodoList.tsx
```diff
@@ -0,0 +1,54 @@
...
+export const TodoList: React.FC<Props> = ({
+  todos,
+  onRemove,
+}) => {
+  return (
+    <section className="todoapp__main">
+      {todos.map((todo) => (
+        <>
+          <div
+            key={todo.id}
```
meforoma, line 19: the top-level element is the fragment, not the <div>
so `key` should belong to the fragment
```suggestion
        <React.Fragment key={todo.id}>
          <div
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,54 @@
...
+              { completed: todo.completed },
+            )}
+          >
+            <label
+              className="todo__status-label"
+            >
+              <input
+                type="checkbox"
+                className="todo__status"
+                checked={todo.completed}
```
meforoma, line 31: > You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.

src/App.tsx
```diff
@@ -1,24 +1,202 @@
...
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          <button
+            type="button"
+            className={classNames(
+              'todoapp__toggle-all',
+              { active: todos.some((todo) => !todo.completed) },
```
meforoma, line 125: i'd keep it active all-the-time
so you can flip all tasks even if all are completed,
but not crucial

meforoma (changes requested):
good job @Grayfox147 👏🏼 
address the few comments of mine.

src/components/TodoList.tsx
Grayfox147: Done!

src/App.tsx
```diff
@@ -1,24 +1,200 @@
...
+                  { completed: tempTodo.completed },
+                )}
+              >
+                <label
+                  className="todo__status-label"
+                >
+                  <input
+                    type="checkbox"
+                    className="todo__status"
+                    checked={tempTodo.completed}
```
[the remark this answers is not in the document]
Grayfox147: This one I am not sure how to implement, you mean replacing the checked prop for defaultChecked ? in this case the temporary todo it should be read only.

src/App.tsx
Grayfox147: Done!

meforoma (approved):
✅

PR #268, opened 2023-02-08
Develop

SalehAkaJim:
[DEMO LINK](https://SalehBUD.github.io/react_todo-app-add-and-delete/)


src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,85 @@
...
+};
+
+export const TodoFooter: React.FC<Props> = ({
+  todosLength,
+  selectFilter,
+  switchFilter,
+  activeClearBtn,
+  clearCompleted,
+}) => {
+  const changeFilterBy = (filter: Filter) => {
```
naziamoff, line 21: seems you repeat this logic. You can change the filter state directly, no need to create a handler each time

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,53 @@
...
+  createTodo: (title: string) => void;
+  isInputDisabled: boolean;
+};
+
+export const TodoHeader: React.FC<Props> = ({
+  onError,
+  createTodo,
+  isInputDisabled,
+}) => {
+  const [newTodo, setNewTodo] = useState<string>('');
```
naziamoff, line 16: isn't this one a `newTodoTitle`?
A bit tricky one here, by its name I expect it to be an object

src/components/TodoMain/TodoMain.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[] | null;
+  tempTodo: Todo | null;
+  deleteTodo: (id: number) => void;
+};
+
+export const TodoMain: React.FC<Props> = ({ todos, tempTodo, deleteTodo }) => {
+  const [isEditActive, setIsEditActive] = useState<number>(-1);
```
naziamoff, line 12: why is it number? What exactly are you checking here?

src/components/TodoMain/TodoMain.tsx
```diff
@@ -0,0 +1,100 @@
...
+      {todos?.map((todo) => {
+        const { completed, title, id } = todo;
+
+        return (
+          <div className={cn('todo', { completed })} key={id}>
+            <label className="todo__status-label">
+              <input type="checkbox" className="todo__status" checked />
+            </label>
+
+            {isEditActive === id ? (
```
naziamoff, line 37: seems weird, because I expect `isEditActive` to store a boolean value, and here it is compared with the id, which is obviously a number

src/components/TodoMain/TodoMain.tsx
```diff
@@ -0,0 +1,100 @@
...
+                </span>
+                <button
+                  type="button"
+                  className="todo__remove"
+                  onClick={() => onDelete(id)}
+                >
+                  ×
+                </button>
+
+                {loading === id && (
```
naziamoff, line 66: same here

naziamoff (changes requested):

src/components/TodoMain/TodoMain.tsx
SalehBUD: When user double clicks on a todo, I'll save the id in here to activate Edit for it. I'll change the name...

src/App.tsx
```diff
@@ -1,24 +1,114 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<ErrorMessages | null>(null);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+
+  const filterTodos = (filterBy: Filter) => {
+    switch (filterBy) {
```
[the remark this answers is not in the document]
naziamoff: @SalehBUD 
```
const filteredTodos = todos.filter((todo) => {
  switch (filterBy) {
    ...your code goes here...
  }
})
```

src/components/TodoMain/TodoMain.tsx
naziamoff: still not clear and looks kinda hardcoded. There must be a better way. Could you pls explain me the issue in more details?

naziamoff (approved):
not bad, still it is better to fix the mentioned issues in the next part of the task

src/components/TodoMain/TodoMain.tsx
SalehBUD: Here is the thing:
I have 10 todos with id [1, 2, 3, ..., 10]
They're rendered on the page with key={1, 2, 3, ..., 10}
When I want to activate Edit form on one of them, I'll pass the key/id of that todo in order to select it...

I've tried passing a boolean and end up activating Edit form for all of the todos....

PR #270, opened 2023-02-13
Solution

Rijo-J:
- [DEMO LINK](https://rijo-j.github.io/react_todo-app-add-and-delete/)


meforoma (approved):
very limited functionality, yet it does add and delete

would be very great if you implemented the focus on a `<form<input` after adding todo
```
ref = useRef...
...
ref.current.focus()
```

in the final implementation with API and full functionality

PR #273, opened 2023-02-13
Solution

CharAnagn:
[DEMO LINK](https://charanagn.github.io/react_todo-app-add-and-delete/)


meforoma (approved):
i'll pass it onto the final implementation of the whole task,
approving this one as it does the individual adding and deleting of todos

https://user-images.githubusercontent.com/39308717/219640142-4da66fde-dbbd-4fe5-af99-b2bf8cf5f75e.mov

PR #277, opened 2023-02-15
not finished

grslckr:
[DEMO LINK](https://grslckr.github.io/react_todo-app-add-and-delete/)


meforoma (approved):
since you already have a PR for the complete project - let's be fixing all there

PR #303, opened 2023-02-21
add code for deleting and adding todos

Anastasiia-Khudych:
[Demo](https://anastasiia-khudych.github.io/react_todo-app-add-and-delete/#/)


src/api/todos.ts
```diff
@@ -0,0 +1,15 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const addTodo = (userId: number, data: any) => {
```
OleksandrProkop, line 9: dont use 'any ' for type

src/api/todos.ts
```diff
@@ -0,0 +1,15 @@
...
+  return client.post<Todo>(`/todos?userId=${userId}`, data);
+};
+
+export const removeTodo = (userId: number, todoId: number) => {
+  return client.delete(`/todos/${todoId}?userId=${userId}`);
```
OleksandrProkop, line 14: todoId is enough, delete this `?userId=${userId}`

src/components/ErrorNotifications/ErrorNotifications.tsx
```diff
@@ -0,0 +1,66 @@
...
+        'Unable to add a todo'
+      )}
+
+      {error === Error.ONDELETE && (
+        'Unable to delete a todo'
+      )}
+
+      {error === Error.ONUPDATE && (
+        'Unable to update a todo'
+      )}
```
OleksandrProkop, line 63: you can cut it short. like that  `{`Unable to ${error} todo`}`

OleksandrProkop (approved):
GJ

src/api/todos.ts
Anastasiia-Khudych: Thanks!

src/api/todos.ts
Anastasiia-Khudych: What type to use instead? It's not a Todo type because it lacks an id.

denys-cheporniuk (approved):

src/api/todos.ts
OleksandrProkop: if you don't want to get a primitive type, you can use object

PR #328, opened 2023-03-03
solution

Deniskakaka:
https://deniskakaka.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,124 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
+import { deleteTodo, getTodos, postTodo } from './api/todos';
+import { Error } from './components/Error/Error';
+import { Filter } from './components/Filter/Filter';
+import { Item } from './components/Todo/Todo';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6470;
 
 export const App: React.FC = () => {
+  const [titleTodo, setTitleTodo] = useState('');
+  const [listTodo, setListTodo] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [tempTodo, setTempTodo] = useState(false);
+
+  const getListTodo = async (filter?: boolean) => {
+    try {
+      const result = await getTodos(USER_ID, filter);
+
+      setListTodo(result);
+    } catch (e) {
+      setError('Oops, something were wrong, please try again later');
+    }
+  };
+
+  useEffect(() => {
+    getListTodo();
+  }, []);
+
+  const clearError = useCallback(() => setError(''), []);
+
+  const onSubmitHandler = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    setTempTodo(true);
+    try {
+      await postTodo(
+        { title: titleTodo, userId: USER_ID, completed: false },
+      );
+
+      getListTodo();
+      setTitleTodo('');
+      setTempTodo(false);
+    } catch (e) {
+      setError('Oops, something were wrong, please try again later');
+    }
+  };
+
+  const onChangeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitleTodo(event.target.value);
+  };
+
+  const deleteTodoHandler = (id: number) => {
+    deleteTodo(id).then(() => getListTodo());
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {/* this buttons is active only if there are some active todos */}
+          <button type="button" className="todoapp__toggle-all active" />
+
+          {/* Add a todo on form submit */}
+          <form onSubmit={onSubmitHandler}>
+            <input
+              type="text"
+              className="todoapp__new-todo"
+              placeholder="What needs to be done?"
+              value={titleTodo}
+              onChange={onChangeHandler}
+            />
+          </form>
+        </header>
+
+        <section className="todoapp__main">
+          {
+            listTodo.map((el: Todo) => (
+              <Item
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,24 +1,124 @@
...
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {/* this buttons is active only if there are some active todos */}
```
Anastasiia-Svintsova, line 68: remove comments, please

src/App.tsx
```diff
@@ -1,24 +1,124 @@
...
+              placeholder="What needs to be done?"
+              value={titleTodo}
+              onChange={onChangeHandler}
+            />
+          </form>
+        </header>
+
+        <section className="todoapp__main">
+          {
+            listTodo.map((el: Todo) => (
```
Anastasiia-Svintsova, line 85: What does `el` mean? Give more meaningful names

src/App.tsx
```diff
@@ -1,24 +1,124 @@
...
+                todo={el}
+                key={el.id}
+                deleteTodo={deleteTodoHandler}
+              />
+            ))
+          }
+          {tempTodo && (
+            <Item
+              todo={{
+                id: Math.random(),
```
Anastasiia-Svintsova, line 96: Not a good way to create an id. You can use different libraries for it
https://dev.to/salehmubashar/create-a-unique-id-in-react-js-3f75

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,26 @@
+import { useEffect, FC } from 'react';
+
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
Anastasiia-Svintsova, line 3: Don't disable eslint. Try to fix the error

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type Props = {
+  setFilter: (filter?: boolean) => void
+};
+
+enum TypeFilter {
+  'all',
+  'active',
+  'completed',
+}
```
Anastasiia-Svintsova, line 12: Create the right enum, please 
https://www.typescriptlang.org/docs/handbook/enums.html
```suggestion
enum FilterType {
  All = 'all',
  Active = 'active',
  Completed = 'completed',
}
```

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const Filter: React.FC<Props> = ({ setFilter }) => {
+  const [typeFilter, setTypeFilter] = useState(TypeFilter[1]);
```
Anastasiia-Svintsova, line 15: Use enum in a right way, please 
```suggestion
  const [typeFilter, setTypeFilter] = useState(FilterType.All);
```

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  return (
+    <nav className="filter">
+      <a
+        href="#/"
+        className={classNames('filter__link', {
+          selected: typeFilter === TypeFilter[0],
+        })}
+        onClick={() => chooseFilter(TypeFilter[0])}
```
Anastasiia-Svintsova, line 29: Same here

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,56 @@
...
+  'all',
+  'active',
+  'completed',
+}
+
+export const Filter: React.FC<Props> = ({ setFilter }) => {
+  const [typeFilter, setTypeFilter] = useState(TypeFilter[1]);
+
+  const chooseFilter = (type: string, filter?: boolean) => {
+    setFilter(filter);
```
Anastasiia-Svintsova, line 18: Why do you need it? `setFilter(filter)` 
Seems that it's redundant a bit

Anastasiia-Svintsova (changes requested):
- Don't show the footer if there are no todos in the list. Also, don't show the clear completed button if there are no completed todos
image

- Look, the todo disappears for a moment before adding. Fix it, please 

https://user-images.githubusercontent.com/99403198/222896571-d14ec24b-bd21-4838-95a8-ddf55a157f0e.mov

- The todo should disappear from the list after the error

https://user-images.githubusercontent.com/99403198/222896611-5d8082f1-172b-43cc-a3cc-ed0a12994f88.mov

- And then I can see the strange behavior for the input and todo item. Watch the screencast

https://user-images.githubusercontent.com/99403198/222896648-130f89bf-6d16-439d-be05-74590deab172.mov

- You should show some error for the deleting process as well, now I can see just infinite loader 

https://user-images.githubusercontent.com/99403198/222896721-172a4ab4-359b-4a65-804d-6e53b5eb684e.mov

- The `items left` amount shouldn't change from filter to filter 

https://user-images.githubusercontent.com/99403198/222896748-bb989f90-17aa-4098-b79a-01cfc18bc13f.mov

src/components/Error/Error.tsx
Deniskakaka: i can't because need change eslint, but i don't have opportunity

src/components/Filter/Filter.tsx
Deniskakaka: i don't see need repead name, like this i use number and take string

src/components/Filter/Filter.tsx
Deniskakaka: here i use for render new list todos, where change type filter

VitaliyBondarenko1982 (changes requested):
Did you make deploy after fixes?
I see same problems from first review.

Deniskakaka:
> Did you make deploy after fixes? I see same problems from first review.

yes i forgot, now i made deploy and you can check

Deniskakaka:
> Hi!)
> 
> I checked first three mentor's remarks and unfortunately they aren't fixed. Could you check wether they're deployed or not. If you need any help - feel free to ask at fe_chat

i deploy mabe need time, it bad ida create one app on some taks, i did it in next task and add new functional, it more confising me, try to check deploy later

etojeDenys:
> > Hi!)
> > I checked first three mentor's remarks and unfortunately they aren't fixed. Could you check wether they're deployed or not. If you need any help - feel free to ask at fe_chat
> 
> i deploy mabe need time, it bad ida create one app on some taks, i did it in next task and add new functional, it more confising me, try to check deploy later

maybe, just re-request me when everything will work okay

Deniskakaka:
> .

it was last time, it not interesting for me create the same thing in several tasks, I had it all implemented in the following task

Viktor-Kost (changes requested):
Denis, hello!
Unfortunately, I still see no changes and can't approve. It would be better if you ask at fe_chat. 

For example, first mentor's remark was about hiding footer, when we don't have todos.

Here is your current code. There is no condition for shown/hidden footer
Screenshot 2023-03-07 at 09 39 27

PR #332, opened 2023-03-07
Solution

OksanaBaloh:
[DEMO LINK](https://OksanaBaloh.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,159 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { createTodo, deleteTodo, getTodos } from './api/todos';
+import { ErrorNotification } from './components/ErrorNotification';
+import { ErrorType } from './types/ErrorType';
+import { FilterBy } from './types/FilterBy';
 
-const USER_ID = 0;
+const USER_ID = 6481;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState(ErrorType.None);
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+  const [filterBy, setFilterBy] = useState(FilterBy.All);
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isInputActive, setIsInputActive] = useState(true);
+
+  const [updatingId, setUpdatingId] = useState([0]);
+
+  const getTodosFromServer = async () => {
+    try {
+      setErrorMessage(ErrorType.None);
+      const todosFromServer = await getTodos(USER_ID);
+
+      setTodos(todosFromServer);
+    } catch (error) {
+      setErrorMessage(ErrorType.Load);
+    }
+  };
+
+  useEffect(() => {
+    getTodosFromServer();
+  }, []);
+
+  const allTodosCount = todos.length;
+
+  const activeTodosCount = useMemo(
+    () => todos.filter(todo => !todo.completed).length,
+    [todos],
+  );
+
+  useEffect(() => {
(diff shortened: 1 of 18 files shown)


src/App.tsx
```diff
@@ -1,24 +1,159 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState(ErrorType.None);
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+  const [filterBy, setFilterBy] = useState(FilterBy.All);
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isInputActive, setIsInputActive] = useState(true);
+
+  const [updatingId, setUpdatingId] = useState([0]);
```
loralevitska, line 23: It should be number according naming. Change naming or variable type
```suggestion
  const [updatingId, setUpdatingId] = useState([0]);
```

src/App.tsx
```diff
@@ -1,24 +1,159 @@
...
+    const filteredTodos = (() => {
+      switch (filterBy) {
+        case FilterBy.All:
+          return todos;
+        case FilterBy.Active:
+          return todos.filter(todo => !todo.completed);
+        case FilterBy.Completed:
+          return todos.filter(todo => todo.completed);
+        default:
+          return todos;
```
loralevitska, line 57:
```suggestion
      switch (filterBy) {
        case FilterBy.All:
          return todos;
          
        case FilterBy.Active:
          return todos.filter(todo => !todo.completed);
          
        case FilterBy.Completed:
          return todos.filter(todo => todo.completed);
          
        default:
          return todos;
```

src/App.tsx
```diff
@@ -1,24 +1,159 @@
...
+        setErrorMessage(ErrorType.None);
+        const newTodo = {
+          userId: USER_ID,
+          title,
+          completed: false,
+        };
+
+        setIsInputActive(false);
+        setTempTodo({ ...newTodo, id: 0 });
+        await createTodo(USER_ID, newTodo);
```
loralevitska, line 82: this method returns instance of created todo which have id. So write result to variable and use for function `setTodos` - add created todo to your array.
And explain please for what you need `tempTodo`?

src/App.tsx
```diff
@@ -1,24 +1,159 @@
...
+    };
+
+    createNewTodo();
+  };
+
+  const removeTodo = async (id: number) => {
+    try {
+      setErrorMessage(ErrorType.None);
+      setUpdatingId(prev => [...prev, id]);
+      await deleteTodo(USER_ID, id);
```
loralevitska, line 99: Same here. Wait the result instance and remove todo from `todos` by id

loralevitska (changes requested):
You did an amazing job! Looks good! Let's do some improvements (check comments)
- Remove default outline
image

-

src/App.tsx
OksanaBaloh: TempTodo is a part of task) After clicking `enter`, there may be a delay in creating a _todo_, but it is important for user to see that the _todo_ is being created. That's why, a _tempTodo_ with loader is displayed at the end of the list, and after creating a "permanent" _todo_, the  _tempTodo_ is deleted.

src/App.tsx
```diff
@@ -1,24 +1,164 @@
...
+        case FilterBy.Completed:
+          return todos.filter(todo => todo.completed);
+
+        default:
+          return todos;
+      }
+    });
+
+    setVisibleTodos(filteredTodos);
+  }, [filterBy, todos]);
```
IvanFesenko, line 65: It's bad practice change state inside useEffect. use useMemo instead  `const [visibleTodos, setVisibleTodos] = useState(todos);`

IvanFesenko (changes requested):
Check comment please

src/App.tsx
```diff
@@ -1,24 +1,159 @@
...
+  const activeTodosCount = useMemo(
+    () => todos.filter(todo => !todo.completed).length,
+    [todos],
+  );
+
+  const visibleTodos = useMemo(() => {
+    switch (filterBy) {
+      case FilterBy.All:
+        return todos;
+
```
sTorba24, line 50: you can remove it, when you have ALL value => default value will do the same :)

sTorba24 (approved):
Cool! Approved :)

PR #337, opened 2023-03-13
solved

Deniso4ka23:
[DEMO LINK](https://Deniso4ka23.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,159 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
 import { UserWarning } from './UserWarning';
+import { TodoType } from './types/TodoType';
+import { Header } from './components/Header';
+import { Footer, SortType } from './components/Footer';
+import { Section } from './components/Section';
 
-const USER_ID = 0;
+const USER_ID = 6526;
+
+const getReorderedList = (sortType: SortType, list: TodoType[]) => {
+  switch (sortType) {
+    case SortType.Active:
+      return list.filter(item => !item.completed);
+    case SortType.Completed:
+      return list.filter(item => item.completed);
+    default:
+      return list;
+  }
+};
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [todos, setTodos] = useState<TodoType[]>([]);
+  const [selectedSortType, setSortType] = useState(SortType.All);
+  const [isError, setError] = useState(false);
+  const [errorText, setErrorText] = useState('');
+  const [isDisable, setDisable] = useState(false);
+  const [tempTodo, setTempTodo] = useState<TodoType | null>();
+  const [isRemoveAll, setRemoveAll] = useState(false);
+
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,24 +1,159 @@
...
+      return list;
+  }
+};
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [todos, setTodos] = useState<TodoType[]>([]);
+  const [selectedSortType, setSortType] = useState(SortType.All);
+  const [isError, setError] = useState(false);
+  const [errorText, setErrorText] = useState('');
```
anstsot, line 27: you can leave just one error state variable
```suggestion
  const [errorText, setErrorText] = useState('');
```

src/App.tsx
```diff
@@ -1,24 +1,159 @@
...
+      >
+        <button
+          type="button"
+          className="delete"
+          onClick={() => setError(false)}
+        >
+          {}
+        </button>
+        {errorText}
+      </div>
```
anstsot, line 156: please create separate component for it

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,63 @@
+import classNames from 'classnames';
+import { TodoType } from '../../types/TodoType';
+
+export enum SortType {
```
anstsot, line 4: maybe it would be better to rename it? because it seems to be filter not sort
```suggestion
export enum FilterType {
```

anstsot (changes requested):
Good job but there are few fixes to improve your code.

and also fix case when there are no completed todos and the filters are on the right side (should be in the middle)
image

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <div
+      className="notification is-danger is-light has-text-weight-normal"
+      hidden={!errorText}
+    >
+      <button
+        type="button"
+        className="delete"
+        onClick={() => setErrorText('')}
+      >
+        {}
```
witflash, line 20: Why do you need this empty block?
You can remove it and create `button` as a self-closed tag.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <header className="todoapp__header">
+      {hasTodos && (
+        <button
+          type="button"
+          className={classNames(
+            'todoapp__toggle-all',
+            { active: activeTodos },
+          )}
+        >
+          {}
```
witflash, line 46: The same here

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,91 @@
...
+  onDelete: (todoId: number) => void;
+};
+
+export const TodoInfo: React.FC<Props> = ({ todo, onDelete, isRemoveAll }) => {
+  const {
+    title,
+    completed,
+    id,
+  } = todo;
+  const [isEdited, setEdit] = useState(false);
```
witflash, line 17: It's better to name it `isEditing` (the same as isLoading for example)

witflash (changes requested):
Good job! 
Almost done, but I have several comments for some improvements.

agniya-i (approved):

PR #348, opened 2023-03-24
React_Todo App Add and Delete(second stage) was done

illia-kots:
[DEMO LINK](https://illia-kots.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/Components/Notifications.tsx
+++ b/src/Components/Notifications.tsx
@@ -1,0 +1,44 @@
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
+import { ErrorTypes } from '../types/ErrorTypes';
+
+type Props = {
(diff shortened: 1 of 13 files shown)


src/Components/Notifications.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const Notifications: React.FC<Props> = ({ setError, error }) => {
+  const [isHidden, setIsHidden] = useState(false);
+
+  const deleteError = () => {
+    setIsHidden(true);
+    setTimeout(() => {
+      setError(null);
+    }, 1000);
```
anstsot, line 17: do you really need the setTimeout here? maybe it would be better to remove the error immediately?
```suggestion
    setError(null);
```

anstsot (approved):
Great job, approved!

PR #350, opened 2023-03-25
Solution

IVxVI:
[DEMO LINK](https://ivxvi.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,190 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useState, useCallback, useEffect, useMemo,
+} from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { getTodos, postTodo, deleteTodo } from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
+import { FilterType } from './types/FilterType';
+import { Footer } from './components/Footer/Footer';
 
-const USER_ID = 0;
+const USER_ID = 6704;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [disabledInput, setDisabledInput] = useState(false);
+  const [query, setQuery] = useState('');
+  const [errorText, setErrorText] = useState('');
+  const [filterType, setFilterType] = useState(FilterType.All);
+  const [loading, setLoading] = useState(false);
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+    addTodo(query);
+  }, [fetchTodos]);
+
+  const handleSubmitForm = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    addTodo(query);
+    setQuery('');
+  };
+
+  const removeCompleted = () => {
```
Volodymyr-Melnyk-Fron10der, line 114: looks like a redundant function

src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+        </header>
+
+        <TodoList
+          todos={visibleTodos}
+          tempTodo={tempTodo}
+          loading={loading}
+          removeTodo={removeTodo}
+          deletedTodoId={deletedTodoId}
+        />
+        {todos.length > 0 && (
```
Volodymyr-Melnyk-Fron10der, line 165:
```suggestion
        {!!todos.length && (
```

src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+      setErrorText('Unable to delete completed todos');
+      closeError();
+    } finally {
+      setDeletedTodoId([]);
+    }
+  };
+
+  useEffect(() => {
+    fetchTodos();
+    addTodo(query);
```
Volodymyr-Melnyk-Fron10der, line 105: why you need to add todo on mounted event?

src/App.tsx
```diff
@@ -1,24 +1,190 @@
...
+      setTempTodo({ ...newTodo, id: 0 });
+
+      const createdTodo = await postTodo(newTodo);
+
+      setTodos(prevTodos => [...prevTodos, createdTodo]);
+    } catch (error) {
+      setErrorText('Unable to add todo');
+      closeError();
+
+      throw error;
```
Volodymyr-Melnyk-Fron10der, line 68: i don't see a reason to throw it

Volodymyr-Melnyk-Fron10der (changes requested):
1. this box shouldn't be visible on the start
image

2. filters should be centered
image

src/App.tsx
IVxVI: Right, thanks, it`s redundant here)

src/App.tsx
IVxVI: Thank you, fixed

src/App.tsx
IVxVI: Fixed

src/App.tsx
IVxVI: Fixed)

anstsot (approved):
Great job, approved!

PR #358, opened 2023-04-02
React_todo-app-add-and-delete

SerhiiVovchenko:
[DEMO LINK](https://SerhiiVovchenko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,135 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useCallback } from 'react';
 import { UserWarning } from './UserWarning';
+import { getTodos, addTodo, deleteTodo } from './api/todos';
+import { Todo } from './types/Todo';
+import { filterTodos } from './helpers/helpers';
+import { FilterParam } from './types/FilterParam';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorType } from './types/ErrorType';
+import { Error } from './components/Error';
 
-const USER_ID = 0;
+const USER_ID = 6755;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [filterType, setFilterType] = useState<FilterParam>(FilterParam.All);
+  const [errorType, setErrorType] = useState(ErrorType.None);
+  const [title, setTitle] = useState('');
+
+  const handleTextChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
+  const fetchTodos = async () => {
+    try {
+      const todosFromServer = await getTodos(USER_ID);
+
+      setTodos(todosFromServer);
+    } catch (error) {
+      setHasError(true);
+      setErrorType(ErrorType.Load);
+    }
+  };
+
+  const addNewTodo = useCallback(
+    (newTodo: Todo): void => setTodos((oldTodos) => [...oldTodos, newTodo]),
+    [],
+  );
+
+  const showError = useCallback((error: ErrorType) => {
+    setErrorType(error);
+    setHasError(true);
+  }, []);
+
+  const hideError = useCallback(() => {
+    setHasError(false);
+  }, []);
+
+  const removeTodo = (todoId: number) => {
+    deleteTodo(todoId)
+
+      .then(() => fetchTodos())
+      .catch(() => {
+        setHasError(true);
+        showError(ErrorType.Delete);
+      });
+  };
+
+  const handleFormSubmit = (event: React.FocusEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!title) {
+      showError(ErrorType.EmptyTitle);
+
+      return;
+    }
+
+    const newTodo = {
+      userId: USER_ID,
+      title,
+      completed: false,
+    };
+
+    const addNewTodoInList = async () => {
+      try {
+        const downloadNewTodo = await addTodo(USER_ID, newTodo);
+
+        addNewTodo(downloadNewTodo);
+      } catch {
+        showError(ErrorType.Add);
+      } finally {
+        setTitle('');
+      }
+    };
+
+    addNewTodoInList();
+  };
+
+  useEffect(() => {
+    fetchTodos();
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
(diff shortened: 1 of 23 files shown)


src/App.tsx
```diff
@@ -1,24 +1,135 @@
...
+  }, []);
+
+  const removeTodo = (todoId: number) => {
+    deleteTodo(todoId)
+
+      .then(() => fetchTodos())
+      .catch(() => {
+        setHasError(true);
+        showError(ErrorType.Delete);
+      });
```
Anastasiia-Svintsova, line 58: Please, try to use only one method to work with asynchronous functions. You use try/catch with async/await above, so why do you use then() here?

src/App.tsx
```diff
@@ -1,24 +1,135 @@
...
+      try {
+        const downloadNewTodo = await addTodo(USER_ID, newTodo);
+
+        addNewTodo(downloadNewTodo);
+      } catch {
+        showError(ErrorType.Add);
+      } finally {
+        setTitle('');
+      }
+    };
```
Anastasiia-Svintsova, line 86: Try not to nest functions. You can move this addNewTodoInList function outside the handleFormSubmit

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,65 @@
...
+  hasError: boolean,
+  onNotificationClose: () => void,
+};
+
+export const Error: React.FC<Props> = ({
+  errorType,
+  hasError,
+  onNotificationClose,
+}) => {
+  let errorMessage = '';
```
Anastasiia-Svintsova, line 16: Why `let` but not state here? Use useState() for this value, please

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,65 @@
...
+    case ErrorType.Update:
+    case ErrorType.Load:
+      errorMessage = `Unable to ${errorType} a todo`;
+
+      break;
+
+    case ErrorType.EmptyTitle:
+      errorMessage = 'Title can\'t be empty';
+
+      break;
```
Anastasiia-Svintsova, line 30:
```suggestion
      errorMessage = `Unable to ${errorType} a todo`;
      break;

    case ErrorType.EmptyTitle:
      errorMessage = 'Title can\'t be empty';
      break;
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,86 @@
...
+        return setFilterType(FilterParam.All);
+
+      case FilterParam.Active:
+        return setFilterType(FilterParam.Active);
+
+      case FilterParam.Completed:
+        return setFilterType(FilterParam.Completed);
+
+      default:
+        return setFilterType(FilterParam.All);
```
Anastasiia-Svintsova, line 38: Try to rename `href` to a more meaningful value
Can't you just do smth like this:
```
const href = event.currentTarget.text || FilterParam.All;
setFilterType(href)

File: src/components/TodoErrorList/TodoErrorList.tsx [hunk 6]
```diff
@@ -0,0 +1,45 @@
...
+  const hideError = () => {
+    setIsErrorVisible(false);
+
+    setError(false);
+  };
+
+  useEffect(() => {
+    setTimeout(() => {
+      setError(false);
+    }, 3000);
```
Anastasiia-Svintsova, line 21: You should clear all timeouts on the component unmount
https://bobbyhadz.com/blog/react-cleartimeout

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <label className="todo__status-label">
+        <input
+          type="checkbox"
+          className="todo__status"
+          checked={completed}
+          onClick={toggleCompleted}
+        />
+      </label>
+
+      <span className="todo__title">{todo.title}</span>
```
Anastasiia-Svintsova, line 33: Use destructuring for the todo

Anastasiia-Svintsova (changes requested):
- The user shouldn't be able to add an empty todo

https://user-images.githubusercontent.com/99403198/229465447-ef6263cc-d337-4025-9c65-7f48f2e63e10.mov

- These requirements are not implemented 
image

https://user-images.githubusercontent.com/99403198/229465952-b784d10f-5e5a-4baa-84b6-bd61343cddf8.mov

- Filters should always be aligned to the center. Fix it, please 
image

src/App.tsx
```diff
@@ -1,24 +1,151 @@
...
+  const showError = useCallback((error: ErrorType) => {
+    setErrorType(error);
+    setHasError(true);
+  }, []);
+
+  const hideError = useCallback(() => {
+    setHasError(false);
+  }, []);
+
+  const removeTodo = async (todoId: any) => {
```
polosanya, line 54: Don't use any. Specify the type of todoId
image

src/App.tsx
```diff
@@ -1,24 +1,151 @@
...
+      ));
+    } catch {
+      setHasError(true);
+      showError(ErrorType.Delete);
+    } finally {
+      setLoadingTodo([0]);
+    }
+  };
+
+  const addNewTodoInList = async (newTodo: any) => {
```
polosanya, line 70: Same here. Please fix all cases

src/api/todos.ts
```diff
@@ -0,0 +1,14 @@
+import { client } from '../utils/fetchClient';
+import { Todo } from '../types/Todo';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+export const addTodo = (userId: number, newTodo: unknown): Promise<Todo> => {
```
polosanya, line 8: Is a type of newTodo unknown? You have a Todo type, which represents the todo

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,41 @@
...
+    options.headers = {
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
polosanya, line 29: It would be great to add some error text

polosanya (changes requested):
Almost perfect!

witflash (changes requested):
Almost good. Last but not least thing. If I repeatedly press Enter button on add Todo it will be created several same todos. 

https://user-images.githubusercontent.com/25587299/229810863-170d334c-80b7-4f7c-b1ef-6f26f77833e4.mov

So what can you do? 
You have to handle `isLoading` (or `isSending`) state which will be `true` immediately after new todo starts sending. And then you have to check this state before creating Todo. If it `true` you have to skip sending. 
After the request will be done (it doesn't matter if it will be successful or not) you should set this state to `false`.

polosanya (approved):
Nice! Good job!

PR #370, opened 2023-04-10
:white_check_mark: add and delete task was implemented

Bogdan-Mykhailov:
[DEMO LINK](https://Bogdan-Mykhailov.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,10 +13,8 @@ import { AddTodo } from './components/AddTodo';
 import { TodoList } from './components/TodoList';
 import { Notification } from './components/Notification';
 import { Filter } from './components/Filter';
-import { getFilteredTodos } from './utils/helpers';
-import { DEFAULT_TASK_ID } from './utils/constants';
-
-const USER_ID = 6972;
+import { closeErrorMessage, getFilteredTodos } from './utils/helpers';
+import { DEFAULT_TASK_ID, USER_ID } from './utils/constants';
 
 export const App: FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
@@ -25,6 +23,7 @@ export const App: FC = () => {
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [isDisabled, setIsDisabled] = useState<boolean>(false);
   const [loadTodoById, setLoadTodoById] = useState([DEFAULT_TASK_ID]);
+
   const fetchTodos = async () => {
     try {
       const getData = await getTodos(USER_ID);
@@ -32,14 +31,14 @@ export const App: FC = () => {
       setTodos(getData);
     } catch {
       setError(ErrorType.LOAD);
-      setTimeout(() => setError(ErrorType.NONE), 3000);
+      closeErrorMessage(ErrorType.NONE);
     }
   };
 
   const addTodo = async (title: string) => {
     if (!title.trim()) {
       setError(ErrorType.EMPTY_TITLE);
-      setTimeout(() => setError(ErrorType.NONE), 3000);
+      closeErrorMessage(ErrorType.NONE);
 
       return;
     }
@@ -85,26 +84,24 @@ export const App: FC = () => {
     fetchTodos();
   }, []);
 
-  const handleDeleteTodo = useCallback(() => {
-    todos.forEach(todo => {
-      if (todo.completed) {
-        removeTodo(todo.id);
-      }
-    });
-  }, [todos]);
-
   const filteredTodos = useMemo(
     () => getFilteredTodos(todos, sortType),
     [todos, sortType],
   );
 
-  const activeTodosLength = useMemo(
+  const activeTodosCount = useMemo(
     () => todos.filter(todo => !todo.completed).length, [todos],
   );
 
-  const handleRemoveError = () => {
+  const handleRemoveError = useCallback(() => {
     setError(ErrorType.NONE);
-  };
+  }, []);
+
+  const handleClearCompleted = useCallback(() => {
+    todos.forEach(({ completed, id }) => {
+      return completed && removeTodo(id);
+    });
+  }, [todos]);
 
   if (!USER_ID) {
     return <UserWarning />;
@@ -118,7 +115,7 @@ export const App: FC = () => {
         <AddTodo
           onAddTodo={addTodo}
           onDisable={isDisabled}
-          activeTodosLength={activeTodosLength}
+          activeTodosCount={activeTodosCount}
         />
 
         <TodoList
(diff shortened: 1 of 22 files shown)


src/App.tsx
```diff
@@ -1,24 +1,148 @@
...
+  const fetchTodos = async () => {
+    try {
+      const getData = await getTodos(USER_ID);
+
+      setTodos(getData);
+    } catch {
+      setError(ErrorType.LOAD);
+      closeErrorMessage(ErrorType.NONE);
+    }
+  };
```
vladyslav-mosor, line 36: Better to use useCallback with such methods. Please fix it here and down below in your code

vladyslav-mosor (changes requested):
1. Your error message when titile is empty looks a little weird. It is better to use completely different sentence in this case.
image

2. You don`t show any messages when it is impossible to add or delete a todo. Plese double check it.
https://vimeo.com/816260054

3. Your notification error is not automatically hiding after 3 seconds.

vladyslav-mosor (approved):

src/App.tsx
```diff
@@ -1,24 +1,152 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import {
+  FC, useCallback,
+  useEffect,
+  useMemo,
+  useState,
```
qoa11a, line 5:
```suggestion
  FC,
  useCallback,
  useEffect,
  useMemo,
  useState,
```

src/App.tsx
```diff
@@ -1,24 +1,152 @@
...
+  const fetchTodos = useCallback(async () => {
+    try {
+      const getData = await getTodos(USER_ID);
+
+      setTodos(getData);
+    } catch {
+      setError(ErrorType.LOAD);
+      clearErrorMessage();
+    }
+  }, []);
```
qoa11a, line 40: image
Rather than wrapping it with useCallback just move this function inside the useEffect.
You can read more about usage of useCallback here: https://react.dev/reference/react/useCallback

src/App.tsx
```diff
@@ -1,24 +1,152 @@
...
+  const [isDisabled, setIsDisabled] = useState<boolean>(false);
+  const [loadTodoById, setLoadTodoById] = useState([DEFAULT_TASK_ID]);
+
+  const clearErrorMessage = () => {
+    setTimeout(() => setError(ErrorType.NONE), 3000);
+  };
+
+  const fetchTodos = useCallback(async () => {
+    try {
+      const getData = await getTodos(USER_ID);
```
qoa11a, line 33: The name `getData` looks like this is a function. Consider giving this variable a better name.

src/App.tsx
```diff
@@ -1,24 +1,152 @@
...
+      title,
+      completed: false,
+      userId: USER_ID,
+    };
+
+    try {
+      setIsDisabled(true);
+      setTempTodo({ id: DEFAULT_TASK_ID, ...newTodo });
+
+      const todo = await postTodo(newTodo as Todo);
```
qoa11a, line 60: It's not good that you say that it's of type `Todo` as it actually has no required id property. Instead, you could use `Omit<Todo, 'id'>` inside your `postTodo` function.

src/App.tsx
```diff
@@ -1,24 +1,152 @@
...
+    } catch {
+      setError(ErrorType.ADD);
+      clearErrorMessage();
+    } finally {
+      setTempTodo(null);
+      setIsDisabled(false);
+    }
+  }, []);
+
+  const removeTodo = useCallback(async (taskId: number) => {
```
qoa11a, line 72:
```suggestion
  const removeTodo = useCallback(async (todoId: number) => {
```

src/App.tsx
```diff
@@ -1,24 +1,152 @@
...
+import { Todo } from './types/Todo';
+import { ErrorType } from './types/Error';
+import { TaskStatus } from './types/Sort';
+import { deleteTodo, getTodos, postTodo } from './api/todos';
+import { AddTodo } from './components/AddTodo';
+import { TodoList } from './components/TodoList';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Filter } from './components/Filter';
+import { getFilteredTodos } from './utils/helpers';
+import { DEFAULT_TASK_ID, USER_ID } from './utils/constants';
```
qoa11a, line 17:
```suggestion
import { DEFAULT_TODO_ID, USER_ID } from './utils/constants';
```

src/App.tsx
```diff
@@ -1,24 +1,152 @@
...
+    fetchTodos();
+  }, []);
+
+  const filteredTodos = useMemo(
+    () => getFilteredTodos(todos, sortType),
+    [todos, sortType],
+  );
+
+  const activeTodosCount = useMemo(
+    () => todos.filter(todo => !todo.completed).length, [todos],
```
qoa11a, line 97:
```suggestion
    () => todos.filter(todo => !todo.completed).length,
    [todos],
```

src/api/todos.ts
```diff
@@ -0,0 +1,16 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+export const postTodo = (data: Todo) => {
```
qoa11a, line 8:
```suggestion
export const postTodo = (data: Omit<Todo, 'id'>) => {
```

src/components/AddTodo/AddTodo.tsx
```diff
@@ -0,0 +1,51 @@
+import {
+  ChangeEvent, FC, FormEvent, useState,
+} from 'react';
+import classNames from 'classnames';
+
+interface AddTodoProps {
+  onDisable: boolean;
+  onAddTodo: (query: string) => void;
+  activeTodosCount: number
+}
```
qoa11a, line 10: As you don't export it you could name it just `Props`.

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,33 @@
+import { FC } from 'react';
+import classNames from 'classnames';
+import { ErrorType } from '../../types/Error';
+
+interface NotificationProps {
+  error: ErrorType;
+  onRemoveError: () => void;
+}
```
qoa11a, line 8: Same here and elsewhere. If it's only local props it could be named `Props`.

qoa11a (changes requested):
Nicely done! Can be a little bit improved.

OleksiiBlezniuk (approved):

PR #377, opened 2023-04-11
Develop react_todo-app-add-and-delete

sergey-lipikhin:
[DEMO LINK](https://sergey-lipikhin.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
+      return;
+    }
+
+    const newTodo: TodoRich = {
+      id: 0,
+      title: todoTitle,
+      completed: false,
+      userId: USER_ID,
+      isLoading: true,
+    };
```
gretoq, line 77: I don't think it's a good idea to add the isLoading property to the Todo object type. Instead, it would be better to manage the component's state and props to control the loading logic. Modifying or creating new types just to add props can make the code more complicated and harder to maintain.

src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
+      .then(todosFromServer => setTodos(
+        todosFromServer.map(todo => (
+          {
+            ...todo,
+            isLoading: false,
+          }
+        )),
+      ))
+      .catch(() => setError(ErrorType.LoadTodos));
+  }, []);
```
gretoq, line 123: I believe that when using handlers with async/await or then()/catch(), it's best to choose one method to maintain consistency and avoid confusion.

src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
+
+  const filteredTodos = filterTodos(todos, selectedFilterOption);
+
+  const activeTodosCount = getActiveTodosCount(todos);
```
gretoq, line 127: You can create an universal function for getting count todos by status.

src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <TodoHeader onTodoAdd={handleTodoAdd} />
+
+        {todos.length > 0 && (
```
gretoq, line 141:
```suggestion
        {todos[0] && (
```

src/api/todos.ts
```diff
@@ -0,0 +1,14 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
```
gretoq, line 3: You can create endpoints for no creating magic strings

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+        <div className="modal overlay">
+          <div className="modal-background has-background-white-ter" />
+          <div className="loader" />
+        </div>
+      </div>
+
+      {tempTodo && (
+        <TodoItem
+          key={tempTodo.id}
```
gretoq, line 46: Why do we need key here?

src/common/helpers.ts
```diff
@@ -0,0 +1,23 @@
...
+        case TodoCompletionType.Active:
+          return !todo.completed;
+        default:
+          return true;
+      }
+    })
+);
+
+export const getActiveTodosCount = (todos: TodoRich[]): number => (
+  todos.filter(todo => todo.completed === false).length
```
gretoq, line 22:
```suggestion
  todos.filter(todo => !todo.completed).length
```

gretoq (changes requested):

Ambulence (approved):
When input any value appear one more form of input in your list. Hope you will fix it in the next task

PR #403, opened 2023-04-14
Develop

ifuride:
[DEMO LINK](https://ifuride.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,178 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
+import { getTodos, addTodo, deleteTodo } from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
 
-const USER_ID = 0;
+const USER_ID = 7033;
+
+function getVisibleTodos(todos: Todo[], currentFilter: string) {
+  let visibleTodos = [...todos];
+
+  if (currentFilter !== Filter.All) {
+    visibleTodos = visibleTodos.filter(({ completed }) => {
+      switch (currentFilter) {
+        case Filter.Active:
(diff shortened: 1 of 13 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,41 @@
...
+    event.preventDefault();
+
+    onSaveNewTodo();
+  };
+
+  return (
+    <header className="todoapp__header">
+      {/* this buttons is active only if there are some active todos */}
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button type="button" className="todoapp__toggle-all active" />
```
mykhalenych, line 27: The button should be conditionally rendered based on the presence of active todos. Right now it is always visible with a comment that suggests it should only be visible under certain conditions.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,96 @@
...
+};
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  filterBy,
+  onChangeFilter,
+  onDeleteCompletedTodos,
+}) => {
+  const activeTodos = [...todos].filter(todo => !todo.completed).length;
+  const completedTodos = [...todos].filter(todo => todo.completed).length;
```
mykhalenych, line 20: the same

src/App.tsx
```diff
@@ -1,24 +1,178 @@
...
+
+      clearError();
+    } finally {
+      setLoadingTodos([0]);
+    }
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, [todos]);
```
mykhalenych, line 127: Screenshot 2023-04-14 at 17 31 58
you have a cyclical dependency here due to todos

mykhalenych (changes requested):

src/components/Header/Header.tsx
ifuride: As far as I understood, this task requires just add and remove actions. Toggle all button is a part of the next task.

src/App.tsx
ifuride: Oops!

etojeDenys (changes requested):
almost done ✅

OleziO (approved):

PR #412, opened 2023-04-19
Develop

vladyslav-krainyk:
[DEMO LINK](https://vladyslav-krainyk.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,226 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { getTodos, addTodo, removeTodo } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoInput } from './components/TodoInput/TodoInput';
+import { FilterType } from './types/FilterType';
+import { Notification } from './components/Notification/Notification';
+import { ErrorType } from './types/ErrorType';
 
-const USER_ID = 0;
+const USER_ID = 7022;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<FilterType>(FilterType.All);
+  const [isError, setIsError] = useState<ErrorType>(ErrorType.None);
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,24 +1,225 @@
...
+
+  const activeTodos = useMemo(
+    () => todos.filter(todo => !todo.completed), [todos],
+  );
+
+  const completedTodos = useMemo(
+    () => todos.filter(todo => todo.completed), [todos],
+  );
+
+  const filteredTodos = () => {
```
loralevitska, line 52: wrap function to hook `useCallback`

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
+function wait(delay: number) {
+  return new Promise(resolve => {
+    setTimeout(resolve, delay);
+  });
+}
+
+// To have autocompletion and avoid mistypes
```
loralevitska, line 11: Remove all comments

loralevitska (changes requested):
Almost done)
Fix infinite loading todo (check hook dependencies)

https://user-images.githubusercontent.com/70009205/233124916-9cf09f12-397b-4cf0-8245-82f76445fcbf.mov
Also, try to memoize functions, especially those you pass into your components, because changing those functions or variables will cause the component to rerender
image

src/App.tsx
```diff
@@ -1,24 +1,226 @@
...
+
+  const activeTodos = useMemo(
+    () => todos.filter(todo => !todo.completed), [todos],
+  );
+
+  const completedTodos = useMemo(
+    () => todos.filter(todo => todo.completed), [todos],
+  );
+
+  const filteredTodos = useCallback(() => {
```
anstsot, line 52: better to use `useMemo` here. then you don;t have to call it like function
```suggestion
  const filteredTodos = useMemo(() => {
```

src/App.tsx
```diff
@@ -1,24 +1,226 @@
...
+              <button
+                type="button"
+                className="todoapp__clear-completed"
+                onClick={handleClearCompleted}
+              >
+                Clear completed
+              </button>
+            )}
+
+          </footer>
```
anstsot, line 214: you can move it to separate component `Footer`

anstsot (changes requested):
It’s almost done! Let’s improve a little
The loading isn't shown by adding new todo item

https://user-images.githubusercontent.com/22824431/233363815-b83c335c-6e5f-419c-a849-566f18d78001.mov

loralevitska (approved):
😊

PR #419, opened 2023-04-26
Solution by DK

DimaK123:
[DEMO LINK](https://Dimak123.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,6 +6,7 @@ import {
   getActiveTodos,
   getCompletedTodos,
   postTodos,
+  deleteTodos,
 } from './api/todos';
 
 import { Todo as TodoType } from './types/Todo';
@@ -27,21 +28,29 @@ export const App: React.FC = () => {
   const [inputValue, setInputValue] = useState('');
   const [isGetError, setGetError] = useState(false);
   const [isPostError, setPostError] = useState(false);
+  const [isDeleteError, setDeleteError] = useState(false);
   const [isInputEmpty, setEmptyInputState] = useState(false);
   const [isInputLocked, setLockInput] = useState(false);
   const [tempTodo, setTempTodo] = useState<string | null>(null);
+  const [isClearAllCompleted, setClearAllCompleted] = useState(false);
+
+  const handleErrorState
+  = (setErrorFunc: (state: boolean) => void, state = true) => {
+    setError(state);
+    setErrorFunc(state);
+
+    if (state === false) {
+      return;
+    }
+
+    setTimeout(() => handleErrorState(setErrorFunc, !state), 3000);
+  };
 
   useEffect(() => {
     getTodos(USER_ID)
       .then(result => setTodos(result))
       .catch(() => {
-        setError(true);
-        setGetError(true);
-
-        setTimeout(() => {
-          setError(false);
-          setGetError(false);
-        }, 2000);
+        handleErrorState(setGetError);
       });
   }, [todos]);
 
@@ -81,26 +90,14 @@ export const App: React.FC = () => {
 
     if (key === 'Enter') {
       if (!inputValue) {
-        setEmptyInputState(true);
-        setError(true);
-
-        setTimeout(() => {
-          setError(false);
-          setEmptyInputState(false);
-        }, 2000);
+        handleErrorState(setEmptyInputState);
       } else {
         setLockInput(true);
         setTempTodo(inputValue);
 
         postTodos(USER_ID, inputValue)
           .catch(() => {
-            setError(true);
-            setPostError(true);
-
-            setTimeout(() => {
-              setError(false);
-              setPostError(false);
-            }, 2000);
+            handleErrorState(setPostError);
           })
           .finally(() => {
             setLockInput(false);
@@ -108,6 +105,22 @@ export const App: React.FC = () => {
             setInputValue('');
           });
       }
+    }
+  };
+
+  const deleteAllCompleted = async () => {
+    try {
+      setClearAllCompleted(true);
+
+      const arrayOfCompletedTodos
+      = await getCompletedTodos(USER_ID);
+
+      const deletePromises
+      = arrayOfCompletedTodos.map(todo => deleteTodos(todo.id));
+
+      await Promise.all(deletePromises);
+    } finally {
+      setClearAllCompleted(false);
     }
   };
 
@@ -119,7 +132,7 @@ export const App: React.FC = () => {
         <header className="todoapp__header">
           <button type="button" className="todoapp__toggle-all active" />
 
-          <form>
+          <form onSubmit={(event) => event.preventDefault()}>
             <input
               type="text"
               className="todoapp__new-todo"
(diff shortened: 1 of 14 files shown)


src/components/Filter/FilterParams.ts
```diff
@@ -0,0 +1,5 @@
+export enum FilterParams {
+  all = 'all',
+  completed = 'completed',
+  active = 'active',
```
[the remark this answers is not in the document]
DimaK123: All done, but I can't figure out how to make server requests not constant, but only when changes are made on the page?

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+    state = true,
+  ) => {
+    setError(state);
+    setErrorFunc(state);
+
+    if (state === false) {
+      return;
+    }
+
+    setTimeout(() => handleErrorState(setErrorFunc, !state), 3000);
```
Anastasiia-Svintsova, line 48: You should clear all timers on the component unmount 
https://bobbyhadz.com/blog/react-cleartimeout

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+  const [isClearAllCompleted, setClearAllCompleted] = useState(false);
+
+  const handleErrorState = (
+    setErrorFunc: (state: boolean) => void,
+    state = true,
+  ) => {
+    setError(state);
+    setErrorFunc(state);
+
+    if (state === false) {
```
Anastasiia-Svintsova, line 44: You can use logical operators instead
```suggestion
    if (!state) {
```

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+  const [isInputLocked, setLockInput] = useState(false);
+  const [tempTodo, setTempTodo] = useState<string | null>(null);
+  const [isClearAllCompleted, setClearAllCompleted] = useState(false);
+
+  const handleErrorState = (
+    setErrorFunc: (state: boolean) => void,
+    state = true,
+  ) => {
+    setError(state);
+    setErrorFunc(state);
```
Anastasiia-Svintsova, line 42: Why do you need 2 identical states? Could you refactor it and leave only 1?

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+    }
+
+    setTimeout(() => handleErrorState(setErrorFunc, !state), 3000);
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(result => setTodos(result))
+      .catch(() => {
+        handleErrorState(setGetError);
```
Anastasiia-Svintsova, line 55: Why do you need to pass the function here? Just set the error inside the catch

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+      });
+  }, [todos]);
+
+  useEffect(() => {
+    getActiveTodos(USER_ID).then(result => setActiveTodos(result.length));
+
+    getCompletedTodos(USER_ID).then(result => setCompletedTodos(result.length));
+  }, [filterParam, todos]);
```
Anastasiia-Svintsova, line 63: Not the best way to get values. Use useMemo() instead

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const visibleTodos = todos.filter(todo => {
```
Anastasiia-Svintsova, line 69: Use useMemo() here

src/components/TempTodo/TempTodo.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+type Props = {
+  title: string
+};
+
+export const TempTodo:React.FC<Props> = ({ title }) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Anastasiia-Svintsova, line 8: Don't disable eslint

Anastasiia-Svintsova (changes requested):
- The user shouldn't be able to add a todo with an empty title. Fix it, please 

https://user-images.githubusercontent.com/99403198/235292666-3932c3f9-f742-43e0-9049-dc40147bf951.mov

- Look, here is infinite fetching. You should fetch it at the beginning and then only after changes. It really decreases the productivity 

https://user-images.githubusercontent.com/99403198/235292740-99f5fa8c-7220-4dad-b1b4-4a54f13f0e75.mov

DimaK123:
> * The user shouldn't be able to add a todo with an empty title. Fix it, please
> 
> 
> Screen.Recording.2023-04-29.at.11.21.10.mov
> 
>     * Look, here is infinite fetching. You should fetch it at the beginning and then only after changes. It really decreases the productivity
> 
> 
> Screen.Recording.2023-04-29.at.11.23.02.mov

So, I've fixed

src/App.tsx
DimaK123: I don't know how to do it using useEffect, so I implemented the timer deletion in a function.

src/App.tsx
DimaK123: Done

src/App.tsx
DimaK123: what do you mean?
I use only one state param, and pass it to both functions.

src/App.tsx
DimaK123: i need pass correct error set-function to the handleErrorState function to show message 'Error, can\'t get todos from server', so i use this function

src/App.tsx
DimaK123: Done

src/App.tsx
DimaK123: i can't do it cause the basic value for the visibleTodos it empty array, so if i try to use useMemo here, i get error from typescript

src/components/TempTodo/TempTodo.tsx
DimaK123: Done

DimaK123:
> * The user shouldn't be able to add a todo with an empty title. Fix it, please
> 
> 
> Screen.Recording.2023-04-29.at.11.21.10.mov
> 
>     * Look, here is infinite fetching. You should fetch it at the beginning and then only after changes. It really decreases the productivity
> 
> 
> Screen.Recording.2023-04-29.at.11.23.02.mov

fixed

src/App.tsx
OleziO: You don't need two states for error, just do it with one state.

src/api/todos.ts
```diff
@@ -0,0 +1,28 @@
...
+    title,
+    completed: false,
+  });
+};
+
+export const deleteTodos = (id: number) => {
+  return client.delete(`/todos/${id}`);
+};
+
+// Add more methods here
```
OleziO, line 28: Remove all comments from the code.

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,77 @@
...
+}) => {
+  const { completed, title, id } = todoItem;
+
+  const [isLoading, setLoading] = useState(false);
+
+  const handleDeleteTodos = (todoId: typeof id) => {
+    setLoading(true);
+
+    deleteTodos(todoId)
+      .then(() => setLoading(false))
```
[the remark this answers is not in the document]
OleziO: Not resolved.

OleziO (changes requested):
Delete does not work.

src/api/todos.ts
DimaK123: Done

polosanya (changes requested):
It is still not possible to remove a todo.
Items left count also doesn't work correctly
todos

DimaK123:
> It is still not possible to remove a todo. Items left count also doesn't work correctly todos [ todos ](https://user-images.githubusercontent.com/96409563/235422391-19a4225b-992b-4928-9d9c-cf07ced8c188.gif) [ ](https://user-images.githubusercontent.com/96409563/235422391-19a4225b-992b-4928-9d9c-cf07ced8c188.gif)

it's seems i fogot to run 'npm run deploy'

PR #427, opened 2023-05-04
solution

PashaMalyshkin:
[DEMO LINK](https://PashaMalyshkin.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,194 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
+import { Category } from './utils/Category';
+import { Error } from './utils/Error';
 import { UserWarning } from './UserWarning';
+import { TodoForm } from './Components/TodoForm';
+import { TodoList } from './Components/TodoList';
+import { TodoFilter } from './Components/TodoFilter';
+import { addTodo, getTodos, removeTodo } from './api/todos';
 
-const USER_ID = 0;
+const USER_ID = 10156;
 
 export const App: React.FC = () => {
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [isError, setIsError] = useState<Error | string>('');
+  const [filterCategory, setFilterCategory] = useState(Category.All);
+  const [newTodo, setNewTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const todosAmount = visibleTodos.length;
+  const hasCompleted = visibleTodos.some(todo => todo.completed);
+
+  const handleError = (error: Error) => {
+    setIsError(error);
+    window.setTimeout(() => {
+      setIsError('');
+    }, 3000);
+  };
+
+  const loadTodos = async () => {
+    try {
+      const todos = await getTodos(USER_ID);
+
+      setVisibleTodos(todos);
+    } catch {
+      handleError(Error.LOAD);
+    }
+  };
+
+  const filterTodos = async (category: Category) => {
+    const allTodos = await getTodos(USER_ID);
+
+    const filteredTodos = allTodos.filter(todo => {
+      switch (category) {
+        case Category.Active:
+          return !todo.completed;
+        case Category.Completed:
+          return todo.completed;
+        default:
+          return true;
+      }
+    });
+
+    setVisibleTodos(filteredTodos);
+  };
+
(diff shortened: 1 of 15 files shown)


prokivskaaa (approved):
Good, but maybe consider adding loader on delete

prokivskaaa (approved):

src/utils/Error.ts
```diff
@@ -0,0 +1,7 @@
+export enum Error {
```
ManoylovAC, line 1: `Error` name for will conflict with global Error object so you can have some unpredictable behaviour. Better rename it.

src/utils/Error.ts
```diff
@@ -0,0 +1,7 @@
...
+  LOAD = 'Unable to load a todo',
+  DELETE = 'Unable to delete a todo',
```
ManoylovAC, line 3: You can make your error names more specific
```suggestion
  UNABLE_DELETE = 'Unable to delete a todo',
```

src/Components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,33 @@
...
+  todos: Todo[],
+  onDelete: (id: number) => void;
+  newTodo: Todo | null;
+}
+export const TodoList: React.FC<Props> = ({
+  todos,
+  onDelete,
+  newTodo,
+}) => {
+  return (
```
ManoylovAC, line 15: you can use implicit return

src/Components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoItem } from '../TodoItem';
+
+interface Props {
+  todos: Todo[],
+  onDelete: (id: number) => void;
+  newTodo: Todo | null;
```
ManoylovAC, line 8: Not sure why you need `newTodo`...
After creating and adding a new todo item, you should add it to the main `todos` list to make sure that all new items appear in the list, not just the latest ones.
I think you have a problem now when you try to add multiple todo items in a row - please check this to understand the problem live and fix it

src/Components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface Props {
+  filterCategory: Category;
+  changeCategory: (category: Category) => void;
+}
+
+export const TodoFilter: React.FC<Props> = ({
+  filterCategory,
+  changeCategory,
+}) => {
+  return (
```
ManoylovAC, line 14: you can use implicit return

src/App.tsx
```diff
@@ -1,24 +1,195 @@
...
+    }
+  };
+  // const handleTodoUpdate = async (todoId: number, title: string) => {
+  //   try {
+  //     await updateTodo(todoId, title);
+  //   } catch {
+  //     handleError(Error.UPDATE);
+  //   }
+  // };
+
```
ManoylovAC, line 128: if you don't need this code, you can remove it or add some explanatory comment

src/App.tsx
```diff
@@ -1,24 +1,195 @@
...
+    if (!todoTitle) {
+      handleError(Error.TITLE);
+
+      return;
+    }
+
+    try {
+      const todo: Todo = {
+        title: todoTitle,
+        id: 0,
```
ManoylovAC, line 75: I think is no need to hardcode here
the id will be generated on the server side

src/App.tsx
```diff
@@ -1,24 +1,195 @@
...
+  useEffect(() => {
+    filterTodos(filterCategory);
+  }, [filterCategory]);
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  const handleAdd = async () => {
+    if (!todoTitle) {
```
ManoylovAC, line 66: please add validation to the empty line
because the name can contain a lot of spaces " "

src/Components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,37 @@
...
+  };
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <input
+        type="text"
+        className="todoapp__new-todo"
+        placeholder="What needs to be done?"
+        value={title}
+        onChange={(event) => onChange(event.target.value)}
```
ManoylovAC, line 32: you can actually store the value for that input in that component state and send the todo-name to the parent component after submitting

ManoylovAC (changes requested):
Nice start and now time to make some improvements.
Check out the comments I left and do you best to improve your application.
Pay special attention to the current logic for adding todos - it is working a bit incorrectly.
Good luck!

jstmpelowycz (approved):

PR #429, opened 2023-05-05
solution

toljan1:
[DEMO LINK](https://toljan1.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,104 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { addTodoToServer, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/todoList';
+import { TodoForm } from './components/todoForm';
+import { TodoFooter } from './components/todoFooter';
+import { Notification } from './components/notification';
+import { FilterBy } from './types/FilterBy';
 
-const USER_ID = 0;
+const USER_ID = 10221;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+  const [title, setTitle] = useState<string | undefined>();
+  const [todoWasDeleted, setTodoWasDeleted] = useState(false);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((fetchedTodos: Todo[]) => {
+        setTodos(fetchedTodos);
+        setTodoWasDeleted(false);
+      })
+      .catch((fetchedError) => {
+        setError(
+          fetchedError?.message
+            ? fetchedError.message
+            : 'Something went wrong',
+        );
+
+        setTodoWasDeleted(false);
+      });
+  }, [todoWasDeleted === true]);
+
+  useEffect(() => {
+    if (title) {
+      addTodoToServer(USER_ID, {
+        userId: USER_ID,
+        title,
+        completed: false,
+      })
+        .then((fetchedTodo: Todo) => {
+          setTodos([...todos, fetchedTodo]);
+          setTitle('');
+        })
+        .catch((fetchedError) => {
+          setError(
+            fetchedError?.message
+              ? fetchedError.message
+              : 'Something went wrong',
+          );
+        });
+    }
+  }, [title]);
+
+  function getFilterCallback(filterType: FilterBy) {
+    switch (filterType) {
+      case FilterBy.ACTIVE:
+        return (todo: Todo) => !todo.completed;
+
+      case FilterBy.COMPLETED:
+        return (todo: Todo) => todo.completed;
+
+      default:
+        return (todo: Todo) => !!todo.id;
+    }
+  }
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <TodoForm addTodo={setTitle} />
+
+        {todos.length ? (
+          <>
+            <TodoList
+              todos={todos.filter(getFilterCallback(filterBy))}
+              setError={setError}
+              setTodoWasDeleted={setTodoWasDeleted}
+            />
+            <TodoFooter
+              setFilterBy={setFilterBy}
+              itemsQuantity={todos.filter(getFilterCallback(filterBy)).length}
+            />
+          </>
+        ) : null}
+      </div>
+
+      <Notification
+        onClose={(value: string | null) => setError(value)}
+        error={error}
+      />
+    </div>
   );
 };
--- a/src/components/notification.tsx
+++ b/src/components/notification.tsx
@@ -1,0 +1,31 @@
+import classNames from 'classnames';
+
+interface Props {
+  onClose: (value: string | null) => void,
+  error: string | null,
+}
+
+export const Notification: React.FC<Props> = ({ onClose, error }) => {
+  return (
+    <div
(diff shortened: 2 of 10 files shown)


src/App.tsx
```diff
@@ -1,24 +1,104 @@
...
+      case FilterBy.ACTIVE:
+        return (todo: Todo) => !todo.completed;
+
+      case FilterBy.COMPLETED:
+        return (todo: Todo) => todo.completed;
+
+      default:
+        return (todo: Todo) => !!todo.id;
+    }
+  }
```
IvanFesenko, line 70: No need to filter in case FilterBy is All

src/App.tsx
```diff
@@ -1,24 +1,104 @@
...
+        {todos.length ? (
+          <>
+            <TodoList
+              todos={todos.filter(getFilterCallback(filterBy))}
+              setError={setError}
+              setTodoWasDeleted={setTodoWasDeleted}
+            />
+            <TodoFooter
+              setFilterBy={setFilterBy}
+              itemsQuantity={todos.filter(getFilterCallback(filterBy)).length}
```
IvanFesenko, line 92: Create separate variable with visible todos instead filtering 2 times

src/components/todoFooter.tsx
```diff
@@ -0,0 +1,72 @@
...
+      id: 'ALL',
+    },
+    {
+      title: 'Active',
+      id: 'ACTIVE',
+    },
+    {
+      title: 'Completed',
+      id: 'COMPLETED',
+    }];
```
IvanFesenko, line 23: Move const outside component. Use FilterBy enam instead hardcoding id

src/components/todoFooter.tsx
```diff
@@ -0,0 +1,72 @@
...
+                  setFilterBy(FilterBy.COMPLETED);
+                  break;
+
+                default:
+                  setFilterBy(FilterBy.ALL);
+                  break;
+              }
+
+              setSelectedType(navItem.id);
+            }}
```
IvanFesenko, line 60: Do not use anonymous function - create separate handler

src/App.tsx
```diff
@@ -1,24 +1,104 @@
...
+import { FilterBy } from './types/FilterBy';
 
-const USER_ID = 0;
+const USER_ID = 10221;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+  const [title, setTitle] = useState<string | undefined>();
```
IvanFesenko, line 18: Reduadnt state. You have similar in `TodoForm`. Remove it here

src/App.tsx
```diff
@@ -1,24 +1,104 @@
...
+        })
+        .catch((fetchedError) => {
+          setError(
+            fetchedError?.message
+              ? fetchedError.message
+              : 'Something went wrong',
+          );
+        });
+    }
+  }, [title]);
```
IvanFesenko, line 57: Redudant

src/components/todoForm.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <header className="todoapp__header">
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button type="button" className="todoapp__toggle-all active" />
+
+      <form
+        onSubmit={(ev) => {
+          ev.preventDefault();
+          addTodo(titleValue);
+          setTitleValue('');
+        }}
```
IvanFesenko, line 20: Create separate handler

src/components/todoForm.tsx
```diff
@@ -0,0 +1,32 @@
...
+      >
+        <input
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={titleValue}
+          onChange={(ev) => setTitleValue(ev.currentTarget.value)}
```
IvanFesenko, line 27: Same here

src/components/todoList.tsx
```diff
@@ -0,0 +1,65 @@
...
+                  .then(() => setTodoWasDeleted(true))
+                  .catch((fetchedError) => {
+                    setError(
+                      fetchedError?.message
+                        ? fetchedError.message
+                        : 'Something went wrong',
+                    );
+                  });
+              }
+            }}
```
IvanFesenko, line 52: Separate handler as well

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
+function wait(delay: number) {
+  return new Promise(resolve => {
+    setTimeout(resolve, delay);
+  });
+}
+
+// To have autocompletion and avoid mistypes
```
OleziO, line 11: Remove all comments from the code.

OleziO (changes requested):

src/utils/fetchClient.ts
OleziO: Not resolved

OleziO (changes requested):

src/utils/fetchClient.ts
toljan1: sorry, i didn`t see this comment. All others are deleted

src/components/todoList.tsx
```diff
@@ -0,0 +1,100 @@
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+
+interface Props {
+  todos: Todo[],
+  tempTodo: Todo | null,
+  idProcessed: number,
+  disableList: boolean,
+  onDelete: (todoId: number | undefined) => () => void,
```
etojeDenys, line 9: The type for todoId should be 'number' instead of 'number | undefined'.

src/components/todoList.tsx
```diff
@@ -0,0 +1,100 @@
...
+              disabled={disableList}
+            />
+          </label>
+
+          <span className="todo__title">{todo.title}</span>
+
+          <button
+            type="button"
+            className="todo__remove"
+            onClick={onDelete(todo.id)}
```
etojeDenys, line 46: Instead of passing a function returning another function, simply pass the function and use currying to accept the 'todo.id'.

src/components/notification.tsx
```diff
@@ -0,0 +1,33 @@
...
+        'is-light',
+        'has-text-weight-normal',
+        {
+          hidden: !error,
+        },
+      )}
+    >
+      <button
+        type="button"
+        aria-label={error || ''}
```
etojeDenys, line 25: Instead of using 'error || ''', use 'error ?? '''. The nullish coalescing operator (??) is more appropriate when dealing with null or undefined values.

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+      .catch((fetchedError) => {
+        setError(
+          fetchedError?.message
+            ? fetchedError.message
+            : 'Something went wrong',
+        );
+
+        setTodoWasDeleted(false);
+      });
+  }, [todoWasDeleted === true]);
```
etojeDenys, line 58: The dependency array in this useEffect should contain only dependencies, not conditions. Consider changing it to [todoWasDeleted] rather than [todoWasDeleted === true].

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+        })
+        .catch(() => {
+          setTempTodo(null);
+          setDisableForm(false);
+          setError('Unable to add a todo');
+        });
+    }
+  };
+
+  const onDelete = (todoId: number | undefined) => () => {
```
etojeDenys, line 90: Consider changing the onDelete function to take only the todoId as an argument, and not return another function. Then, change its usage accordingly.

src/components/todoList.tsx
etojeDenys, line 46:
```suggestion
            onClick={() => onDelete(todo.id)}
```

src/types/Todo.ts
```diff
@@ -0,0 +1,6 @@
+export interface Todo {
+  id?: number;
```
etojeDenys, line 2: does id can be optional?

etojeDenys (changes requested):

src/components/todoList.tsx
```diff
@@ -0,0 +1,104 @@
...
+            'is-active': idProcessed === todo.id,
+          })}
+          >
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
+      ))}
+
+      {tempTodo !== null ? (
```
OleziO, line 66: Replace `tempTodo !== null` with `tempTodo`, it will be the same.

OleziO (changes requested):

OleziO (approved):

PR #433, opened 2023-05-08
solution

Maks-Tsarenko:
[DEMO LINK](https://Maks-Tsarenko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,173 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useEffect, useState, useCallback, FormEvent,
+} from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Notification } from './components/Notification';
+import { createTodo, deleteTodo, getTodos } from './api/todos';
+import { TodoStatus } from './types/TodoStatus';
+import { ErrorType } from './types/ErrorType';
 
-const USER_ID = 0;
+export const USER_ID = 9944;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<TodoStatus>(TodoStatus.All);
+  const [isError, setIsError] = useState(false);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [errorType, setErrorType] = useState<ErrorType>(ErrorType.NoError);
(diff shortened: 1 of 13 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
...
+    title,
+    completed: false,
+  });
+};
+
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+// Add more methods here
```
OleziO, line 20: Remove all comments from the code.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
```
OleziO, line 4: Remove all comments from the code.

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
 
-const USER_ID = 0;
+export const USER_ID = 9944;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<TodoStatus>(TodoStatus.All);
+  const [isError, setIsError] = useState(false);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [errorType, setErrorType] = useState<ErrorType>(ErrorType.NoError);
```
anstsot, line 21: `isError` variable can be calculated based on `errorType`
```suggestion
  const [newTodoTitle, setNewTodoTitle] = useState('');
  const [errorType, setErrorType] = useState<ErrorType>(ErrorType.NoError);
  const isError = errorType !== ErrorType.NoError;
```

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
+      setTodos(newTodoList);
+    } catch {
+      setIsError(true);
+      setErrorType(ErrorType.AddError);
+      setTimeout(() => {
+        setIsError(false);
+      }, 3000);
+    }
+  };
+
```
anstsot, line 39: try to use `useEffect` hook instead of calling setTimeout for each api call
```suggestion
    }
  };

  useEffect(() => {
    if (isError) {
      setTimeout(() => {
        setIsError(false);
      }, 3000);
    }
  }, [isError]);
```

src/App.tsx
```diff
@@ -1,24 +1,173 @@
...
+  useEffect(() => {
+    getTodoList();
+  }, []);
+
+  const filteredTodos = useCallback(() => {
```
anstsot, line 44: you can use `useMemo` to save the filtered todos in variable
```suggestion
  const filteredTodos = useMemo(() => {
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

PR #437, opened 2023-05-10
Develop

IhorChaikovskyi:
[DEMO LINK](https://IhorChaikovskyi.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,8 +3,8 @@ import React, { useState, useEffect, useMemo } from 'react';
 import { UserWarning } from './UserWarning';
 import { Todo } from './types/Todo';
 import { Select } from './types/Select';
-import { getTodos } from './api/todos';
-import { Main } from './components/Main';
+import { getTodos, postDelete } from './api/todos';
+import { TodoList } from './components/TodoList';
 import { Footer } from './components/Footer';
 import { Notification } from './components/Notification';
 import { Errors } from './types/Errors';
@@ -16,12 +16,11 @@ export const App: React.FC = () => {
   const [todoList, setTodoList] = useState<Todo[] | null>(null);
   const [selectedFilter, setSelectedFilter] = useState(Select.ALL);
   const [typeError, setTypeError] = useState<Errors | null>(null);
-  const [notificationError, setNotificationError] = useState(false);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
-  const [loadersTodosId, setLoadersTodoId] = useState<number[] | null>(null);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
 
-  const filteringList = useMemo(() => {
-    const filter = todoList && todoList.filter((todo) => {
+  const filteredList = useMemo(() => {
+    return todoList && todoList.filter((todo) => {
       switch (selectedFilter) {
         case Select.ACTIVE:
           return !todo.completed;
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,24 +1,114 @@
...
+          return todo.completed;
+        default:
+          return true;
+      }
+    });
+
+    return filter;
+  }, [selectedFilter, todoList]);
+
+  const countItemLeft = () => {
```
TarasHoliuk, line 38: instead of creating such function I'd save it into a variable and count it in this way:

todos.filter(todo => !todo.completed).length

src/App.tsx
```diff
@@ -1,24 +1,114 @@
...
+import { Header } from './components/Header';
 
-const USER_ID = 0;
+export const USER_ID = 9960;
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[] | null>(null);
+  const [selectedFilter, setSelectedFilter] = useState(Select.ALL);
+  const [typeError, setTypeError] = useState<Errors | null>(null);
+  const [notificationError, setNotificationError] = useState(false);
```
TarasHoliuk, line 19: you may get rid of `notificationError` state - just check typeError: `!!typeError === notificationError` in all cases

src/App.tsx
```diff
@@ -1,24 +1,114 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadersTodosId, setLoadersTodoId] = useState<number[] | null>(null);
```
TarasHoliuk, line 21: in the case when array is stored in the state set empty array as the default value
```suggestion
  const [loadingTodoIds, setLoadingTodoIds] = useState<number[] | null>([]);
```
also, `set` function name should always match the state variable name

src/App.tsx
```diff
@@ -1,24 +1,114 @@
...
+
+  const filteringList = useMemo(() => {
+    const filter = todoList && todoList.filter((todo) => {
```
TarasHoliuk, line 24:
```suggestion
    return todoList && todoList.filter((todo) => {
```

src/App.tsx
```diff
@@ -1,24 +1,114 @@
...
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[] | null>(null);
+  const [selectedFilter, setSelectedFilter] = useState(Select.ALL);
+  const [typeError, setTypeError] = useState<Errors | null>(null);
+  const [notificationError, setNotificationError] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadersTodosId, setLoadersTodoId] = useState<number[] | null>(null);
+
+  const filteringList = useMemo(() => {
```
TarasHoliuk, line 23:
```suggestion
  const filteredList = useMemo(() => {
```

src/App.tsx
```diff
@@ -1,24 +1,114 @@
...
+          setTypeError={setTypeError}
+          setNotificationError={setNotificationError}
+          todoList={todoList}
+          setTempTodo={setTempTodo}
+          setTodoList={setTodoList}
+        />
+
+        {todoList && (
+          <>
+            <Main
```
TarasHoliuk, line 80:
```suggestion
            <TodoList
```
it's more descriptive

src/App.tsx
```diff
@@ -1,24 +1,114 @@
...
+          counterItemLeft={countItemLeft}
+          setTypeError={setTypeError}
+          setNotificationError={setNotificationError}
+          todoList={todoList}
+          setTempTodo={setTempTodo}
+          setTodoList={setTodoList}
+        />
+
+        {todoList && (
+          <>
```
TarasHoliuk, line 79: fragment is redundant here

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+  setTypeError,
+  setNotificationError,
+  todoList,
+  setTodoList,
+  loadersTodosId,
+}) => {
+  const [loaderTodo, setLoaderTodo] = useState(false);
+  const { id, title, completed } = todo || { id: 0 };
+
+  const deleteClickHandler = () => {
```
TarasHoliuk, line 27: instead of passing all props which are needed to create this function to `TodoItem` (but directly related to a separate item) component (e.g. `setTypeError`, `setTodoList`) create this function in `App.tsx` and pass it as prop

The same with `Footer` component

TarasHoliuk (changes requested):
GJ! Let's improve it a bit

OleziO (approved):

src/App.tsx
```diff
@@ -1,24 +1,137 @@
...
+        {todoList && (
+          <TodoList
+            filteringList={filteredList}
+            tempTodo={tempTodo}
+            loadingTodoIds={loadingTodoIds}
+            deleteClickHandler={deleteClickHandlerItem}
+          />
+        )}
+
+        {(todoList?.length !== 0 && todoList) && (
```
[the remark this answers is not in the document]
IhorChaikovskyi: It does not work because 0 remains
image

PR #446, opened 2023-05-12
Create add and delete function

AnyaLyaLya:
[DEMO LINK](https://AnyaLyaLya.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,268 @@
+/* eslint-disable jsx-a11y/mouse-events-have-key-events */
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { getTodos, addTodo, deleteTodo } from './api/todos';
+// eslint-disable-next-line import/extensions
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+const USER_ID = 10305;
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,24 +1,268 @@
...
+const USER_ID = 10305;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = React.useState<Todo[]>([]);
+  const [isHovered, setIsHovered] = React.useState(false);
+  const [filter, setFilter] = React.useState('all');
+  const [isLoading, setIsLoading] = React.useState(true);
+  const [isEditing, setIsEditing] = React.useState(false);
+  const [error, setError] = React.useState<string | null>(null);
+  const [title, setTitle] = React.useState<string>('');
```
Kavon-Skela, line 20: I think that type definition here is extra

Kavon-Skela (changes requested):
You've almost reached the goal. Nice job! But it is some bugs, that I found:
1. I can't see a delete button
2. Your filters are not centered
3. There is no loader during loading todo
4. Error message should have red color

ManoylovAC (changes requested):
Hi there 👋 
Sorry I can't see and can't use your `Delete` feature.
This is required part of the task so please check your markup or/and code and fix it.
image

Also it will be good to split application into smaller components such as Footer, Header, TodoForm or something similar. In real projects, decomposition is an important part of the work.

Good luck!

SerhiiKirik (approved):
Nice 🔥

PR #448, opened 2023-05-12
Solution

Kavon-Skela:
[DEMO LINK](https://Kavon-Skela.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,368 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  FormEvent,
+  useCallback,
+  useEffect,
+  useState,
+} from 'react';
+import cn from 'classnames';
+import { getAllTodos } from './api/todos';
+import { client } from './utils/fetchClient';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
+import { Filters, Errors } from './utils/enums';
 
-const USER_ID = 0;
+const USER_ID = 10306;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedFilter, setSelectedFilter] = useState('All');
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState(Errors.NoErrors);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingTodoId, setDeletingTodoId] = useState<number | null>(null);
+
+  const delayErrorDisappear = useCallback(() => {
+    setTimeout(() => {
+      setHasError(false);
+    }, 3000);
+  }, []);
+
+  useEffect(() => {
+    getAllTodos(USER_ID)
+      .then(setTodos);
+  }, []);
+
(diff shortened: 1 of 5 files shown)


src/utils/enums.ts
```diff
@@ -0,0 +1,13 @@
+export enum Filters {
+  All = 'All',
+  Active = 'Active',
+  Completed = 'Completed',
+}
+
+export enum Errors {
```
liliia-livandovskaya, line 7: Consider renaming `Errors` enum to a more specific name like `TodoErrors` or `AppErrors` to avoid confusion with other error enums in the future.

src/App.tsx
```diff
@@ -1,24 +1,368 @@
...
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
+import { Filters, Errors } from './utils/enums';
 
-const USER_ID = 0;
+const USER_ID = 10306;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedFilter, setSelectedFilter] = useState('All');
```
liliia-livandovskaya, line 19: Consider using `useState<Filters>(Filters.All)` for the selectedFilter state so that its type is tied to the Filters enum.

src/App.tsx
```diff
@@ -1,24 +1,368 @@
...
+      case Filters.All:
+        return true;
+
+      case Filters.Active:
+        return !completed;
+
+      case Filters.Completed:
+        return completed;
+
+      default:
```
liliia-livandovskaya, line 130: Change `return 0;` to `return false;` to match the expected return type of a boolean in the filter function.

src/App.tsx
```diff
@@ -1,24 +1,368 @@
...
+  useEffect(() => {
+    getAllTodos(USER_ID)
+      .then(setTodos);
+  }, []);
+
+  const handleFilterSelection = useCallback((value: string) => {
+    setSelectedFilter(value);
+  }, [selectedFilter]);
+
+  const handleChangeInputValue = useCallback((value: string) => {
```
liliia-livandovskaya, line 42: Remove `[todoTitle]` from the useCallback dependencies array since it is not used inside the callback.

src/App.tsx
```diff
@@ -1,24 +1,368 @@
...
+      setHasError(false);
+    }, 3000);
+  }, []);
+
+  useEffect(() => {
+    getAllTodos(USER_ID)
+      .then(setTodos);
+  }, []);
+
+  const handleFilterSelection = useCallback((value: string) => {
```
liliia-livandovskaya, line 38: Remove `[selectedFilter]` from the useCallback dependencies array since it is not used inside the callback.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
liliia-livandovskaya, line 1: Avoid using 'any' type. You can use 'unknown' as a safer alternative.

src/App.tsx
```diff
@@ -1,24 +1,368 @@
...
+            <button type="button" className="todo__remove">×</button>
+
+            <div className="modal overlay">
+              <div className="modal-background has-background-white-ter" />
+              <div className="loader" />
+            </div>
+          </div> */}
+
+          {/* This todo is being edited */}
+          {/* <div className="todo">
```
liliia-livandovskaya, line 245: remove comments

liliia-livandovskaya (approved):

PR #454, opened 2023-05-15
created context, add ability to add and delete todos

Zad0rnik:


ymys1 (changes requested):
Sorry, but I cant find your DEMO link, so I cant test it
 Code looks good, but please add DEMO

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+      return;
+    }
+
+    setIsCreating(true);
+
+    const newTodo = {
+      id: 0,
+      userId: 10407,
+      title: todoTitle,
```
Pozdnya, line 29: U can create another file for important variables and use it with export. 
Oh sorry, I see this variable in App.tsx

src/App.tsx
```diff
@@ -1,24 +1,66 @@
...
-    return <UserWarning />;
-  }
+  const {
+    todos,
+    error,
+    setTodos,
+    setError,
+  } = useTodoContext();
+
+  const loadTodos = async () => {
```
Pozdnya, line 21: Implement useCallback for this function

Pozdnya (approved):
Make sense using useContext, good job!!!

src/components/FooterFilter.tsx
```diff
@@ -0,0 +1,49 @@
...
+          className={cn('filter__link', {
+            selected: filter === Filter.ALL,
+          })}
+          onClick={() => setFilter(Filter.ALL)}
+        >
+          All
+        </a>
+
+        <a
+          href="#/active"
```
liliia-livandovskaya, line 22: Prevent default behavior on link click to avoid unnecessary page reloads.

src/App.tsx
```diff
@@ -1,24 +1,66 @@
...
+    try {
+      setError(null);
+      const todosFromServer = await getTodos(USER_ID);
+
+      setTodos(todosFromServer);
+    } catch {
+      setError(Error.LOAD);
```
liliia-livandovskaya, line 28: Replace 'catch' with 'catch (error)' to handle the error and provide a more specific message or log the error for debugging purposes.

src/App.tsx
```diff
@@ -1,24 +1,66 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import { useEffect } from 'react';
+import { useTodoContext } from './context/TodoContext';
+import { getTodos } from './api/todos';
+import { Error } from './types/Error';
```
liliia-livandovskaya, line 5: The 'Error' type import should be changed to a different name like 'AppError' to avoid confusion with the built-in 'Error' class in JavaScript.

src/App.tsx
```diff
@@ -1,24 +1,66 @@
...
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          <button type="button" className="todoapp__toggle-all active" />
```
liliia-livandovskaya, line 42: Add an 'onClick' event handler to the 'todoapp__toggle-all' button to handle the functionality of toggling all todos.

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+  error: null,
+  setError: () => {},
+});
+
+export const useTodoContext = () => useContext(TodoContext);
+
+export const TodoContextProvider: React.FC<PropsWithChildren> = ({
+  children,
+}) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
```
liliia-livandovskaya, line 36: The 'useState' for 'todos' should be initialized with the initial state coming from the API.

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,58 @@
+import { useState } from 'react';
+import { useTodoContext } from '../context/TodoContext';
+import { addTodo } from '../api/todos';
+import { Error } from '../types/Error';
```
liliia-livandovskaya, line 4: It's better to use custom error handling instead of using a generic Error type. Create a custom Error enum in the 'types' folder and import it here.

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+    setTodos,
+    setError,
+  } = useTodoContext();
+
+  const [todoTitle, setTodoTitle] = useState('');
+  const [isCreating, setIsCreating] = useState(false);
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
```
liliia-livandovskaya, line 17: Clear the error state when the input is valid. Add 'setError(null);' before 'setIsCreating(true);'

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { deleteTodo } from '../api/todos';
+import { Todo } from '../types/Todo';
+import { Error } from '../types/Error';
+
+interface Props {
+  key: number;
+  todo: Todo;
+}
+
+export const TodoItem: React.FC<Props> = ({ key, todo }) => {
```
liliia-livandovskaya, line 13: Remove the 'key' prop from the TodoItem component, as it is a reserved property in React and is not necessary in this case.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,66 @@
+import cn from 'classnames';
+import { useState } from 'react';
+import { useTodoContext } from '../context/TodoContext';
+import { deleteTodo } from '../api/todos';
+import { Todo } from '../types/Todo';
+import { Error } from '../types/Error';
+
+interface Props {
+  key: number;
```
liliia-livandovskaya, line 9: The 'key' prop should not be used in the Props interface since it is a reserved property in React. Please remove 'key' from Props.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const handleTodoDelete = async () => {
+    setIsTodoDeleting(true);
+
+    try {
+      await deleteTodo(todo.id);
+
+      setTodos((prevTodos) => prevTodos.filter((t) => t.id !== todo.id));
+      setIsTodoDeleting(false);
+    } catch (error) {
+      setError(Error.DELETE);
```
liliia-livandovskaya, line 26: It is better to use the error object from the catch block to set a more specific error message. Change 'setError(Error.DELETE)' to 'setError(error.message || Error.DELETE)'.

src/App.tsx
Zad0rnik: this should be done in the next task, in that task I already have this logic

src/components/FooterFilter.tsx
```diff
@@ -0,0 +1,58 @@
...
+          onClick={(event) => {
+            event.preventDefault();
+            setFilter(Filter.COMPLETED);
+          }}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {todos.some((todo) => todo.completed) && (
```
ManoylovAC, line 51: It will be clearer and better to store the result of this check in a variable and use the variable here.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [isTodoDeleting, setIsTodoDeleting] = useState(false);
+
+  const handleTodoDelete = async () => {
+    setIsTodoDeleting(true);
+
+    try {
+      await deleteTodo(todo.id);
+
+      setTodos((prevTodos) => prevTodos.filter((t) => t.id !== todo.id));
+      setIsTodoDeleting(false);
```
ManoylovAC, line 23: you can do it in one time in `finally` block instead here and same thing in catch block

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+      return;
+    }
+
+    setError(null);
+    setIsCreating(true);
+
+    const newTodo = {
+      id: 0,
+      userId: 10407,
```
ManoylovAC, line 29: If you use need some constant values you can hold it in constant variables

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+      setError(TodoError.INVALID_INPUT);
+
+      return;
+    }
+
+    setError(null);
+    setIsCreating(true);
+
+    const newTodo = {
+      id: 0,
```
ManoylovAC, line 28: hardcoded id not the best idea
I think you no need this param in this case at all
```suggestion
```

ManoylovAC (changes requested):
Nice, all works good!
Please clean up few issues was mentioned.

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+    setIsCreating(true);
+
+    const newTodo = {
+      userId: USER_ID,
+      title: todoTitle,
+      completed: false,
+    };
+
+    try {
+      const todo = await addTodo(newTodo);
```
ManoylovAC, line 35: In theory, the form component shouldn't do any work other than preparing the data. This can be done in some parent component to share responsibility.

For example, a doctor prescribes medicine, but does not sell it. It is up to the patient to decide where, when and how to purchase them. In the same way, the form collects, validates and prepares data like a recipe and passes it on. And then it depends on what is needed: either upload data to the server or do something else.

Please think of this as the Principles of Single Responsibility

src/types/Error.ts
```diff
@@ -0,0 +1,6 @@
+export enum TodoError {
+  LOAD = 'Unable to load todos',
```
ManoylovAC, line 2: it will be more accurate with clarifying messages
```suggestion
   UNABLE_LOAD = 'Unable to load todos',
   UNABLE_ADD = 'Unable....',
```

ManoylovAC (changes requested):
Very close
I left a few small and one useful comment to think about.
I think it will be useful to redo the form a little.
Good luck!

src/App.tsx
SerhiiKirik: Agree

SerhiiKirik (approved):
Great 🔥

PR #459, opened 2023-05-15
react_todo-app-add-and-delete

ymys1:
[DEMO LINK](https://ymys1.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,169 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { client } from './utils/fetchClient';
+import { Todo, TodoStatus } from './types/Todo';
+import { ErrorNotification } from './ErrorNotification';
+import { TodosList } from './TodosList';
+import { TodosHeader } from './TodosHeader';
+import { TodosFooter } from './TodosFooter';
 
-const USER_ID = 0;
+const USER_ID = 10332;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [statusFilter, setStatusFilter] = useState<TodoStatus>(TodoStatus.All);
+  const [todoAddQuery, setTodoAddQuery] = useState('');
+  const [disableInput, setDisableInput] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
+
+  const loadTodos = async () => {
+    try {
+      const response = await client.get<Todo[]>('/todos?userId=10332');
+
+      setTodos(response);
+    } catch {
+      setError('Unable to load todos');
+    }
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,24 +1,169 @@
...
+import { TodosFooter } from './TodosFooter';
 
-const USER_ID = 0;
+const USER_ID = 10332;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [statusFilter, setStatusFilter] = useState<TodoStatus>(TodoStatus.All);
+  const [todoAddQuery, setTodoAddQuery] = useState('');
```
Goto-Jesus, line 17: You can call this like this 'title'

src/App.tsx
```diff
@@ -1,24 +1,169 @@
...
+    setTempTodo({
+      id: 0,
+      title: todoAddQuery,
+      completed: false,
+      userId: USER_ID,
+    });
+
+    try {
+      const response = await client.post<Todo>('/todos', {
+        title: todoAddQuery,
```
Goto-Jesus, line 57: After you write just (title,) instead of (title: todoAddQuery,)

Goto-Jesus (approved):
nice but you can create

src/ErrorNotification.tsx
```diff
@@ -0,0 +1,46 @@
...
+  error: string;
+};
+
+export const ErrorNotification:React.FC<Props> = ({ error }) => {
+  const [hidden, setHidden] = useState(false);
+
+  const autoHideError = useCallback(
+    () => {
+      setTimeout(() => {
+        setHidden(true);
```
liliia-livandovskaya, line 14: There is no clearTimeout for the timeout set in autoHideError. This can lead to memory leaks when the component is unmounted before the timeout is completed. Use useEffect to clean up the timeout when the component is unmounted or the error prop changes.

src/ErrorNotification.tsx
```diff
@@ -0,0 +1,46 @@
...
+      }, 3000);
+    }, [],
+  );
+
+  autoHideError();
```
liliia-livandovskaya, line 19: The autoHideError function is called directly in the component body, which may cause multiple timeouts to be set during re-renders. Consider using useEffect to call this function only on initial render or when the error prop changes.

src/ErrorNotification.tsx
```diff
@@ -0,0 +1,46 @@
...
+        aria-label="Close error notification"
+      />
+      {error}
+
+      {/* show only one message at a time */}
+      {/* Unable to add a todo
+      < br />
+      Unable to delete a todo
+      < br />
+      Unable to update a todo */}
```
liliia-livandovskaya, line 43: remove comments

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comments

src/TodosList.tsx
```diff
@@ -0,0 +1,37 @@
...
+  onDelete: (id: number) => void;
+};
+
+export const TodosList:React.FC<Props> = ({
+  todos,
+  tempTodo,
+  loadingTodoId,
+  onDelete,
+}) => {
+  return (
```
ManoylovAC, line 19: you can use implicit return

ManoylovAC (changes requested):
Good job
I have left for you some small tips for improvement that you will quickly deal with.
Please pay special attention to naming - it is an important part of programming.

PR #482, opened 2023-05-17
Implemented Adding/Deleting Todos

vksazonov:
[DEMO LINK](https://vksazonov.github.io/react_todo-app-add-and-delete/)
I know that I have some problems with adding a Todo to my list, but I dunno what's the problem.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,118 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import { UserWarning } from './UserWarning';
+import { createTodo, getTodos, removeTodo } from './api/todos';
+import { Todo } from './types/Todo';
+import { Errors, Sort } from './utils/enums';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
+// eslint-disable-next-line import/no-cycle
+import { Header } from './components/Header';
 
-const USER_ID = 0;
+export const USER_ID = 10284;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [sort, setSort] = useState(Sort.All);
+
+  const [isLoading, setIsLoading] = useState(false);
+
+  const [Error, setError] = useState<Errors | null>(null);
+
+  const loadTodos = useCallback(async () => {
+    try {
+      const todoList = await getTodos(USER_ID);
+
+      setTodos(todoList);
+    } catch {
+      setError(Errors.Url);
+    }
+  }, []);
+
+  if (Error) {
+    setTimeout(() => {
+      setError(null);
+    }, 3000);
+  }
+
+  const addTodo = useCallback(async (todo: Todo) => {
+    try {
+      setIsLoading(true);
+      await createTodo(todo);
+    } catch {
+      setError(Errors.Add);
+    } finally {
+      loadTodos();
+      setIsLoading(false);
+    }
+  }, []);
+
+  const deleteTodo = useCallback(async (todoId:number) => {
+    setError(null);
+    try {
+      await removeTodo(todoId);
+    } catch {
+      setError(Errors.Delete);
+    }
+
+    loadTodos();
+  }, []);
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  const visibleTodos = todos.filter(todo => {
+    switch (sort) {
+      case Sort.Active: return !todo.completed;
+      case Sort.Completed: return todo.completed;
+      default: return todos;
+    }
+  });
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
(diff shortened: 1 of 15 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,93 @@
...
+  todos,
+  onAdd,
+  setError,
+  setTempTodo,
+  setIsLoading,
+  loadTodos,
+}) => {
+  const [query, setQuery] = useState('');
+  const [isDisabled, setIsDisabled] = useState(false);
+
```
liliia-livandovskaya, line 31: I think the issue is in the Header component. Here is an example of suggested fix 
```
// Header Component
export const Header:FC<Props> = ({
  todos,
  onAdd,
  setError,
  setTempTodo,
  setIsLoading,
  loadTodos,
}) => {
  const [query, setQuery] = useState('');
  const [isDisabled, setIsDisabled] = useState(false);

  const handleSubmit = useCallback(async (event: FormEvent) => {
    event.preventDefault();

    if (!query) {
      setError(Errors.Title);

      return;
    }

    const todoData: TodoData = {
      completed: false,
      title: query,
      userId: USER_ID,
    };

    setTempTodo({ ...todoData, id: 0 });
    setIsLoading(true);
    try {
      setIsDisabled(true);
      const newTodo = await createTodo(todoData);

      onAdd(newTodo);
      setQuery('');
    } catch {
      setError(Errors.Add);
    } finally {
      // Load todos only after successfully adding a new one.
      // This prevents your tempTodo from disappearing.
      if (!setError) {
        loadTodos();
      }
      setTempTodo(null);
      setIsLoading(false);
      setIsDisabled(false);
    }
  }, [query]);

  // Remaining code...
};
```

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+      {hasError === Errors.Title && (
+        <p>{Errors.Title}</p>
+      )}
+
+      {hasError === Errors.Add && (
+        <p>{Errors.Add}</p>
+      )}
+
+      {hasError === Errors.Url && (
```
liliia-livandovskaya, line 35: The condition for displaying the Delete error should be 'hasError === Errors.Delete' instead of 'hasError === Errors.Url'.

src/App.tsx
```diff
@@ -1,24 +1,108 @@
...
+    try {
+      const todoList = await getTodos(USER_ID);
+
+      setTodos(todoList);
+    } catch {
+      setError(Errors.Url);
+    }
+  }, []);
+
+  if (Error) {
```
liliia-livandovskaya, line 36: Use 'if (error)' instead of 'if (Error)' to check for errors.

src/App.tsx
```diff
@@ -1,24 +1,108 @@
...
+    try {
+      await removeTodo(todoId);
+    } catch {
+      setError(Errors.Delete);
+    }
+
+    loadTodos();
+  }, []);
+
+  useEffect(() => {
```
liliia-livandovskaya, line 53: Add an empty array dependency '[]' to useEffect to ensure it runs only once.

src/App.tsx
```diff
@@ -1,24 +1,108 @@
...
+    loadTodos();
+  }, []);
+
+  const visibleTodos = todos.filter(todo => {
+    switch (sort) {
+      case Sort.Active: return !todo.completed;
+      case Sort.Completed: return todo.completed;
+      default: return todos;
```
liliia-livandovskaya, line 61: Return 'todo' instead of 'todos' in the default case.

src/App.tsx
```diff
@@ -1,24 +1,108 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [sort, setSort] = useState(Sort.All);
+
+  const [isLoading, setIsLoading] = useState(false);
+
+  const [Error, setError] = useState<Errors | null>(null);
```
liliia-livandovskaya, line 24: Use camelCase for variable naming. Rename 'Error' to 'error'.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,93 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
liliia-livandovskaya, line 1: Consider removing the 'eslint-disable jsx-a11y/control-has-associated-label' comment and adding an 'aria-label' attribute to the button element in line 69 to improve accessibility.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,93 @@
...
+  setError,
+  setTempTodo,
+  setIsLoading,
+  loadTodos,
+}) => {
+  const [query, setQuery] = useState('');
+  const [isDisabled, setIsDisabled] = useState(false);
+
+  const handleSubmit = useCallback(async (event: FormEvent) => {
+    event.preventDefault();
```
liliia-livandovskaya, line 33: Add the missing dependencies 'setError', 'setTempTodo', 'setIsLoading', 'loadTodos', and 'onAdd' to the dependency array of useCallback.

liliia-livandovskaya (changes requested):
Pay attention to the comments

src/App.tsx
vksazonov: But this Error is going from the state

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <p>{Errors.Add}</p>
+    )}
+
+    {hasError === Errors.Delete && (
+      <p>{Errors.Delete}</p>
+    )}
+
+    {hasError === Errors.Update && (
+      <p>{Errors.Update}</p>
+    )}
```
jstmpelowycz, line 40: seems like something as switch-case statement matches your goals here

jstmpelowycz (changes requested):
things to comment on:
- it should not be possible to create a todo with an empty title
- don't forget to optimize components rendering where needed

PR #489, opened 2023-05-20
Develop

Roman-Malysh:
https://Roman-Malysh.github.io/react_todo-app-add-and-delete/


App.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const todosFromServer = await getTodos(USER_ID);
+
+        setTodos(todosFromServer);
+      } catch (error) {
+        console.log('122112');
```
TarasHoliuk, line 43: ??

TarasHoliuk (approved):
I've approved the previous task (the app is complete - all features work), so I'll approve this one as well (code is 99% the same, correct?)

App.tsx
Roman-Malysh: yes, I am just doing all at once. Will change log into error .

Roman-Malysh: It actually was changed. You can see it in the App.tsx. Idc why this log shows up :)

PR #494, opened 2023-05-25
solution

LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,95 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { Header } from './components/header/Header';
+import { List } from './components/list/List';
+import { Footer } from './components/footer/Footer';
 import { UserWarning } from './UserWarning';
+import { Warnings } from './components/warnings/Warnings';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { Select } from './types/Select';
+import { Error } from './types/Error';
 
-const USER_ID = 0;
+const USER_ID = 10514;
+
+const { All, Active, Completed } = Select;
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<[] | Todo[]>([]);
+  const [filterBy, setFilterBy] = useState(All);
+  const [error, setError] = useState('');
+  const [processings, setProcessings] = useState<number[] | []>([]);
+
+  const loadTodos = () => {
+    getTodos(USER_ID)
+      .then((todos) => setTodoList(todos))
+      .catch(() => setError(Error.Get));
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  useEffect(() => {
+    setTimeout(() => setError(''), 3000);
+  }, [error]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
+  function filterList() {
+    switch (filterBy) {
+      case Active:
+        return todoList.filter((todo: Todo) => !todo.completed);
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      case Completed:
+        return todoList
+          .filter((todo: Todo) => todo.completed || todo.completed === null);
+
+      default:
+        return todoList;
+    }
+  }
+
+  function getactiveId(id : number | null) {
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    !id ? setProcessings([]) : setProcessings(prev => [...prev, id]);
+  }
+
+  return (
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <Header
+          setTodoList={setTodoList}
+          todoList={todoList}
+          setError={setError}
+          setProcessings={(id) => getactiveId(id)}
+        />
+
+        <List
+          setTodoList={setTodoList}
+          todoList={filterList()}
+          setError={setError}
+          setProcessings={(id) => getactiveId(id)}
+          processings={processings}
+
+        />
+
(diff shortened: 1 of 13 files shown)


src/types/Todo.ts
```diff
@@ -0,0 +1,6 @@
+export interface Todo {
+  id: number;
+  userId: number;
+  title: string;
+  completed: boolean | null;
```
vlad-demchuk, line 5: Consider changing the type of 'completed' property to 'boolean' only, instead of 'boolean | null', as a todo item should have a definite completed status (true or false).

src/types/Error.ts
```diff
@@ -0,0 +1,7 @@
+export enum Error {
+  Patch = 'Unable to update a todo',
+  Get = 'Unable to get a todos',
```
vlad-demchuk, line 3: Change the message to 'Unable to get todos' for better grammar.

src/types/Error.ts
```diff
@@ -0,0 +1,7 @@
+export enum Error {
```
vlad-demchuk, line 1: Consider renaming the enum to 'ErrorMessage' for better readability and understanding.

src/components/header/Header.tsx
```diff
@@ -0,0 +1,95 @@
...
+      if (!inputValue) {
+        setError(Error.Title);
+
+        return;
+      }
+
+      const newTodo = {
+        title: inputValue,
+        completed: null,
+        userId: 10514,
```
vlad-demchuk, line 39: The 'completed' property of a new todo should be set to false, not null.

src/api/todos.ts
```diff
@@ -0,0 +1,22 @@
...
+    title,
+    completed: false,
+  });
+};
+
+export const deleteTodos = (id: number) => {
+  return client.delete(`/todos/${id}`);
+};
+
+export const patchTodos = (id: number, data: Partial<Todo>) => {
```
vlad-demchuk, line 20: The function name 'patchTodos' should be more descriptive. Consider renaming it to 'updateTodo'.

src/App.tsx
```diff
@@ -1,24 +1,95 @@
...
   }
 
+  function filterList() {
+    switch (filterBy) {
+      case Active:
+        return todoList.filter((todo: Todo) => !todo.completed);
+
+      case Completed:
+        return todoList
+          .filter((todo: Todo) => todo.completed || todo.completed === null);
```
vlad-demchuk, line 48: Consider checking for 'todo.completed === false' instead of 'todo.completed === null', as the 'completed' property should be a boolean.

src/types/Select.tsx
```diff
@@ -0,0 +1,5 @@
+export enum Select {
```
vlad-demchuk, line 1: The enum name 'Select' is not very descriptive. Consider renaming it to 'TodoFilter' or something more descriptive of its purpose.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
vlad-demchuk, line 34: Provide a more descriptive error message when throwing an error, such as 'throw new Error('Request failed with status ' + response.status);'

src/App.tsx
```diff
@@ -1,24 +1,95 @@
...
+import { Select } from './types/Select';
+import { Error } from './types/Error';
 
-const USER_ID = 0;
+const USER_ID = 10514;
+
+const { All, Active, Completed } = Select;
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<[] | Todo[]>([]);
```
vlad-demchuk, line 18: it is enough just use `<Todo[]>`
```suggestion
  const [todoList, setTodoList] = useState<Todo[]>([]);
```

vlad-demchuk: fix all such cases with arrays

src/api/todos.ts
```diff
@@ -0,0 +1,22 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+export const postTodos = (userId: number, title: string) => {
```
vlad-demchuk, line 8:
```suggestion
export const postTodo = (userId: number, title: string) => {
```

src/api/todos.ts
```diff
@@ -0,0 +1,22 @@
...
+  return client.post<Todo>(`/todos?userId=${userId}`, {
+    userId,
+    title,
+    completed: false,
+  });
+};
+
+export const deleteTodos = (id: number) => {
```
vlad-demchuk, line 16:
```suggestion
export const deleteTodo = (id: number) => {
```

src/api/todos.ts
vlad-demchuk, line 20:
```suggestion
export const patchTodo = (id: number, data: Partial<Todo>) => {
```

src/types/Select.tsx
vlad-demchuk: if file doesn't include jsx, no need to make extension tsx, use just ts

vlad-demchuk (changes requested):

LBondarchuk:
Hello. thanks for the feedback

src/components/list/List.tsx
```diff
@@ -0,0 +1,75 @@
+import cn from 'classnames';
+// import { useState } from 'react';
```
OleziO, line 2: Remove all comments from the code.

OleziO (changes requested):

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
```
OleziO, line 4: Remove all comments from the code.

OleziO (changes requested):

LBondarchuk:
sorry, didn't look that deep

Volodymyr-Melnyk-Fron10der (approved):

PR #498, opened 2023-05-25
solution

ladrom:
[demo](https://ladrom.github.io/react_todo-app-add-and-delete/)


src/utils/FetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+
+function request<T>(
+  url: string,
+  method: RequestMethod = 'GET',
+  data: any = null, // we can send any data to the server
+): Promise<T> {
+  const options: RequestInit = { method };
+
+  if (data) {
+    // We add body and Content-Type only for the requests with data
```
witflash, line 22: You don't need to leave such comments. Ideally, there should be no comments at all.

witflash (approved):
Great job! Well done 👍

PR #501, opened 2023-05-26
Solution

Zibi95:
[DEMO LINK](https://Zibi95.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/hooks/useTodos.ts
+++ b/src/hooks/useTodos.ts
@@ -1,0 +1,28 @@
+import { useEffect, useState } from 'react';
+import { TodoType } from '../types/Todo';
+import { getTodos } from '../api/todos';
+
+export const useTodos = (userId?: number) => {
+  const [todos, setTodos] = useState<TodoType[]>([]);
+  const [error, setError] = useState('');
+
+  const showError = (errorMsg: string) => {
+    setError(errorMsg);
+
+    setTimeout(() => setError(''), 3000);
+  };
+
+  useEffect(() => {
+    if (userId) {
+      getTodos(userId)
+        .then(data => setTodos(data))
+        .catch(() => showError('Unable to load todos'));
+    }
+  }, [userId]);
+
+  return {
+    todos,
+    error,
+    showError,
+  };
+};
--- a/src/hooks/useTodos.ts
+++ b/src/hooks/useTodos.ts
@@ -1,9 +1,10 @@
 import { useEffect, useState } from 'react';
 import { TodoType } from '../types/Todo';
-import { getTodos } from '../api/todos';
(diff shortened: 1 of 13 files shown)


src/hooks/useTodos.ts
```diff
@@ -0,0 +1,69 @@
...
+  const handleDeleteTodo = async (todoId: number) => {
+    deleteTodo(todoId)
+      .then(() => setTodos(
+        prevState => prevState.filter((todo) => todo.id !== todoId),
+      ))
+      .catch(() => showError('Unable to delete a todo'));
+  };
+
+  const handleClearCompleted = (completedTodos: TodoType[]) => {
+    completedTodos.forEach(({ id }) => handleDeleteTodo(id));
```
Radoslaw-Czerniawski, line 49: This generally may be problematic if more than one requests fails, as for each on of them you will set and show the error. You could await all these promises and if one of them fails just once show the error

image

I didn't memoize these functions but like mentioned in the previous comment it could be done so for all of them, remember to add appropriate dependencies to their dependency arrays

src/hooks/useTodos.ts
```diff
@@ -0,0 +1,69 @@
...
+        .then((data) => setTodos(data))
+        .catch(() => showError('Unable to load todos'));
+    }
+  }, [userId]);
+
+  return {
+    todos,
+    tempTodo,
+    error,
+    showError,
```
Radoslaw-Czerniawski, line 64: showError is not used outside of this hook so it doesn't need to be returned here

Radoslaw-Czerniawski (changes requested):
Some changes needed, might also take one more look into the code for more in-depth re-review later

src/hooks/useTodos.ts
```diff
@@ -0,0 +1,69 @@
+import { useEffect, useState } from 'react';
+import { TodoType } from '../types/Todo';
+import { deleteTodo, getTodos, postTodo } from '../api/todos';
+
+export const useTodos = (userId: number) => {
+  const [todos, setTodos] = useState<TodoType[]>([]);
+  const [tempTodo, setTempTodo] = useState<TodoType | null>(null);
+  const [error, setError] = useState('');
+
+  const showError = (errorMsg: string) => {
```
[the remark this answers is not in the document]
Zibi95: I still don't get it. I read a lot about memoizing hooks and how I shouldn't use them too much. I tought that I should start thinking about those hooks when I see that my component is rendering slow.

DorotaLeniecDev (changes requested):
On Demo page loader is not working when removing todo

DorotaLeniecDev (approved):
Good job 💪

PR #513, opened 2023-06-02
Develop

MikiOOOO:
https://MikiOOOO.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,142 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useCallback, useEffect, useState } from 'react';
+// import classNames from 'classnames';
+import { TodoData } from './types/TodoData';
+import { addTodo, getTodos, deleteTodo } from './api/todos';
+import { Notification } from './components/Notification';
+import { Status } from './types/Status';
+import { TodoList } from './components/TodoList';
+import { Filter } from './components/Filter';
+import { ActionError } from './types/ActionError';
+import { NewTodo } from './components/NewTodo';
+import { Todo } from './components/Todo';
 
-const USER_ID = 0;
+const USER_ID = 10524;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<TodoData[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<TodoData[]>([]);
+  const [errorMessage, setErrorMessage] = useState<ActionError | string>('');
+  const [tempTodo, setTempTodo] = useState<TodoData | null>(null);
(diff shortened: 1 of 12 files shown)


src/types/Status.ts
```diff
@@ -0,0 +1 @@
+export type Status = 'all' | 'active' | 'completed';
```
d-wasilewski, line 1: You can make it an enum instead of a type. This way you will be able to use it in code like Status.ALL instead of strings

src/components/TodoList.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export const TodoList = ({ onTodoDelete, todos }: TodoListProps) => {
+  return (
+    <>
+      {todos.map((todo) => (
+        <Todo
+          key={todo.id}
+          todo={todo}
+          isTempTodo={false}
```
d-wasilewski, line 16: I think you don't even have to pass it explicitly here if you want it.to be false (it will be undefined but it will work the same)

src/components/NewTodo.tsx
```diff
@@ -0,0 +1,27 @@
+import { useState } from 'react';
+
+interface NewTodoProps {
+  onTodoAdd: (newTodoTitle: string) => void;
+}
+
+export const NewTodo = ({ onTodoAdd }: NewTodoProps) => {
+  const [newTodoTitle, setNewTodoTitle] = useState<string>('');
```
d-wasilewski, line 8:
```suggestion
  const [newTodoTitle, setNewTodoTitle] = useState('');
```

src/App.tsx
```diff
@@ -1,24 +1,142 @@
...
+        setTodos(newTodos);
+        setFilteredTodos(newTodos);
+      })
+      .catch(() => setErrorMessage(ActionError.delete));
+  }, [todos]);
+
+  const clearCompletedTodos = () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    completedTodos.forEach(completedTodo => handleTodoDelete(completedTodo.id));
```
d-wasilewski, line 96: you can use .map instead of .forEach and wrap it into Promise.all, because each handleTodoDelete returns a promise

src/App.tsx
```diff
@@ -1,24 +1,142 @@
...
-const USER_ID = 0;
+const USER_ID = 10524;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<TodoData[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<TodoData[]>([]);
+  const [errorMessage, setErrorMessage] = useState<ActionError | string>('');
```
d-wasilewski, line 19:
```suggestion
  const [errorMessage, setErrorMessage] = useState<ActionError>('');
```

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #515, opened 2023-06-02
part 2

bjeziorek:
[DEMO LINK](https://bjeziorek.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,131 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { Todo } from './components/Todo';
+import { Error } from './components/Error';
+import { Todo as TodoType } from './types/Todo';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
+import { Filter } from './types/Filter';
 
-const USER_ID = 0;
+export const USER_ID = 10583;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<TodoType[]>([]);
+  const [tempTodos, setTempTodos] = useState<TodoType[]>([]);
+  const [error, setError] = useState<string>('');
+  const [isInputLocked, setIsInputLocked] = useState(false);
+  const [filter, setFilter] = useState<Filter>('all');
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((data) => setTodos(data))
+      .catch(() => setError('Unable to fetch data'));
+  }, []);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setError('');
+    }, 3000);
+  }, [error]);
+
+  useEffect(() => {
+    tempTodos.forEach((todo) => addTodo({
+      title: todo.title,
+      completed: todo.completed,
+      userId: USER_ID,
+    })
+      .then((res) => {
+        if (res) {
+          const newTodo = { ...todo };
+
+          setTempTodos((prev) => prev.filter((task) => task.id !== todo.id));
+          newTodo.id = res.id;
+          setTodos((prev) => [...prev, newTodo]);
+        }
+      })
+      .catch(() => setError('Unable to add a todo'))
+      .finally(() => setIsInputLocked(false)));
+  }, [tempTodos]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
+  const deleteTask = (id: number): void => {
+    setTodos((prev) => prev.filter((todo) => todo.id !== id));
+  };
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+  const removeAllCompleted = () => {
+    todos
+      .filter((todo) => todo.completed)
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
+import { Todo as TodoType } from './types/Todo';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
+import { Filter } from './types/Filter';
 
-const USER_ID = 0;
+export const USER_ID = 10583;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<TodoType[]>([]);
+  const [tempTodos, setTempTodos] = useState<TodoType[]>([]);
```
d-wasilewski, line 16: There should be always only 1 temp todo instead of an array of them

src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        {todos?.length && (
```
d-wasilewski, line 74: This array will always exist
```suggestion
        {todos.length && (
```

d-wasilewski: Try to refresh the page and see that there is 0 hanging as a footer. You have to either cast this todos.length to boolean or write `todos.length > 0`. This way you will have a boolean expression instead of 0 that won't get shown. I know we sometimes ask you to shorten the expressions, and it's better in general, but this is one case when you have to be more cautious with it
```suggestion
        {todos.length > 0 && (
```

src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
+            isInputLocked={isInputLocked}
+            setIsInputLocked={setIsInputLocked}
+          />
+        )}
+
+        <section className="todoapp__main">
+          {todos
+            ?.filter((todo: TodoType) => {
+              switch (filter) {
+                case 'all':
```
d-wasilewski, line 87: Create an enum for filter types so that you don't have to use unknown strings
```suggestion
                case FilterType.ALL:
```

d-wasilewski: and you don't need the case for 'all' here, it's your default case

src/api/todos.ts
```diff
@@ -0,0 +1,15 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+export const addTodo = (data: Omit<Todo, 'createdAt'
```
d-wasilewski, line 8: Does`Todo` even have these properties that you have to omit them?

src/components/Error.tsx
```diff
@@ -0,0 +1,35 @@
...
+  error:string,
+}
+
+export const Error: React.FC<ErrorProps> = ({ error }) => {
+  const [message, setMessage] = useState(error);
+  const [isHidden, setIsHidden] = useState(false);
+
+  useEffect(() => {
+    setMessage(error);
+    if (error.length) {
```
d-wasilewski, line 15:
```suggestion
    if (error) {
```

src/components/Error.tsx
```diff
@@ -0,0 +1,35 @@
...
+      setIsHidden(false);
```
d-wasilewski, line 16: you should just set this onClick, and you don't need a state for `message` and useEffect here

src/components/Todo.tsx
```diff
@@ -0,0 +1,96 @@
...
+        if (res) {
+          deleteTask(todo.id);
+        }
+      })
+      .catch(() => setError('Unable to delete a todo'));
+  };
+
+  const handleEnter = (event: KeyboardEvent<HTMLInputElement>) => {
+    if (event.keyCode === 13) {
+      event.preventDefault();
```
d-wasilewski, line 38: Enter should confirm the action instead of interrupting it

d-wasilewski (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
+import { Todo as TodoType } from './types/Todo';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
+import { Filter } from './types/Filter';
 
-const USER_ID = 0;
+export const USER_ID = 10583;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<TodoType[]>([]);
+  const [tempTodos, setTempTodos] = useState<TodoType | null>(null);
```
d-wasilewski, line 16:
```suggestion
  const [tempTodo, setTempTodo] = useState<TodoType | null>(null);
```

src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
+  const [error, setError] = useState<string>('');
```
d-wasilewski, line 17:
```suggestion
  const [error, setError] = useState('');
```

src/App.tsx
```diff
@@ -1,24 +1,131 @@
...
+            setTempTodos={setTempTodos}
+            setError={setError}
+            isInputLocked={isInputLocked}
+            setIsInputLocked={setIsInputLocked}
+          />
+        )}
+
+        <section className="todoapp__main">
+          {todos
+            ?.filter((todo: TodoType) => {
```
d-wasilewski, line 87:
```suggestion
           .filter((todo: TodoType) => {
```

src/components/Error.tsx
d-wasilewski: > you should just set this onClick, and you don't need a state for message and don'e use useEffect here

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #516, opened 2023-06-02
Solution

KielpinskiJ:
[DEMO LINK](https://KielpinskiJ.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,218 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  createTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoHeader } from './components/TodoHeader/TodoHeader';
+import { TodoFooter } from './components/TodoFooter/TodoFooter';
+import {
+  TodoNotification,
+} from './components/TodoNotification/TodoNotification';
 
-const USER_ID = 0;
+const USER_ID = 10594;
 
 export const App: React.FC = () => {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,24 +1,218 @@
...
+  };
+
+  const handleDismissNotification = () => {
+    setErrorMessage('');
+  };
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filter) {
+      case 'all':
+        return true;
```
d-wasilewski, line 150: it's your default case, you can omit it here

src/App.tsx
```diff
@@ -1,24 +1,218 @@
...
+      case 'active':
```
d-wasilewski, line 151: let's create an enum for that 
```suggestion
      case FilterType.ACTIVE:
```

src/api/todos.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const createTodo = async (title: string, userId: number) => {
+  const response = await client.post<Todo>('/todos', {
+    title,
+    userId,
+    completed: false,
+  });
+
+  return response;
```
d-wasilewski, line 15: These functions should return a promise
```suggestion
  return client.post<Todo>('/todos', {
    title,
    userId,
    completed: false,
  });

```

d-wasilewski (changes requested):
error messages should disappear after 3 seconds

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,76 @@
...
+      {' items left'}
+    </span>
+
+    <nav className="filter">
+      <a
+        href="#/"
+        className={
+          classNames(
+            'filter__link',
+            { selected: filter === 'all' },
```
d-wasilewski, line 32: You should use it in all places where it occurs
```suggestion
            { selected: filter === FilterType.ALL },
```

d-wasilewski (changes requested):
error message still doesn't disappear after 3 seconds

kasiakubisiak (approved):

PR #519, opened 2023-06-05
Develop

marshalek37:
[DEMO LINK](https://<your_account>.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,94 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { client } from './utils/fetchClient';
+import { Todo } from './types/Todo';
+import Footer from './components/Footer';
+import TodoList from './components/TodoList';
 
-const USER_ID = 0;
+export const userId = 10590;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,24 +1,171 @@
...
+import Footer from './components/Footer';
+import TodoList from './components/TodoList';
 
-const USER_ID = 0;
+export const userId = 10590;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
```
d-wasilewski, line 12: If there is no error just place an empty string here
```suggestion
  const [error, setError] = useState('');
```

src/App.tsx
```diff
@@ -1,24 +1,171 @@
...
+  const [filterType, setFilterType] = useState('all');
```
d-wasilewski, line 13: Do not rely on the unknown string, make enum for this.
```suggestion
  const [filterType, setFilterType] = useState(FilterType.ALL);
```

src/App.tsx
```diff
@@ -1,24 +1,171 @@
...
+      try {
+        const response = await client.post<Todo>('/todos', {
+          title: newTodo,
+          completed: false,
+          userId,
+        });
+
+        setTodos((prevTodos) => [...prevTodos, response]);
+        setTempTodo(null);
+        setIsLoading(false);
```
d-wasilewski, line 77: this should be only in `.finally` clause instead of both try and catch

d-wasilewski (changes requested):
Add loader when the todo is being deleted, and make the temp todo only covered by the loader like on the original solution

PR #530, opened 2023-06-07
Solution

kacper-lyczba:
[DEMO LINK](https://suaby514.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Remove all the completed todos after the `Clear completed` button click:
 
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://suaby514.github.io/react_todo-app-add-and-delete/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://kacper-lyczba.github.io/react_todo-app-add-and-delete/) and add it to the PR description.
 
 ## IF you want to implement smooth animations
 


SuaBy514:
Test failed due to missing animation library from the optional instruction

src/App.tsx
```diff
@@ -1,24 +1,118 @@
...
+          />
+        </header>
+
+        <TodoList
+          todos={currentTodos}
+          tempTodo={tempTodo}
+          onDeleteTodo={handleDeleteTodo}
+        />
+
+        {todos.length && (
```
d-wasilewski, line 105: Try to refresh the page and see that there is `0` hanging as a footer. You have to either cast this `todos.length` to boolean or write ``todos.length > 0`. This way you will have a boolean expression instead of 0 that won't get shown. I know we sometimes ask you to shorten the expressions, and it's better in general, but this is one case when you have to be more cautious with it
```suggestion
        {!!todos.length && (
```

src/App.tsx
```diff
@@ -1,24 +1,118 @@
...
+    if (errorMessage) {
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, [errorMessage]);
+
+  useEffect(() => {
+    switch (filterType) {
+      case 'completed': {
```
d-wasilewski, line 51:
```suggestion
      case FIlterTypes.COMPLETE: {
```

src/App.tsx
```diff
@@ -1,24 +1,118 @@
...
+      }
+    }
+  }, [todos, filterType]);
+
+  useEffect(() => {
+    if (tempTodo) {
+      addTodo(tempTodo)
+        .then((thisTodo) => {
+          setTodos([...todos, thisTodo]);
+          setTempTodo(null);
```
d-wasilewski, line 72: you can set it to null in a `.finally` clause instead of both catch and then

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,57 @@
...
+      onError('Title can\'t be empty');
+    } else {
+      const thisTodo: Todo = {
+        id: 0,
+        userId,
+        title,
+        completed: false,
+      };
+
+      onTodoSubmit(thisTodo);
```
d-wasilewski, line 41: I think that in this place (on form submit) you should just put all the logic related to this (request, loading, setting tempTodo) instead of doing it in useEffect in App

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,79 @@
...
+                  readOnly
+                />
+              </label>
+
+              <span className="todo__title">{title}</span>
+
+              <button
+                type="button"
+                className="todo__remove"
+                onClick={() => onDeleteTodo(String(todo.id))}
```
d-wasilewski, line 42: I think you can just make this function accept a number instead of casting it to string

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,79 @@
...
+      {todos.map((todo) => {
+        const { id, title, completed } = todo;
+
+        return (
+          <CSSTransition
+            key={id}
+            timeout={300}
+            classNames="item"
+          >
+            <div key={id} className={classNames('todo', { completed })}>
```
d-wasilewski, line 27: You already have the key on parent element (CSSTransition)
```suggestion
            <div className={classNames('todo', { completed })}>
```

d-wasilewski (changes requested):
Deleting operation
> covered the todo with the loader while wating for the API response;

kasiakubisiak (approved):
good job, just fix failing tests

PR #535, opened 2023-06-09
Solution

8Melya8:
[DEMO LINK](https://8melya8.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -23,3 +23,4 @@ body {
 @import "./todoapp";
 @import "./todo";
 @import "./filter";
+@import "./transition";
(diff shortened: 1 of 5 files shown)


demetrxx (changes requested):
Take into account following checklist requirements:

> if the title is empty show the "Title can't be empty" notification at the bottom;

> immediately after sending a request create a todo with id: 0 and save it to the tempTodo variable in the state (NOT to the todos array);

> show an independent TodoItem after the list if tempTodo is not null;

> temp TodoItem should have the loader (check the original markup);

> when deleting: covered the todo with the loader while wating for the API response;

src/styles/index.scss
```diff
@@ -23,3 +23,4 @@ body {
 @import "./todoapp";
```
etojeDenys, line 23: It seems that you have added a new import for './transition' in the index.scss file. Make sure the corresponding transition.scss file exists and contains the necessary styles for the new todo features (adding and deleting todos).

etojeDenys (approved):
lgtm

PR #536, opened 2023-06-09
React todo app add and delete task was done

SamirMamedov1010:
[DEMO LINK](https://SamirMamedov1010.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,177 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useState,
+  useMemo,
+} from 'react';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
+import { FilterValues, USER_ID } from './constants';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [errorType, setErrorType] = useState('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [isTodoAdding, setIsTodoAdding] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [selectedFilter, setSelectedFilter] = useState(FilterValues.ALL);
+  const [removingTodoIds, setRemovingTodoIds] = useState<number[]>([]);
+
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (selectedFilter) {
+        case FilterValues.COMPLETED:
+          return todo.completed;
+
+        case FilterValues.ACTIVE:
+          return !todo.completed;
+
+        case FilterValues.ALL:
+          return true;
+
+        default: return todo;
+      }
+    });
+  }, [selectedFilter, todos]);
+
+  const hasActive = todos.some(todoItem => !todoItem.completed);
+
+  const getTodosFromServer = async () => {
+    try {
+      const response = await getTodos(USER_ID);
+
+      setTodos(response);
+    } catch (error) {
+      setHasError(false);
+      setErrorType('upload');
+    }
+  };
+
+  const addTodoToServer = useCallback(async () => {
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,24 +1,177 @@
...
+
+  const hasActive = todos.some(todoItem => !todoItem.completed);
+
+  const getTodosFromServer = async () => {
+    try {
+      const response = await getTodos(USER_ID);
+
+      setTodos(response);
+    } catch (error) {
+      setHasError(false);
```
Volodymyr-Melnyk-Fron10der, line 52: hm, looks like smth wrong here)
```suggestion
      setHasError(true);
```

src/App.tsx
```diff
@@ -1,24 +1,177 @@
...
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [errorType, setErrorType] = useState('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
```
Volodymyr-Melnyk-Fron10der, line 21: isn't errorType enough for you? You can just check if `!!errorType ? errorType : null` wherever you want to show the error

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  tempTodo: Todo;
+};
+
+export const TodoItem: React.FC<Props> = ({ tempTodo }) => (
```
Volodymyr-Melnyk-Fron10der, line 8: use desctructure assignment
```suggestion
export const TodoItem: React.FC<Props> = ({ tempTodo: { title, .... } }) => (
```

Volodymyr-Melnyk-Fron10der (changes requested):

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api/';
+
+// returns a promise resolved after a given delay
```
OleziO, line 4: Remove all comments.

OleziO (changes requested):

OleziO (changes requested):
After selecting completed - all todos become active
image
image
image

src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
+            />
+          </>
+        )}
+      </div>
+
+      {errorType ? (
+        <ErrorNotification
+          errorType={errorType}
+        />
+      ) : null}
```
TarasHoliuk, line 163: use [conditional rendering](https://react.dev/learn/conditional-rendering) instead of ternary

src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
+      const addedTodo = await addTodo(newTodoToFetch);
+
+      setTodos(prevTodos => [...prevTodos, addedTodo]);
+    } catch (error) {
+      setErrorType('add');
+    } finally {
+      setIsTodoAdding(false);
+      setTempTodo(null);
+    }
+  }, [title]);
```
TarasHoliuk, line 79:
```suggestion
  }, []);
```
pass `title` as an argument - anyway it's available in the Header component too (in this case function will not be created again after each title change)

src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
+
+  const hasActive = todos.some(todoItem => !todoItem.completed);
+
+  const getTodosFromServer = async () => {
+    try {
+      const response = await getTodos(USER_ID);
+
+      setTodos(response);
+    } catch (error) {
+      setErrorType('upload');
```
TarasHoliuk, line 50: Create enum for error types

src/App.tsx
TarasHoliuk: moreover - it's better to have `title` state in the Header component in this case

src/components/Header.tsx
```diff
@@ -0,0 +1,57 @@
...
+        disabled={!hasActive}
+      />
+
+      <form onSubmit={handleSubmit}>
+        <input
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={title}
+          onChange={(event) => setTitle(event.target.value)}
```
TarasHoliuk, line 51: create `handleSubmit` - it's good practice to pass a named function instead of anonymous

TarasHoliuk (changes requested):
GJ :+1: I've added a few minor comments

SanyaBratashchuk (approved):

PR #545, opened 2023-06-16
add task solution

Ryshon:
[DEMO LINK](https://Ryshon.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/error/error.tsx
+++ b/src/components/error/error.tsx
@@ -1,0 +1,36 @@
+import { useState } from 'react';
+import cn from 'classnames';
+
+ type Props = {
+   error: string;
+   closeError: () => void;
+ };
+
+export const Error: React.FC<Props> = ({ error, closeError }) => {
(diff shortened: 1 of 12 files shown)


src/components/error/error.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+ type Props = {
+   error: string;
+   closeError: () => void;
+ };
+
+export const Error: React.FC<Props> = ({ error, closeError }) => {
+  const [isVisible, setIsVisible] = useState(true);
+
+  setTimeout(() => {
```
Anastasiia-Svintsova, line 12: You should clear all timers on the component unmount to avoid unpredictable side effects https://bobbyhadz.com/blog/react-cleartimeout

Anastasiia-Svintsova (changes requested):
- These requirements are not fulfilled. There is no loader while todo is adding 
image

- Filters should always be aligned to the center. Check the original markup, please 
image

PR #547, opened 2023-06-19
add task solution

NikaNika12:
[DEMO LINK](https://NikaNika12.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,158 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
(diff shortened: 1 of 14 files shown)


NikaNika12:
здесь много ошибок которые я не знаю как исправить

sTorba24 (changes requested):
Good work!

so what you can do

1) Get completed
2) Call `removeTodo` function for every id

demetrxx (changes requested):
image
Seems like this 0 should not be there after clearing the completed todos

sTorba24 (approved):
Amazing!

PR #563, opened 2023-07-03
solution

annholovko:
[DEMO LINK](https://annholovko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,161 @@
-/* eslint-disable max-len */
+/* eslint-disable react-hooks/rules-of-hooks */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useMemo,
+  useEffect,
+  useState,
+} from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { createTodo, deleteTodo, getTodos } from './api/todos';
+import { TodoList } from './components/TodoList';
+import { TodoForm } from './components/TodoForm';
+import { Error } from './components/Error';
+import { Filters } from './types/Filters';
+import { Filter } from './components/Filter';
 
-const USER_ID = 0;
+const USER_ID = 10892;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filter, setFilter] = useState(Filters.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+
+  const removeTodo = useCallback(async (todoId: number) => {
+    try {
+      setLoadingTodo(prevTodoId => [...prevTodoId, todoId]);
+      await deleteTodo(todoId);
+      setTodos(prevTodos => prevTodos.filter(todo => todo.id !== todoId));
+    } catch {
+      setError('Unable to delete a todo');
+    } finally {
+      setTempTodo(null);
```
ulositska, line 106: Why do we need to set it to null? I don't think, we're changing it during remove 💭

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+      setLoadingTodo([0]);
+    }
+  }, []);
+
+  const handleclearCompletedTodo = () => {
+    completedTodos.forEach(async (todo) => {
+      await removeTodo(todo.id);
+    });
+  };
+
```
ulositska, line 116: woww veeery nice reuse of the code 🤩

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
 
-const USER_ID = 0;
+const USER_ID = 10892;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filter, setFilter] = useState(Filters.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodo, setLoadingTodo] = useState([0]);
```
ulositska, line 25: That's not a suggestion, only a question. Why did you decide to store the ids of loading todos? During student review, I've seen the other person also do it, but can't wrap my head around why))

ulositska (approved):
Nailed it 🥇 
Reading you code, everything makes sense, app works as it should, so great job! Especially, liked  the code optimization and thoughtful use of async functions:brain::brain:

src/App.tsx
ulositska: nvm, I'm doing today's task and now I GET IT 🤦

annholovko: I also tried to do it without loadindgTodo state and didn't understand why it works in a wrong way😁

src/App.tsx
annholovko: U are absolutely right! I just copied this part from addTodo and forgot remove it🤦🏽‍♀️

silchyn (approved):

PR #565, opened 2023-07-04
add task solution

KatyaZaremska:
[DEMO LINK](https://KatyaZaremska.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,85 @@
+import { deleteTodo } from '../api/todos';
+import { Filter } from '../types/Filter';
+import { Todo } from '../types/Todo';
+
+interface FooterProps {
+  todos: Todo[],
+  filter: Filter,
+  handleFilter: (filter: Filter) => void,
+  handleError: (error: string) => void,
+  handleIsUpdating: (status: boolean) => void,
+  handleUpdatingIds: (ids: number[]) => void,
+}
+
+export const Footer = ({
+  todos, filter, handleFilter, handleError, handleIsUpdating, handleUpdatingIds,
+}: FooterProps) => {
+  const hasCompleted = todos.some(todo => todo.completed);
+  const todosLeft = todos.filter(todo => !todo.completed).length;
+
+  const handleCleaner = () => {
+    handleUpdatingIds([]);
+    handleIsUpdating(false);
+  };
+
+  const handleDeleteCompleted = () => {
+    handleIsUpdating(true);
+    const completedIds = todos
+      .filter(todo => todo.completed)
+      .map(todo => todo.id);
+
+    handleUpdatingIds(completedIds);
+
+    completedIds.map(id => deleteTodo(id)
+      .catch(() => handleError('Unable to delete a todo'))
+      .finally(() => {
+        if (id === completedIds[completedIds.length - 1]) {
+          handleCleaner();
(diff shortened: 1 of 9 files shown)


ivankhrsh:
Error message should disappear after 3 seconds, on your demo it's constant

src/components/TodoList.tsx
```diff
@@ -0,0 +1,101 @@
...
+          id, title, completed,
+        } = todo;
+
+        return (
+          <div className={`todo ${completed ? 'completed' : ''}`} key={id}>
+            <label className="todo__status-label">
+              <input
+                type="checkbox"
+                className="todo__status"
+                // checked={completed}
```
ivankhrsh, line 60: it's better to delete comments before pushing code

src/components/Footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+  handleFilter: (filter: Filter) => void,
+  handleError: (error: string) => void,
+  handleIsUpdating: (status: boolean) => void,
+  handleUpdatingIds: (ids: number[]) => void,
+}
+
+export const Footer = ({
+  todos, filter, handleFilter, handleError, handleIsUpdating, handleUpdatingIds,
+}: FooterProps) => {
+  const hasCompleted = todos.some(todo => todo.completed);
```
ivankhrsh, line 17: You can compare todosLeft and todos length, if it's different - we have completed todos. So you can get this value without iteration

src/components/Footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </a>
+
+        <a
+          href="#/completed"
+          className={`filter__link ${filter === Filter.Completed ? 'selected' : ''}`}
+          onClick={() => handleFilter(Filter.Completed)}
+        >
+          Completed
+        </a>
+      </nav>
```
ivankhrsh, line 72: You can map your Filter.ts enum like that to reduce repeating code

`Object.values(Filter).map((filter) => ...some code)`

ivankhrsh:
When there is no 'Clear completed' button others buttons stick to right corner, but should be at center.


image

image

src/components/Notification.tsx
```diff
@@ -0,0 +1,21 @@
...
+/* eslint-disable jsx-a11y/control-has-associated-label */
+interface NotificationProps {
+  error: string,
+}
+
+export const Notification = ({ error }:NotificationProps) => {
+  const [isHidden, setIsHidden] = useState(false);
+
+  return (
+    <div className={`notification is-danger is-light has-text-weight-normal ${isHidden ? 'hidden' : ''}`}>
```
ivankhrsh, line 12: it's better to move condition to separate variable and also use correct code style with line breaks

src/components/TodoList.tsx
```diff
@@ -0,0 +1,101 @@
...
+      case Filter.All:
+        return true;
+      case Filter.Active:
+        return !todo.completed;
+      case Filter.Completed:
+        return todo.completed;
+      default:
+        throw new Error('wrong filter selected');
+    }
+  });
```
ivankhrsh, line 37: Maybe it's better to filter in app and pass visibuleTodos

ivankhrsh (changes requested):

silchyn (approved):

PR #571, opened 2023-07-04
Develop

ulositska:
[DEMO LINK](https://ulositska.github.io/react_todo-app-add-and-delete/)


silchyn (changes requested):
> disable the input until receiving a response from the API

Don't clear it. Also, why is everything loading when adding new todo?

silchyn (approved):

PR #572, opened 2023-07-04
solution

dimoshen:
[DEMO LINK](https://dimoshen.github.io/react_todo-app-add-and-delete/)


silchyn (changes requested):
> - show an independent TodoItem after the list if tempTodo is not null;
> 
> - temp TodoItem should have the loader (check the original markup);
>

silchyn (approved):

PR #574, opened 2023-07-04
todos beta ver

vitvadym:
[DEMO LINK](https://vitvadym.github.io/react_todo-app-add-and-delete/)


silchyn (changes requested):
There is no loader and temp todo

silchyn (changes requested):
> disable the input until receiving a response from the API

PR #584, opened 2023-07-05
Develop

fenderukr:
244a8c50507ba964eff5041b33725f2e83ee5e0c

https://fenderukr.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,134 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import classNames from 'classnames';
+import { FilterPanel } from './FilterPanel';
+import { InputForm } from './InputForm';
+import { Notifications } from './Notifications';
+import { TodoList } from './TodoList';
 import { UserWarning } from './UserWarning';
+import { client } from './utils/fetchClient';
+import { Todo } from './types/Todo';
+import { FilterType } from './types/FilterTypeEnum';
 
-const USER_ID = 0;
+const USER_ID = 10592;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [removeTodoIsClicked, setRemoveTodoIsClicked] = useState(false);
+  const [editTodoIsClicked, setEditTodoIsClicked] = useState(false);
+  const [filterMode, setFilterMode] = useState<FilterType>(FilterType.All);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [fetchTodosStatus, setFetchTodosStatus] = useState(false);
+  const [onEmptyFormSubmit, setOnEmptyFormSubmit] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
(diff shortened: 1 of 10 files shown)


src/Notifications.tsx
```diff
@@ -0,0 +1,69 @@
...
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+
+    return undefined;
+  }, [isNotificationHidden]);
+
+  return (
+    <div className="notification is-danger is-light has-text-weight-normal">
```
lerastarynets, line 59: The notification div should be conditionally rendered based on the 'isNotificationHidden' state. If 'isNotificationHidden' is true, the notification should not be displayed.

src/Notifications.tsx
```diff
@@ -0,0 +1,69 @@
...
+    setIsNotificationHidden] = useState(false);
+
+  const hideNotification = () => {
+    setIsNotificationHidden(true);
+    setRemoveTodoIsClicked(false);
+    setEditTodoIsClicked(false);
+    setOnEmptyFormSubmit(false);
+  };
+
+  const showErrorText = () => {
```
lerastarynets, line 30: The function 'showErrorText' should return a default error message in case none of the conditions match. This is to ensure that an error message is always shown when the notification is visible.

src/utils/TodoItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <div
+      className={classNames('todo', {
+        completed: todo.completed,
+      })}
+      onDoubleClick={onDoubleClick}
+
+    >
+      <label className="todo__status-label">
+        <input type="checkbox" className="todo__status" checked />
```
lerastarynets, line 30: The checkbox input is always checked. It should be dependent on the 'completed' status of the todo. Use `checked={todo.completed}` instead of `checked`.

src/InputForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+            },
+            body: JSON.stringify(todoData),
+          },
+        );
+
+        const result = await response.json();
+
+        setFilteredTodos((prevTodos) => [...prevTodos, result]);
+      } catch (error) {
+        console.log(`An error occurred while adding a todo - ${error}`);
```
lerastarynets, line 52: Avoid using `console.log` for error handling. Consider using a more sophisticated logging mechanism or displaying the error to the user.

src/InputForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+      event.preventDefault();
+      if (!inputValue) {
+        setOnEmptyFormSubmit(true);
+
+        return;
+      }
+
+      const todoData = {
+        id: 0,
+        userId: 10592,
```
lerastarynets, line 30: Consider fetching the userId from a global state or user context to make the code more reusable and not hard-coded.

src/InputForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+        completed: false,
+        title: inputValue,
+      };
+
+      setTempTodo(todoData);
+
+      try {
+        const response = await fetch(
+          'https://mate.academy/students-api/todos?userId=10592', {
```
lerastarynets, line 39: Pls move api calls into `api/todos`

src/App.tsx
```diff
@@ -1,24 +1,134 @@
...
+        <div className="todoapp__content">
+          <header className="todoapp__header">
+            {/* this buttons is active only if there are some active todos */}
+            <button
+              type="button"
+              className={classNames('todoapp__toggle-all', {
+                active: allTodosCompleted,
+                inactive: notAllTodosCompleted,
+              })}
+            />
```
lerastarynets, line 84: The 'todoapp__toggle-all' button doesn't seem to have an onClick event attached to it. If this button is for toggling all todos, you should add a function that handles this.

PR #592, opened 2023-07-05
add and delete todo

YeMyr:
[DEMO LINK](https://yemyr.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,24 +1,128 @@
...
+      setTempTodo(null);
+    }
+  }, []);
+
+  const removeTodo = useCallback(async (todoId: number) => {
+    try {
+      setDeletingTodo(prevTodoId => [...prevTodoId, todoId]);
+      await deleteTodo(todoId);
+      setTodos(prevTodos => prevTodos.filter(todo => todo.id !== todoId));
+    } catch {
```
roman-mirzoian, line 68: For the `catch` block missing `error` to better understand what went wrong.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
```
roman-mirzoian, line 4: Try to avoid such commenting of the code, it is better to name the functions so that they explain what they do through their name, for example: `waitPromiseResolving` or `waitRequestResolving`.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+export function wait(delay: number) {
+  return new Promise(resolve => {
+    setTimeout(resolve, delay);
+  });
+}
+
+// To have autocompletion and avoid mistypes
+type RequestMethod = 'GET' | 'POST' | 'PATCH' | 'DELETE';
```
roman-mirzoian, line 12: Try to avoid unnecessary commenting of the code, here it is clear why we wrote the type :)

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+
+function request<T>(
+  url: string,
+  method: RequestMethod = 'GET',
+  data: any = null, // we can send any data to the server
```
roman-mirzoian, line 17: Try to avoid using `any` type for typing, it breaks the principle of using TypeScript.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+): Promise<T> {
+  const options: RequestInit = { method };
+
+  if (data) {
+    // We add body and Content-Type only for the requests with data
```
roman-mirzoian, line 22: Try to avoid unnecessary commenting of the code, here it is clear that we will pass to the body if the date exists :)

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(1300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
roman-mirzoian, line 34: If it happens that there is an error, then we do not report anything about what exactly this error is.

roman-mirzoian (changes requested):
The application works according to the task and works well, but the code itself needs to be improved a little.

PR #594, opened 2023-07-06
Develop

vladimirshub:
DEMO LINK](https://vladimirshub.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -20,7 +20,7 @@ export const App: React.FC = () => {
 
   useEffect(() => {
     getTodos(USER_ID)
-      .then((todosFromServer: Todo[]) => setTodos(todosFromServer))
+      .then(setTodos)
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,24 +1,155 @@
...
+
+        {todos.length > 0 && (
+          <footer className="todoapp__footer">
+            <span className="todo-count">
+              {`${activeTodos.length} items left`}
+            </span>
+
+            <Filter filter={filter} setFilter={setFilter} />
+
+            {completedTodos.length > 0 && (
```
TarasHoliuk, line 137: it's better to add `visibility: hidden` (add a proper class) instead of using conditional rendering to save the place of the button in the layout and keep the filter centered:
image

just FYI - implement it in the next step task :)

TarasHoliuk (approved):
Amazing :fire:

PR #608, opened 2023-07-18
Dev

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import React, { useEffect, useState } from 'react';
 import { CSSTransition, TransitionGroup } from 'react-transition-group';
 import { Etodos, ResponseError } from './types/enum';
@@ -30,6 +29,7 @@ export const App: React.FC = () => {
   const [respError, setRespError] = useState<ResponseError>(ResponseError.NOT);
   const [isToggleActiveTodos, setIsToggleActiveTodos] = useState(true);
   const [isTodoInputDisabled, setIsTodoInputDisabled] = useState(false);
+  const [creatingTodoTitle, setCreatingTodoTitle] = useState('');
 
   const toggleTodosActive = () => {
     const promiseList = todos.map((todo) => {
@@ -66,7 +66,8 @@ export const App: React.FC = () => {
 
   const deleteTodo = (id: number) => {
     setIsTodoInputDisabled(true);
-    deleteTodoOnServer(id)
+
+    return deleteTodoOnServer(id)
       .then(() => {
         getTodos(user.id).then((todoList) => {
           setTodos(todoList);
@@ -139,6 +140,7 @@ export const App: React.FC = () => {
           setIsShowFooter={setIsShowFooter}
           isDisable={isTodoInputDisabled}
           setIsDisable={setIsTodoInputDisabled}
+          setCreatingTodoTitle={setCreatingTodoTitle}
         />
 
         <section className="todoapp__main">
@@ -151,13 +153,32 @@ export const App: React.FC = () => {
               >
                 <TodoItem
                   todo={todoObj}
-                  deleteTodo={deleteTodo}
                   key={todoObj.id}
+                  deleteTodo={deleteTodo}
                   updateTodo={updateTodo}
                 />
               </CSSTransition>
             ))}
+            {creatingTodoTitle && (
+              <CSSTransition
+                key={0}
+                timeout={300}
+                classNames="temp-item"
+              >
+                <TodoItem
+                  todo={{
+                    id: 0,
+                    userId: user.id,
+                    title: creatingTodoTitle,
+                    completed: false,
+                  }}
+                  deleteTodo={deleteTodo}
+                  updateTodo={updateTodo}
+                />
+              </CSSTransition>
+            )}
           </TransitionGroup>
+
         </section>
 
         {isShowFooter && (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,182 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+/* eslint-disable no-console */
+import React, { useEffect, useState } from 'react';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
+import { Etodos, ResponseError } from './types/enum';
+import { UserData } from './types/userData';
+import { Todo } from './types/Todo';
+import { deleteTodoOnServer, getTodos, updateTodoProp } from './api';
+import { TodoItem } from './components/Todo';
+import { TodoHeader } from './components/TodoHeader';
+import { Footer } from './components/Footer';
+import { Notification } from './components/Notification';
 
-const USER_ID = 0;
+const defaultUser = {
+  createdAt: '2023-07-10T13:09:53.578Z',
+  email: 'gookidoo@gmail.com',
+  id: 11038,
+  name: 'Віктор Булденко',
+  phone: null,
+  updatedAt: '2023-07-10T13:09:53.578Z',
+  username: null,
+  website: null,
+};
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [user] = useState<UserData>(defaultUser);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isUncomplete, setIsUncomplete] = useState<number>(0);
+  const [isShowFooter, setIsShowFooter] = useState<boolean>(true);
+  const [sortTodosBy, setSortTodosBy] = useState<Etodos>(Etodos.ALL);
+  const [respError, setRespError] = useState<ResponseError>(ResponseError.NOT);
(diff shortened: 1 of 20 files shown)


src/api.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const getTodosCompletedStatus = (
+  f: (arg: Todo[]) => void,
+  arg: boolean,
+) => {
+  return fetch(`${BASE_URL}/todos?completed=${arg}`)
+    .then((data) => data.json())
+    .then((todoList) => f(todoList))
+    .catch((error) => new Error(error.message));
+};
```
DarkMistyRoom, line 49: you have two files with API logic. remove redundant one

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+  website: null,
+};
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [user] = useState<UserData>(defaultUser);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isUncomplete, setIsUncomplete] = useState<number>(0);
```
DarkMistyRoom, line 26: this value is based on todos. No need to store it in a state - use a simple variable

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+  const [isShowFooter, setIsShowFooter] = useState<boolean>(true);
```
DarkMistyRoom, line 27: as well as this one

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+  const [sortTodosBy, setSortTodosBy] = useState<Etodos>(Etodos.ALL);
+  const [respError, setRespError] = useState<ResponseError>(ResponseError.NOT);
```
DarkMistyRoom, line 29:
```suggestion
  const [error, setError] = useState<ResponseError>(ResponseError.NOT);
```

DarkMistyRoom: don't complicate names

src/App.tsx
```diff
@@ -1,24 +1,203 @@
...
+  const [isToggleActiveTodos, setIsToggleActiveTodos] = useState(true);
+  const [isTodoInputDisabled, setIsTodoInputDisabled] = useState(false);
```
DarkMistyRoom, line 31: Consider using `isLoading` instead

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <div className="notification is-danger is-light has-text-weight-normal">
+      <button
+        type="button"
+        className="delete"
+        onClick={() => setRespError(ResponseError.NOT)}
+      />
+      {respError === ResponseError.ADD && 'Unable to add a todo'}
+      {respError === ResponseError.DELETE && 'Unable to delete a todo'}
+      {respError === ResponseError.UPDATE && 'Unable to update a todo'}
+      {respError === ResponseError.EMPTY && "Title can't be empty"}
```
DarkMistyRoom, line 24: better move this values to enum:
`export enum ResponseError {
  ADD = 'Unable to add a todo', ... }`

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,88 @@
...
+  setCreatingTodoTitle: (arg: string) => void;
+};
+
+export const TodoHeader: React.FC<Props> = ({
+  todos,
+  toggleTodosActive,
+  setIsShowFooter,
+  setRespError,
+  setTodos,
+  checkCompletedTodo,
```
DarkMistyRoom, line 26: why do you need to check it here? looks better to do it in the App

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,88 @@
...
+  userID,
```
DarkMistyRoom, line 27: no need to lift a state like this. better to create a method in the App and pass it here

DarkMistyRoom (changes requested):
- try to push `package.json` to fix this error

image

I'm reviewing only the `add and delete` part - it works fine

src/api.ts
victor-buldenko: it's a file for the server request function, others use this functions.

src/App.tsx
victor-buldenko: you can't use variables in other components, because linter show error about reassigning variable

DarkMistyRoom (changes requested):
Good! Keep improving and welcome to conversation in the chat if you have anything to discuss

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <div className="notification is-danger is-light has-text-weight-normal">
+      <button
+        type="button"
+        className="delete"
+        onClick={() => setError(ResponseError.NOT)}
+      />
+      {error === ResponseError.ADD && ResponseError.ADD}
+      {error === ResponseError.DELETE && ResponseError.DELETE}
+      {error === ResponseError.UPDATE && ResponseError.UPDATE}
+      {error === ResponseError.EMPTY && ResponseError.EMPTY}
```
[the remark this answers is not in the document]
victor-buldenko: genius!)

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const todoFormHandler = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    setIsLoading(true);
+
+    if (!todoInput.trim()) {
+      setIsLoading(false);
+
+      return setError(ResponseError.EMPTY);
+    }
+
```
etojeDenys, line 37:
```suggestion
   if (!todoInput.trim()) {
      setIsLoading(false);

      return setError(ResponseError.EMPTY);
    }

    setIsLoading(true);
```

to avoid redundant setIsLoading invokes

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,72 @@
...
+    setCreatingTodoTitle(todoInput);
+
+    headerAddTodo(todoInput);
+
+    return setTodoInput('');
```
etojeDenys, line 42:
```suggestion
   setTodoInput('');
```
why do we need to return this function

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  const todoInputHandler = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setError(ResponseError.NOT);
+    setTodoInput(e.target.value);
+  };
+
+  return (
+    <header className="todoapp__header">
+      {Boolean(todos.length) && (
```
etojeDenys, line 52:
```suggestion
      {!!todos.length && (
```

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+  };
+
+  const checkCompletedTodo = (arr: Todo[]) => {
+    isUncomplete = 0;
+    for (let i = 0; i < arr.length; i += 1) {
+      if (!arr[i].completed) {
+        isUncomplete += 1;
+      }
+    }
+  };
```
etojeDenys, line 68: we can move this logic to useMemo, and remove all function calls

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+      })
+      .catch(() => setError(ResponseError.ADD))
+      .finally(() => setIsLoading(false));
+  };
+
+  const deleteCompletedTodo = () => {
+    getTodos(user.id, 'completed=true')
+      .then((todoList) => {
+        return Promise.all(
+          todoList.map((todo: Todo) => deleteTodoOnServer(todo.id)),
```
etojeDenys, line 89: instead of making redundant request to BE we can filter out todos array

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+      case Etodos.ACTIVE:
+        return todos.filter((todo) => todo.completed === false);
+
+      case Etodos.COMPLETED:
+        return todos.filter((todo) => todo.completed === true);
+
+      default:
+        return [...todos];
+    }
+  };
```
etojeDenys, line 127: we can also move this logic to useMemo, and remove the function call

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,214 @@
...
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos(user.id).then((todosList) => {
+      checkCompletedTodo(todosList);
+      setTodos(todosList);
+      setIsLoading(false);
+    }).catch(err => new Error(err.message))
+      .finally(() => setIsLoading(false));
```
loralevitska, line 150: You already have `setIsLoading(false);` in `finally` case, do it's redundant for `then`
```suggestion
    getTodos(user.id).then((todosList) => {
      checkCompletedTodo(todosList);
      setTodos(todosList);
    }).catch(err => new Error(err.message))
      .finally(() => setIsLoading(false));
```

loralevitska (approved):
Good job!

PR #616, opened 2023-07-21
sol

GGsmou:
[DEMO LINK](https://GGsmou.github.io/react_todo-app-add-and-delete/)

All comments and redundant func calls are left for future tasks)


src/variables.ts
```diff
@@ -0,0 +1 @@
+export const USER_ID = 11106;
```
And678, line 1: Maybe `constants.ts`?)

And678 (approved):
GJ, :+1:

PR #625, opened 2023-07-26
finished the task

artkas03:
[DEMO LINK](https://artkas03.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/TodosContext.tsx
+++ b/src/TodosContext.tsx
@@ -1,0 +1,224 @@
+/* eslint-disable no-console */
+import React, { useEffect, useState } from 'react';
+import { Todo } from './services/types';
+import { Filter } from './services/enums';
+import {
+  addTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from './api/todos';
+
+const USER_ID = 11138;
+
+interface FilterParams {
+  filterBy?: Filter,
+}
+
+function filterTodosByCompleted(todos: Todo[], filter: Filter): Todo[] {
+  let todosCopy = [...todos];
+
+  switch (filter) {
+    case (Filter.ACTIVE): {
+      todosCopy = todosCopy.filter(todo => !todo.completed);
+      break;
+    }
+
+    case (Filter.COMPLETED): {
+      todosCopy = todosCopy.filter(todo => todo.completed);
+      break;
+    }
+
+    default: {
+      break;
+    }
+  }
+
+  return todosCopy;
+}
+
+function filterTodos(todos: Todo[], { filterBy }: FilterParams): Todo[] {
+  let todosCopy = [...todos];
+
+  if (filterBy) {
+    todosCopy = filterTodosByCompleted(todos, filterBy);
+  }
+
+  return todosCopy;
+}
+
+function deleteAllTodos(todosIds: number[]): Promise<boolean> {
+  return new Promise((resolve) => {
+    todosIds.forEach(id => deleteTodo(id)
+      .then(() => resolve(true)));
+  });
+}
+
+interface ContextProps {
+  todos: Todo[],
+  visibleTodos: Todo[],
+  todoAdd: (newQuery: string) => Promise<boolean>,
+  handleAllCompletedToggle: (isAllActive: boolean) => void,
+  clearAllCompleted: () => void,
+  todoChange: (newTodo: Todo) => Promise<boolean>,
+  todoDelete: (todoId: number) => Promise<boolean>,
+  isTodosHasCompleted: () => boolean,
+  isEveryTodoCompleted: () => boolean,
+  filterBy: Filter,
+  setFilterBy: (newFilter: Filter) => void,
+  errorMessage: string,
+  areCompletedDeletingNow: boolean,
+  tempTodo: Todo | null,
+  setTempTodo: (newTodo: Todo | null) => void,
+}
+
+export const TodosContext = React.createContext<ContextProps>({
+  todos: [],
+  visibleTodos: [],
+  todoAdd: () => new Promise<boolean>(() => {}),
+  handleAllCompletedToggle: () => {},
+  clearAllCompleted: () => {},
+  todoChange: () => new Promise<boolean>(() => {}),
+  todoDelete: () => new Promise<boolean>(() => {}),
+  isTodosHasCompleted: () => false,
+  isEveryTodoCompleted: () => false,
+  filterBy: Filter.ALL,
+  setFilterBy: () => {},
+  errorMessage: '',
+  areCompletedDeletingNow: false,
+  tempTodo: null,
+  setTempTodo: () => {},
+});
+
+interface ProviderProps {
+  children: React.ReactNode,
+}
+
+export const TodosProvider: React.FC<ProviderProps> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState(Filter.ALL);
+  const [areCompletedDeletingNow, setAreCompletedDeletingNow]
+    = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const visibleTodos = filterTodos(todos, { filterBy });
+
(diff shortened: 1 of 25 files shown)


src/components/TodoErrors/TodoErrors.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { TodosContext } from '../../TodosContext';
+
+export const TodoErrors: React.FC = () => {
+  const {
+    errorMessage,
+  } = useContext(TodosContext);
+
+  const handleOnClick = (event: FormEvent<HTMLButtonElement>) => {
+    event.currentTarget.parentElement?.classList.add('hidden');
+  };
```
adv555, line 12: might be better to use setErrorMessage from context
```
() => setErrorMessage('')
```

src/TodosContext.tsx
```diff
@@ -0,0 +1,209 @@
...
+    getTodos(USER_ID).then(setTodos);
+  }, []);
+
+  const handleErrorOccuring = (errorTitle: string) => {
+    setErrorMessage(errorTitle);
+
+    setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
+  };
```
adv555, line 112: might be better to move it to TodoErrors component, because you will use it not only on todo loading
create a timer and clear the timer before the effect runs again
```
useEffect(() => {
    const timeout = setTimeout(closeNotification, 3000);

    return () => clearTimeout(timeout);
  }, [error]);
```

src/TodosContext.tsx
```diff
@@ -0,0 +1,209 @@
...
+    todoDelete,
+    isTodosHasCompleted,
+    isEveryTodoCompleted,
+    filterBy,
+    setFilterBy,
+    errorMessage,
+    areCompletedDeletingNow,
+    tempTodo,
+    setTempTodo,
+  };
```
adv555, line 202: The 'value' object passed as the value prop to the Context provider changes every render. 
Might be better to consider wrapping it in a useMemo and moving setState functions to a different provider, they wouldn't change each render

adv555 (approved):
Nice work!✨ 
same tiny comments from the previous PR

src/components/TodoErrors/TodoErrors.tsx
artkas03: why? I am handling here only classes to hide error visually but the existence of this block I am handling with useEffect that shows and then hides this block by setting error to an empty string

adv555: As for me, looks more clear 🤷
if you already have setErrorMessege, might be better to use it for all similar cases
In my opinion,  in React better to toggle or change class names by using state variables.

artkas03: @adv555 you might be right but I still can't get you))
I am controlling error existence in the context but in TodoErrors I just show it and handle actions with the button
Do you think that my code doesn't work with different errors or do you just mean that it would be better to do all actions with an error block in the error block?

adv555: It's just my thoughts, for me looks a bit tricky. 
I think it works perfectly.  I approved it from the beginning 😊

adv555 (approved):

src/TodosContext.tsx
```diff
@@ -0,0 +1,209 @@
...
+} from './api/todos';
+import { Todo } from './types';
+
+const USER_ID = 11138;
+
+interface FilterParams {
+  filterBy?: Filter,
+}
+
+function filterTodosByCompleted(todos: Todo[], filter: Filter): Todo[] {
```
anstsot, line 17: can be moved to `utils` folder

src/TodosContext.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+export const TodosProvider: React.FC<ProviderProps> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState(Filter.ALL);
+  const [areCompletedDeletingNow, setAreCompletedDeletingNow]
+    = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const visibleTodos = filterTodos(todos, { filterBy });
```
anstsot, line 100: use `useMemo` hook

src/TodosContext.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  useEffect(() => {
+    getTodos(USER_ID).then(setTodos);
```
anstsot, line 103: add error catching

src/TodosContext.tsx
```diff
@@ -0,0 +1,209 @@
...
+  }, []);
+
+  const handleErrorOccuring = (errorTitle: string) => {
+    setErrorMessage(errorTitle);
+
+    setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
```
anstsot, line 111: move to `TodoErrors` component on mount

src/TodosContext.tsx
```diff
@@ -0,0 +1,209 @@
...
+        title: normalizedQuery,
+        completed: false,
+      };
+
+      setTempTodo({ ...newTodo, id: 0 });
+
+      addTodo(newTodo)
+        .then((data) => {
+          setTodos([...todos, data]);
+          resolve(true);
```
anstsot, line 136:
```suggestion
          resolve();
```

src/TodosContext.tsx
```diff
@@ -0,0 +1,209 @@
...
+        })
+        .catch(() => {
+          handleErrorOccuring('Unable to add a todo');
+          reject(new Error(errorMessage));
+        });
+    })
+      .catch(error => error);
```
anstsot, line 143:
```suggestion
```

src/TodosContext.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  const todoDelete = (todoId: number) => {
+    return new Promise<boolean>((resolve, reject) => {
+      deleteTodo(todoId)
+        .then(() => {
+          setTodos(prevTodos => prevTodos.filter(todo => {
+            return todo.id !== todoId;
+          }));
+
+          resolve(true);
```
anstsot, line 154:
```suggestion
          resolve();
```

anstsot (changes requested):
Let’s add a few fixes to improve your code. Please add several completed todos to test `Clear competed` button

src/TodosContext.tsx
YegorKorenkov: Not fixed
image

src/TodosContext.tsx
YegorKorenkov: Not fixed
image

src/components/TodoErrors/TodoErrors.tsx
```diff
@@ -0,0 +1,42 @@
...
+          'is-light',
+          'has-text-weight-normal',
+          { hidde: isHidden },
+        )
+      }
+    >
+      <button
+        type="button"
+        className="delete"
+        onClick={() => setIsHidden(true)}
```
YegorKorenkov, line 37: the error is not hidden when clicking on the cross
image

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,71 @@
...
+        >
+          Completed
+        </a>
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        onClick={clearAllCompleted}
+        style={{ opacity: isTodosHasCompleted() ? 1 : 0 }}
```
YegorKorenkov, line 65: Please use display: none; or unmount this button from DOM, because now I can click on it
image

YegorKorenkov (changes requested):
Good job! Please fix minor issues

anstsot (approved):

src/components/TodoFooter/TodoFooter.tsx
artkas03: used not display: none but pointerEvents: none to not move filters to right side

PR #627, opened 2023-07-27
solution

vpdrabynko:
[DEMO LINK](https://vpdrabynko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,103 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import * as postService from './api/todos';
+import { Status } from './types/Status';
+import { Errors } from './components/Errors/Errors';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,24 +1,103 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosFilterBy, setTodosFilterBy] = useState<Status>(
+    Status.ALL,
+  );
+  const [isError, setIsError] = useState('');
+
+  const showError = (text: string) => {
+    setIsError(text);
+    setTimeout(() => {
```
Anastasiia-Svintsova, line 22: You should clear all timers on the component unmount to avoid unpredictable side effects https://bobbyhadz.com/blog/react-cleartimeout

Anastasiia-Svintsova (changes requested):
- Temp TodoItem should have the loader (check the original markup);
- Covered the todo with the loader while waiting for the API response on deleting;
- In case of API error show `Unable to delete a todo` notification at the bottom (the todo must stay in the list);
- The user shouldn't be able to add an empty todo 
chrome-capture-2023-6-27

maxim2310 (changes requested):
Almost perfect, just make sure that your loader is visible while deleting todo and would be nice to add loader while loading todos

lerastarynets (approved):
You've got this, keep pushing forward!

PR #628, opened 2023-07-27
add and delete todos - solution

MariiaMonakhova:
[DEMO LINK](https://MariiaMonakhova.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoContext.tsx
+++ b/src/components/TodoContext.tsx
@@ -1,0 +1,99 @@
+import React, { useEffect, useState } from 'react';
+import { TodosContextType } from '../types/TodoContext';
+import { Todo } from '../types/Todo';
+import { createTodo, deleteTodo, getTodos } from '../api/todos';
+import { ErrorType } from '../types/Error';
+
+const USER_ID = 11121;
+
+export const TodosContext = React.createContext<TodosContextType | null>(null);
+
+type Props = {
+  children: React.ReactNode,
+};
+
+export const TodosProvider: React.FC<Props> = ({ children }) => {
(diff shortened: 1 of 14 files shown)


src/components/TodoContext.tsx
```diff
@@ -0,0 +1,99 @@
...
+      todos,
+      error,
+      resetError,
+      addTodo,
+      removeTodo,
+      handleSetError,
+      disabledInput,
+      tempTodo,
+      isLoading,
+    }}
```
adv555, line 94: The 'value' object passed as the value prop to the Context provider changes every render.
Might be better to consider wrapping it in a useMemo and moving setState functions to a different provider, they wouldn't change each render

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+  function handleDeleteTodo(todoId: number) {
+    removeTodo(todoId);
+  }
+
+  return (
+    <div className={cn('todo', { completed })}>
+      <div className={cn('modal overlay', { 'is-active': isLoading })}>
+        <div className="modal-background has-background-white-ter" />
+        <div className="loader" />
+      </div>
```
adv555, line 30: When I'm trying to delete one todo, the loader works a bit incorrectly. 
it should be shown only on the selected todo, at the moment all todos have it.  
you can try something like that
```
<div className={classNames('modal overlay', {
        'is-active': isLoading && isSelected,
      })}
      >
```

adv555 (approved):
Nice work!✨ 
Check some comments below

anstsot (approved):

PR #674, opened 2023-08-01
Solution

kylyvnyk:
https://kylyvnyk.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,166 @@
-/* eslint-disable max-len */
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
+import { TodoForm } from './components/TodoForm';
+import { Status } from './types/Status';
+import { Footer } from './components/Footer';
+import { TodoList } from './components/TodoList';
+import { ErrorNotification } from './components/Error';
+import { Error } from './types/Error';
 
-const USER_ID = 0;
+const USER_ID = 11126;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>([]);
+  const [loading, setLoading] = useState(false);
+  const [errorNotification, setErrorNotification] = useState('');
+  const [status, setStatus] = useState(Status.All);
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
+    0,
+  ), [todos]);
+
+  const showErrorNotification = (notification: string): void => {
+    setErrorNotification(notification);
+
+    setInterval(() => {
+      setErrorNotification('');
+    }, 3000);
+  };
```
YegorKorenkov, line 35: 1. Move this logic to Error component
2. setInterval using incorrect here, because this callback function will be triggered every 3 seconds, use setTimeout instead
3. you need to delete the timer if a new error occurs

src/App.tsx
```diff
@@ -1,24 +1,166 @@
...
+      case Status.Completed:
+        return visibleTodos.filter(todo => todo.completed);
+
+      case Status.Active:
+        return visibleTodos.filter(todo => !todo.completed);
+
+      default:
+        return visibleTodos;
+    }
+  }
```
YegorKorenkov, line 109: Please move this function to _utils_ folder. you don't need it in the component because it takes up extra space and it will be recreated with each re-render

src/components/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </a>
+      </nav>
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        onClick={removeCompleted}
+        disabled={!completedTodo}
+      >
+        Clear completed
+      </button>
```
YegorKorenkov, line 66: Please hide this button, if there are no completed todos
image

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+    userId,
+    setNotification,
+    tempTodo,
+  }) => {
+    const [title, setTitle] = useState<string>('');
+
+    const handelSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
+
+      if (!title) {
```
YegorKorenkov, line 26: I have ability to add empty todo with spaces only
image

src/components/TodoList.tsx
```diff
@@ -0,0 +1,35 @@
...
+          todo={todo}
+          key={todo.id}
+          deleteTodos={deleteTodos}
+          removingId={removingId}
+        />
+      ))}
+      {tempTodo && (
+        <TodoElement
+          todo={tempTodo}
+        />
```
YegorKorenkov, line 31: Please add the loader to temp todo
image

YegorKorenkov (changes requested):

anstsot (approved):

PR #682, opened 2023-08-04
Develop

LiliiaDanylenko:
[DEMO LINK](https://LiliiaDanylenko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,212 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import classNames from 'classnames';
+import { Todo } from './types/Todo';
+import { FILTERS } from './types/FILTERS';
+import * as todoService from './api/todos';
+import { USER_ID } from './utils/constants';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { TodoList } from './components/TodoList';
+import { Filter } from './components/Filter';
+import { NewTodo } from './components/NewTodo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [filter, setFilter] = useState(FILTERS.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [loading, setLoading] = useState<number[]>([]);
+
+  const addTodo = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    setErrorMessage('');
+
+    if (query.trim() === '') {
+      setErrorMessage('Title can\'t be empty');
+      setQuery('');
+
+      return;
+    }
+
+    setIsAdding(true);
+    setTempTodo({
+      id: 0,
+      userId: USER_ID,
+      title: query.trim(),
+      completed: false,
+    });
+    todoService.createTodo({
+      userId: USER_ID,
+      title: query.trim(),
+      completed: false,
+    })
+      .then(newTodo => {
+        setQuery('');
+        setTodos((currentTodos: Todo[]) => [...currentTodos, newTodo]);
+      })
+      .catch(error => {
+        setErrorMessage('Unable to add a todo');
+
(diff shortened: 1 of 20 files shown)


src/App.tsx
```diff
@@ -1,24 +1,212 @@
...
+
+            <Filter
+              filter={filter}
+              setFilter={setFilter}
+            />
+
+            <button
+              type="button"
+              className="todoapp__clear-completed"
+              disabled={complitedTodos === 0}
```
TarasHoliuk, line 189:
```suggestion
              disabled={completedTodos === 0}
```
typo

src/App.tsx
```diff
@@ -1,24 +1,212 @@
...
+            >
+              Clear completed
+            </button>
+          </footer>
+        )}
+      </div>
+
+      <div className={classNames(
+        'notification is-danger is-light has-text-weight-normal',
+        { hidden: errorMessage.length === 0 },
```
TarasHoliuk, line 200:
```suggestion
        { hidden: !errorMessage },
```
empty string is a falsy value: https://www.freecodecamp.org/news/falsy-values-in-javascript/

src/App.tsx
```diff
@@ -1,24 +1,212 @@
...
+        setTodos(currentTodos => (
+          currentTodos.filter(currentTodo => currentTodo.id !== todoId)
+        ));
+      })
+      .catch(() => setErrorMessage('Unable to delete a todo'))
+      .finally(() => setLoading([]));
+  };
+
+  const clearCompletedTodos = () => {
+    todos.filter(todo => {
```
TarasHoliuk, line 75: the filter is not the correct method here
generally, you are removing some todos for sure) but not because of `filter`. you may use for each for such cases. But even better is using `map` and Promise.all (google it, or ask in the chat if you need it)

src/App.tsx
```diff
@@ -1,24 +1,212 @@
...
+      completed: false,
+    })
+      .then(newTodo => {
+        setQuery('');
+        setTodos((currentTodos: Todo[]) => [...currentTodos, newTodo]);
+      })
+      .catch(error => {
+        setErrorMessage('Unable to add a todo');
+
+        throw error;
```
TarasHoliuk, line 52: do not throw it - it will case page crash
you already handle it and show message

src/App.tsx
```diff
@@ -1,24 +1,212 @@
...
+import { NewTodo } from './components/NewTodo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [filter, setFilter] = useState(FILTERS.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [loading, setLoading] = useState<number[]>([]);
```
TarasHoliuk, line 20: that's perfect you have such logic :fire: 
let's just rename it to make name more descriptive

loadingTodoIds, etc

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,25 @@
...
+  addTodo, query, setQuery, isAdding,
+}) => {
+  return (
+    <form onSubmit={addTodo}>
+      <input
+        type="text"
+        className="todoapp__new-todo"
+        placeholder="What needs to be done?"
+        value={query}
+        onChange={event => setQuery(event.target.value)}
```
TarasHoliuk, line 20: it's better to move to a separate function above and use it here

TarasHoliuk (changes requested):
Indeed - very good job :+1: 
I added a few minor comments to make it even better

volodymyr-soltys97 (approved):
Nice work 🔥

src/App.tsx
```diff
@@ -68,11 +66,11 @@ export const App: React.FC = () => {
         ));
       })
       .catch(() => setErrorMessage('Unable to delete a todo'))
-      .finally(() => setLoading([]));
+      .finally(() => setIsProcessed([]));
   };
 
   const clearCompletedTodos = () => {
-    todos.filter(todo => {
+    todos.map(todo => {
```
TarasHoliuk, line 73: just FYI - the map would be the correct choice if you save the result (array of Promises) in the variable and put it into Promise.all. In this case, still `forEach` would be preferred

TarasHoliuk (approved):
:fire: Approved
Left 1 comment just FYI

PR #688, opened 2023-08-06
solution

IzzetYusufov:
[DEMO LINK](https://IzzetYusufov.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoBody/TodoBody.tsx
+++ b/src/components/TodoBody/TodoBody.tsx
@@ -1,0 +1,142 @@
+import classNames from 'classnames';
+import { useEffect, useRef, useState } from 'react';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  filteringBy: Todo[],
+  newTodoId: number[],
+  tempTodo: Omit<Todo, 'id'> | null,
+  deleteTodo: (todoId: number) => void,
+  updateTodo: (updatedTodo: Todo) => Promise<void>,
+};
+
+export const TodoBody: React.FC<Props> = ({
+  filteringBy,
+  newTodoId,
+  tempTodo,
+  deleteTodo,
+  updateTodo,
+}) => {
+  const [isEditing, setIsEditing] = useState<number | null>(null);
+  const [editValue, setEditValue] = useState('');
+  const editFocus = useRef<HTMLInputElement | null>(null);
+
+  // Activate input OnDoubleclick to edit the todo
+  useEffect(() => {
+    if (editFocus.current) {
+      editFocus.current.focus();
+    }
+  }, [editValue]);
+
+  // Submit the edited todo
+  const handleEditingTodo = (
+    event: React.FormEvent<HTMLFormElement>,
+    todo: Todo,
+  ) => {
+    event.preventDefault();
+
+    if (editValue === todo.title) {
+      return;
+    }
+
+    updateTodo({ ...todo, title: editValue })
+      .then(() => {
+        setIsEditing(null);
+        setEditValue('');
+      });
+  };
+
+  // Reset the edited todo on KeyUp "Escape"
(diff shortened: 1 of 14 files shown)


src/components/TodoBody/TodoBody.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const handleEditingTodo = (
+    event: React.FormEvent<HTMLFormElement> | null,
+    todo: Todo,
+  ) => {
+    if (event) {
+      event.preventDefault();
+    }
+
+    switch (editValue) {
+      case Errors.NULL:
```
volodymyr-soltys97, line 42: Comparing with Errors.NULL is not the right way to check if a string is empty. You should directly check if editValue is an empty string.

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,96 @@
...
+    if (value === filterBy) {
+      return;
+    }
+
+    setFilterBy(value);
+  };
+
+  // delete all complied todos
+  const deleteComplitedTodos = () => {
+    todos.map(todo => todo.completed && deleteTodo(todo.id));
```
volodymyr-soltys97, line 36: You should use 'forEach' instead of 'map' because 'map' is used to create a new array based on the result of the function. But here, you're not using the array that 'map' returns, so 'forEach' is more appropriate.

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,107 @@
...
+  setErrorMessage,
+  updateTodo,
+  addTodo,
+}) => {
+  const [inputValue, setInputValue] = useState('');
+  const [isDisabledInput, setIsDisabledInput] = useState(false);
+  const headerInputFocus = useRef<HTMLInputElement | null>(null);
+
+  // Autofocus on main input
+  useEffect(() => {
```
volodymyr-soltys97, line 32: The useEffect hook has 'todos' as a dependency and will re-run every time 'todos' changes. This could lead to unnecessary re-renders and affect performance. Consider alternative approaches, like using a callback ref to handle focusing.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
volodymyr-soltys97, line 34: The error thrown here is empty. It's a good practice to provide a descriptive error message or use the response status and status text from the fetch API.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      }
+
+      return response.json();
+    });
+}
+
+export const client = {
+  get: <T>(url: string) => request<T>(url),
+  post: <T>(url: string, data: any) => request<T>(url, 'POST', data),
```
volodymyr-soltys97, line 43: Instead of using 'any' for the 'data' parameter, consider creating an interface or type that describes the shape of the data you expect.

src/components/TodoBody/TodoBody.tsx
```diff
@@ -0,0 +1,154 @@
...
+      setIsEditing(null);
+      setEditValue('');
+    }
+  };
+
+  return (
+    <section className="todoapp__main">
+      {/* Todos from server response */}
+      {filteringBy.map(todo => (
+        <div
```
volodymyr-soltys97, line 67: Use a destructuring for `todo`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #693, opened 2023-08-10
solution

ShamievDavid:
[DEMO LINK](https://ShamievDavid.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,166 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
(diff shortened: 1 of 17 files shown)


etojeDenys (changes requested):
1. 1 item left
image


2. let's show loader only on items that adding or deleting

ShamievDavid:
* when i started to code updating of the todo, i understood that i should select whole todo, not only 'id' of the item.

src/App.tsx
```diff
@@ -1,24 +1,162 @@
...
+const USER_ID = 11298;
+
+const getFilterdTodos = (
+  todos: Todo[],
+  filterBy: Status,
+) => {
+  switch (filterBy) {
+    case Status.all:
+      return todos;
+
```
IvanFesenko, line 24: You may skip case when in return same value as default or put it before default

IvanFesenko (approved):
Well done

PR #697, opened 2023-08-14
add and delete

diana-tuz:
[DEMO LINK](https://diana-tuz.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,79 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
 
-const USER_ID = 0;
+import { UserWarning } from './UserWarning';
+import { TodosFilter } from './components/TodosFiltered/TodosFiltered';
+import { Todo } from './types/Todo';
+import { client } from './utils/fetchClient';
+import { TodoList } from './components/TodoList/TodoList';
+import { URL, USER_ID } from './utils/Url';
+import { TodoApp } from './components/TodoApp/TodoApp';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [allTodos, setAllTodos] = useState<Todo[]>([]);
+
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  useEffect(() => {
+    client.get(URL).then(data => {
+      const todosData = data as Todo[];
+
+      setTodos(todosData);
+      setAllTodos(todosData);
+    })
+      .catch(error => {
+        setErrorMessage('Unable to load todos');
+        throw new Error('An error occurred:', error);
+      });
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+
+        <TodoApp
+          setTodos={setTodos}
+          todos={todos}
+          setErrorMessage={setErrorMessage}
+        />
+        <section className="todoapp__main">
+          <TodoList
+            todos={todos}
+            setTodos={setTodos}
+            setErrorMessage={setErrorMessage}
+          />
+        </section>
+
+        {/* Hide the footer if there are no todos */}
+        <footer className="todoapp__footer">
+          <TodosFilter
+            todos={todos}
+            setTodos={setTodos}
+            allTodos={allTodos}
+          />
+        </footer>
+      </div>
+
+      {/* Notification is shown in case of any error */}
+      {/* Add the 'hidden' class to hide the message smoothly */}
+      <div className="notification is-danger is-light has-text-weight-normal">
+        <button type="button" className="delete" />
+
+        {/* show only one message at a time */}
+        {errorMessage
+          && (
+            <>
+              {errorMessage}
+            </>
+          )}
+      </div>
+    </div>
   );
 };
--- a/src/components/TodoApp/TodoApp.tsx
+++ b/src/components/TodoApp/TodoApp.tsx
@@ -1,0 +1,89 @@
+import React, { useState } from 'react';
+import { client } from '../../utils/fetchClient';
+import { URL, USER_ID } from '../../utils/Url';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  setTodos: (a: Todo[]) => void;
+  setErrorMessage: (a: string) => void;
+  todos: Todo[];
+};
+
+export const TodoApp: React.FC<Props> = ({
+  todos,
+  setTodos,
+  setErrorMessage,
+}) => {
+  const [value, setValue] = useState<string>('');
+
+  const handleAddTodo = (event: React.ChangeEvent<HTMLInputElement>) => {
+    event.preventDefault();
+    const newValue: string = event.target.value;
+
+    setValue(newValue);
+  };
+
+  const addNewTodo = async () => {
+    if (value.trim() === '') {
+      return;
+    }
+
+    try {
+      const newTodoData = {
+        title: value,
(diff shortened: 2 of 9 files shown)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+    setIsEditing(true);
+  };
+
+  return (
+    <li
+      className={classNames('todo', {
+        completed,
+        editing: isEditing,
+      })}
+      key={id}
```
anastasiiavorobiova, line 22: The 'key' prop is unnecessary on this 'li' element. 'key' is special in React and should be provided to the elements inside a list to give the elements a stable identity.

src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
+        </footer>
+      </div>
+
+      {/* Notification is shown in case of any error */}
+      {/* Add the 'hidden' class to hide the message smoothly */}
+      <div className="notification is-danger is-light has-text-weight-normal">
+        <button type="button" className="delete" />
+
+        {/* show only one message at a time */}
+        {errorMessage
```
anastasiiavorobiova, line 70: It's unnecessary to use a fragment here. You can output errorMessage directly without wrapping it in a fragment.

src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
+  useEffect(() => {
+    client.get(URL).then(data => {
+      const todosData = data as Todo[];
+
+      setTodos(todosData);
+      setAllTodos(todosData);
+    })
+      .catch(error => {
+        setErrorMessage('Unable to load todos');
+        throw new Error('An error occurred:', error);
```
anastasiiavorobiova, line 27: You're throwing an error in the catch block, but there's no handler for this error. It's better to handle this error or remove the throw statement.

src/components/TodosFiltered/TodosFiltered.tsx
```diff
@@ -0,0 +1,100 @@
...
+};
+
+export const TodosFilter: React.FC<Props> = ({ todos, setTodos, allTodos }) => {
+  enum Status {
+    ALL = 'all',
+    ACTIVE = 'active',
+    COMPLETED = 'Completed',
+  }
+
+  const [filterBy, setFilterBy] = useState<Status>(Status.ALL);
```
anastasiiavorobiova, line 19: The 'COMPLETED' status in the Status enum is not consistent with the two other statuses. It should be 'completed' instead of 'Completed'.

src/components/TodosFiltered/TodosFiltered.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const getFilteredTodos = async (filter: Status) => {
```
anastasiiavorobiova, line 21: The function 'getFilteredTodos' does not need to be async as there are no asynchronous operations in the function.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      }
+
+      return response.json();
+    });
+}
+
+export const client = {
+  get: <T>(url: string) => request<T>(url),
+  post: <T>(url: string, data: any) => request<T>(url, 'POST', data),
+  patch: <T>(url: string, data: any) => request<T>(url, 'PATCH', data),
```
anastasiiavorobiova, line 44: The 'data' parameter in the 'patch' method should have a specific type instead of 'any'.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // we wait for testing purpose to see loaders
+  return wait(300)
+    .then(() => fetch(BASE_URL + url, options))
+    .then(response => {
+      if (!response.ok) {
+        throw new Error();
```
anastasiiavorobiova, line 34: You are throwing an empty Error. It would be better to add some information to the error, like a message or the status code of the response.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      }
+
+      return response.json();
+    });
+}
+
+export const client = {
+  get: <T>(url: string) => request<T>(url),
+  post: <T>(url: string, data: any) => request<T>(url, 'POST', data),
```
anastasiiavorobiova, line 43: The 'data' parameter in the 'post' method should have a specific type instead of 'any'.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+}
+
+// To have autocompletion and avoid mistypes
+type RequestMethod = 'GET' | 'POST' | 'PATCH' | 'DELETE';
+
+function request<T>(
+  url: string,
+  method: RequestMethod = 'GET',
+  data: any = null, // we can send any data to the server
```
anastasiiavorobiova, line 17: Consider specifying a type instead of 'any' for the 'data' parameter to maintain type safety.

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { useState } from 'react';
+import { client } from '../../utils/fetchClient';
+import { URL, USER_ID } from '../../utils/Url';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  setTodos: (a: Todo[] | any) => void;
```
anastasiiavorobiova, line 7: Avoid using 'any' in TypeScript. It helps to maintain type safety throughout your codebase. Instead, strictly type your state updates to ensure that the function only ever receives an array of Todo objects.

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,89 @@
...
+    addNewTodo();
+  };
+
+  const handleKeyPress = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      handleSubmit(event as unknown as React.FormEvent<HTMLFormElement>);
+    }
+  };
+
+  const activeTodos = todos.map((todo: Todo) => !todo.completed);
```
anastasiiavorobiova, line 61: Your 'activeTodos' assignment is incorrect. The map function returns an array of boolean values, not an array of active todos. Use 'filter' instead of 'map' to get an array of active todos.

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  return (
+    <header className="todoapp__header">
+      {/* this buttons is active only if there are some active todos */}
+      {activeTodos
```
anastasiiavorobiova, line 66: The condition in the render function might be incorrect. It appears that you're trying to check if there are any active todos, but 'activeTodos' is an array. You should check if 'activeTodos.length > 0' instead.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,108 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+
+import { Todo } from '../../types/Todo';
+import { client } from '../../utils/fetchClient';
+
+type Props = {
+  todos: Todo[];
+  setTodos: (a: Todo[] | any) => void;
```
anastasiiavorobiova, line 9: Avoid using 'any' type in the function 'setTodos'. It would be better to strictly type it to avoid possible type errors.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,108 @@
...
+                id={`toggle-view-${id}`}
+                checked={checked}
+                onClick={handleChecked}
+              />
+            </label>
+
+            <span className="todo__title">{title}</span>
+
+            <button
+              type="button"
```
anastasiiavorobiova, line 75: This code does not show a loader while waiting for the API response when deleting a todo. It can be implemented by setting the 'isUpdating' state to true before the API call and setting it back to false after the call.

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const handleAddTodo = (event: React.ChangeEvent<HTMLInputElement>) => {
+    event.preventDefault();
+    const newValue: string = event.target.value;
+
+    setValue(newValue);
+  };
+
+  const addNewTodo = async () => {
+    if (value.trim() === '') {
```
anastasiiavorobiova, line 27:
```suggestion
    if (!value.trim()) {
```

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

diana-tuz:
sorry, I didn't see all comments at the first time(

anastasiiavorobiova (approved):
Fantastic work!

PR #702, opened 2023-08-17
add task solution

kbekher:
[DEMO LINK](https://kbekher.github.io/react_todo-app-add-and-delete/)


src/TodosContext.tsx
```diff
@@ -0,0 +1,64 @@
...
+      case Status.ACTIVE:
+        return todos.filter(todo => !todo.completed);
+
+      case Status.COMPLETED:
+        return todos.filter(todo => todo.completed);
+
+      default:
+        return todos;
+    }
+  };
```
etojeDenys, line 43: let's rewrite it using useMemo to use this as var instead of the func

etojeDenys (changes requested):

maxim2310 (approved):
Well done!!!

PR #705, opened 2023-08-24
add task solution

IvanVaverchak:
[DEMO LINK](https://IvanVaverchak.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,248 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import { createTodos, getTodos, removeTodos } from './api/todos';
+import { TodoFilter } from './components/TodoFilter';
+import { ErrorType } from './types/ErrorMessage';
 
-const USER_ID = 0;
+const USER_ID = 11356;
+
+function getFilteredTodos(
+  todos: Todo[],
+  sortBy: string,
+) {
+  const filteredTodos = todos;
+
+  switch (sortBy) {
+    case Status.ALL:
+      return filteredTodos;
+
+    case Status.ACTIVE:
+      return filteredTodos.filter(todo => todo.completed === false);
+
+    case Status.COMPLETED:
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,24 +1,248 @@
...
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage(ErrorType.LoadingError);
+        setIsMessageClosed(false);
+        setTimeout(() => {
+          setIsMessageClosed(true);
+        }, 3000);
+      });
+  }, [todos]);
```
maxim2310, line 55: here you problem, you get todos -> set it -> useEffect fired again (because of it dependency)

src/App.tsx
```diff
@@ -1,24 +1,248 @@
...
+
+  const filteredTodos = getFilteredTodos(todos, status);
+  const itemsLeft = getFilteredTodos(todos, Status.ACTIVE);
+  const completedTodos = getFilteredTodos(todos, Status.COMPLETED);
```
maxim2310, line 59: you dont need to filtered you todos in every render, you need to filter it only when user click on filter button

src/App.tsx
```diff
@@ -1,24 +1,248 @@
...
+        setErrorMessage(ErrorType.DeleteError);
+      })
+      .finally(() => {
+        setIsLoading(false);
+        setIsDeleting([]);
+      });
+  }
+
+  const ClearCompletedHendlere = (event: React.MouseEvent<HTMLButtonElement>) => {
+    todos.map(todo => todo.completed === true && deleteTodo(event, todo.id));
```
maxim2310, line 116:
```suggestion
    todos.forEach(todo => todo.completed === true && deleteTodo(event, todo.id));
```

maxim2310 (changes requested):
Good job, but  you have infinite requests 
image

src/App.tsx
maxim2310: not fixed

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

src/App.tsx
IvanVaverchak: I thought my function doesn't filter the array if Status.All
  const filteredTodos = todos;

  switch (sortBy) {
    case Status.ALL:
      return filteredTodos;

IvanFesenko (approved):

PR #707, opened 2023-08-24
Develop

Natalia-Tsiutsiura:
[DEMO LINK](https://Natalia-Tsiutsiura.github.io/react_todo-app-add-and-delete/)

Could you please give me a hint why the loading doesn't work when Im deleting and adding rows, and how to avoid double rows when Im adding them?

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,47 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
 
-const USER_ID = 0;
+import { TodosHeader } from './Components/TodosHeader';
+import { TodosFooter } from './Components/TodosFooter';
+import { Todoslist } from './Components/TodosList';
+import { TodosProvider } from './Components/TodosContext';
+import { UserWarning } from './UserWarning';
+import { ErrorNotification } from './Components/ErrorNotification';
+import { ErrorMessage } from './Enum/ErrorMessage';
+import { getTodos } from './api/todos';
+import { useTodo } from './Hooks/UseTodo';
+import { USER_ID } from './variables/userId';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const { setTodo, setIsError } = useTodo();
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(data => setTodo(data))
+      .catch(() => {
+        setIsError(ErrorMessage.LOADING);
+        setTodo([]);
+      });
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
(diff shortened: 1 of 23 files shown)


IvanFesenko (changes requested):
If you have questions - better ask help in fe_chat.
In hight level try to debug what is flow on adding todos, look like removing temTodo should be not in finally

src/Components/TodosContext.tsx
```diff
@@ -0,0 +1,63 @@
...
+  loading: false,
+  setLoading: () => [],
+});
+
+type Props = {
+  children: React.ReactNode;
+};
+
+export const TodosProvider: React.FC<Props> = ({ children }) => {
+  const [todo, setTodo] = useState<Todo[]>([]);
```
etojeDenys, line 26:
```suggestion
  const [todos, setTodos] = useState<Todo[]>([]);
```

etojeDenys (changes requested):
todos did not render when I reloaded the page, let's fix this case
image

src/Components/TodosFooter.tsx
```diff
@@ -0,0 +1,81 @@
...
+    setTodos,
+    filter,
+    setFilter,
+  } = useTodo();
+  const selectedTodos = todos.filter(todo => !todo.completed);
+  const compleatedTodosLength = todos.filter(todo => todo.completed).length;
+
+  const deleteTodos = () => {
+    setTodos(selectedTodos);
+  };
```
VitaliyBondarenko1982, line 18: clear completed button doesn't work correctly
you only remove todos on state, but not on server. after reload page completed todos stilll exists.
and should be visible loader for each todo which in delete process.

src/App.tsx
```diff
@@ -1,24 +1,56 @@
...
+import { getTodos } from './api/todos';
+import { useTodo } from './Hooks/UseTodo';
+import { USER_ID } from './variables/userId';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const { setTodos, setIsError, setLoading } = useTodo();
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [errorVisibility, setErrorVisibility] = useState(false);
+  const [processings] = useState<number[]>([]);
```
VitaliyBondarenko1982, line 18:
```suggestion
  const [processings, setProcessings] = useState<number[]>([]);
```
do you really not need setProcessings function for update processings state?
looks like this state can be used for manage todos which in some process on server (updating, deleting)

VitaliyBondarenko1982 (changes requested):
Need some improvements.

src/App.tsx
Natalia-Tsiutsiura: I planned to add it later

src/Components/TodosItems.tsx
```diff
@@ -0,0 +1,97 @@
...
+        getTodos(USER_ID)
+          .then((data) => {
+            setTodos(data);
+          })
+          .catch(() => {
+            setIsError(ErrorMessage.UPDATE);
+            setErrorVisibility(true);
+          })
+          .finally(() => setShowLoadind(false));
+      });
```
etojeDenys, line 39: let's create fetchTodos function and reuse it everywhere in this component

etojeDenys (approved):
looks good, small improvement:

PR #709, opened 2023-08-25
Added adding and deleting todos

vova-sherstov:
[DEMO LINK}(https://vova-sherstov.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,150 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { TodoForm } from './components/TodoForm';
+import { TodoList } from './components/TodoList';
+import { TodoFooter } from './components/TodoFooter';
+import { ErrorMessage } from './components/ErrorMessage';
+import { Todo } from './types/Todo';
+import { FilterOptions } from './types/FilterOptions';
+import * as postService from './api/todos';
+import { ErrorMessages } from './types/errorMessages';
+import { prepareTodos } from './utils/prepareTodos';
 
-const USER_ID = 0;
+const USER_ID = 11272;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterOption, setFilterOption] = useState(FilterOptions.All);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
+  const [isDisabled, setIsDisabled] = useState(false);
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState(ErrorMessages.None);
+
+  const preparedTodos = useMemo(() => {
+    return prepareTodos(todos, filterOption);
+  }, [todos, filterOption]);
+
+  const allTodoCompleted = todos.every(todo => todo.completed);
+  const hasTodo = todos.length > 0;
+
+  const activeTodosCount = todos
+    .filter(todo => !todo.completed)
+    .length;
+
+  const completedTodoIds = todos
+    .filter(todo => todo.completed)
+    .map(todo => todo.id);
+
+  useEffect(() => {
+    postService.getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage(ErrorMessages.Get);
+        setHasError(true);
+      });
+  }, []);
+
+  const addTodo = (title: string) => {
+    if (!title) {
+      setHasError(true);
+      setErrorMessage(ErrorMessages.EmptyTitle);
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,24 +1,150 @@
...
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
+  const [isDisabled, setIsDisabled] = useState(false);
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState(ErrorMessages.None);
+
+  const preparedTodos = useMemo(() => {
+    return prepareTodos(todos, filterOption);
```
Anastasiia-Svintsova, line 28: Consider renaming this function to show that it's an action
```suggestion
    return getPreparedTodos(todos, filterOption);
```

src/App.tsx
```diff
@@ -1,24 +1,150 @@
...
+  }, [todos, filterOption]);
+
+  const allTodoCompleted = todos.every(todo => todo.completed);
```
Anastasiia-Svintsova, line 31: In my opinion, it's better to check if there is at least one active todos and use some() method for better performance

src/App.tsx
```diff
@@ -1,24 +1,150 @@
...
+        })
+        .finally(() => {
+          setDeletingTodoIds([]);
+        });
+    });
+  };
+
+  useEffect(() => {
+    deleteTodo();
+  }, [deletingTodoIds]);
```
Anastasiia-Svintsova, line 100: Would be better to wrap the deleteTodo to useCallback() to memorize it. And change the dependencies for this useEffect. And you really need to do it with useEffect? 🤔  Seems that deleting should start on the button click 
```suggestion
    deleteTodo();
  }, [deleteTodo]);
```

src/App.tsx
```diff
@@ -1,24 +1,150 @@
...
+
+  useEffect(() => {
+    setTimeout(() => {
```
Anastasiia-Svintsova, line 103: You should clear all timers on the component unmount to avoid unpredictable side effects https://bobbyhadz.com/blog/react-cleartimeout

src/components/TodoForm/TodoForm.tsx
```diff
@@ -0,0 +1,62 @@
...
+}) => {
+  const [todoTitle, setTodoTitle] = useState('');
+
+  const titleInput = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (hasTodo) {
+      titleInput.current?.focus();
+    }
+  }, [isDisabled]);
```
Anastasiia-Svintsova, line 25: Don't you need to include `hasTodo` in dependencies?

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+            {selectedTodoId !== todo.id ? (
+              <span className="todo__title">{todo.title}</span>
+            ) : (
+              <form>
+                <input
+                  type="text"
+                  className="todo__title-field"
+                  placeholder="Empty todo will be deleted"
+                  value={todo.title}
```
Anastasiia-Svintsova, line 59: Use destructuring for objects for better code readability

Anastasiia-Svintsova (changes requested):
Keep going! 

- The user shouldn't be able to add a todo with an empty title 

https://github.com/mate-academy/react_todo-app-add-and-delete/assets/99403198/2c9f0fd7-1975-493b-8d8c-ba0a27786209

vova-sherstov:
Thank you for your feedback!
1. fixed adding a todo with an empty title;
2. renamed function;
3. allTodoCompleted replaced by hasActiveTodo using the some() method;
4. changed the todo delete method (I used useEffect hook because the deletion function passed in props did not work by ;clicking. Loader covered todo, but they were not deleted);
5. added clearTimeout() method and moved this useEffect hook to ErrorMessage component;

src/components/TodoForm/TodoForm.tsx
vova-sherstov: If i include hasTodo in dependencies the input field will be in focus immediately

TarasHoliuk: In this case, there should be one more variable to decide if you want to focus it
because current behavior is a bit unexpected when you read code

It is not a critical thing I guess - try to fix it in the next part of the task

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,100 @@
...
+            >
+              <div className="modal-background has-background-white-ter" />
+              <div className="loader" />
+            </div>
+          </div>
+        );
+      })}
+
+      {tempTodoTitle && (
+        <div className="todo temp-item-enter temp-item-enter-active">
```
TarasHoliuk, line 84: to avoid duplication and to separate more complex logic in the next task don't forget to move it to a separate component

TarasHoliuk (approved):
GJ :+1: 
Added a few minor comments
just take them into account in the next task

PR #710, opened 2023-08-25
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */
 /* eslint-disable consistent-return */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { FormEvent, useState } from 'react';
+import React, { FormEvent, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { Todo } from './types/Todo';
 import { SortBy } from './types/SortBy';
@@ -17,24 +18,30 @@ export const App: React.FC = () => {
   const [inputValue, setInputValue] = useState<string>('');
   const [isLoadingTodo, setIsLoadingTodo] = useState(false);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isTodoDeleted, setIsTodoDeleted] = useState(false);
+  const [makeAnyChange, setMakeAnyChange] = useState<boolean>(false);
   const {
     isLoading,
     todos,
     errorMessage,
-    todosNotCompleted,
     handleError,
-    handleIsTodoDeleted,
-  } = useGetTodos(USER_ID, tempTodo);
+  } = useGetTodos(USER_ID, makeAnyChange);
+
+  const todosNotCompleted = useMemo(() => {
+    return todos.filter(todo => todo.completed === false).length;
+  }, [todos, isTodoDeleted]);
 
   const deleteOneTodo = async (todoId: number) => {
+    setIsTodoDeleted(true);
     try {
       await deleteTodo(USER_ID, todoId);
     } catch (error) {
       handleError(Errors.delete);
     }
 
-    handleIsTodoDeleted(true);
+    setIsTodoDeleted(false);
     setSelectedTodo(prevSelectedTodo => [...prevSelectedTodo, todoId]);
+    setMakeAnyChange(!makeAnyChange);
   };
 
   const handleDeleteTodo = (value: number) => deleteOneTodo(value);
@@ -76,6 +83,7 @@ export const App: React.FC = () => {
     setIsLoadingTodo(false);
     setTempTodo(null);
     setInputValue('');
+    setMakeAnyChange(!makeAnyChange);
   };
 
   if (!USER_ID) {
@@ -88,7 +96,10 @@ export const App: React.FC = () => {
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          <button type="button" className="todoapp__toggle-all active" />
+          <button
+            type="button"
+            className="todoapp__toggle-all"
+          />
 
           <form onSubmit={handleAddTodo}>
             <input
--- a/src/components/Todos.tsx
+++ b/src/components/Todos.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-mixed-operators */
 import { useMemo } from 'react';
 import { SortBy, Todo } from '../types';
 
@@ -57,7 +58,7 @@ export const Todos: React.FC<Props> = ({
             </button>
 
             <div
-              className={(isLoading && selectedTodo.includes(todo.id))
+              className={((isLoading && selectedTodo.includes(todo.id)))
                 ? 'modal overlay is-active'
                 : 'modal overlay'}
             >
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -1,12 +1,11 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { getTodos } from './api/todos';
 import { Errors, Todo } from './types';
 
-export const useGetTodos = (USER_ID: number, tempTodo: Todo | null) => {
+export const useGetTodos = (USER_ID: number, makeAnyChange: boolean) => {
   const [isLoading, setIsLoading] = useState(false);
   const [todos, setTodos] = useState<Todo[]>([]);
   const [errorMessage, setErrorMessage] = useState<Errors>(Errors.noEroor);
-  const [isTodoDeleted, setIsTodoDeleted] = useState(false);
 
   useEffect(() => {
     setIsLoading(true);
@@ -23,12 +22,7 @@ export const useGetTodos = (USER_ID: number, tempTodo: Todo | null) => {
     }
 
     fetchTodos();
-    setIsTodoDeleted(false);
-  }, [isTodoDeleted, tempTodo]);
-
-  const todosNotCompleted = useMemo(() => {
-    return todos.filter(todo => todo.completed === false).length;
-  }, [todos, isTodoDeleted]);
+  }, [makeAnyChange]);
 
   useEffect(() => {
     if (errorMessage) {
@@ -42,16 +36,10 @@ export const useGetTodos = (USER_ID: number, tempTodo: Todo | null) => {
     setErrorMessage(error);
   };
 
-  const handleIsTodoDeleted = (value: boolean) => {
-    setIsTodoDeleted(value);
-  };
-
   return {
     isLoading,
     todos,
     errorMessage,
-    todosNotCompleted,
     handleError,
-    handleIsTodoDeleted,
   };
 };


src/App.tsx
```diff
@@ -1,24 +1,299 @@
...
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
 
-const USER_ID = 0;
+const USER_ID = 11361;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [updatedTodos, setUptatedTodos] = useState<Todo[]>([]);
```
Radoslaw-Czerniawski, line 11: This state is redundant, `todos` should be fine on it's own

src/App.tsx
```diff
@@ -1,24 +1,299 @@
...
+
+  useEffect(() => {
+    if (errorMessage) {
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, [errorMessage]);
+
+  useEffect(() => {
```
Radoslaw-Czerniawski, line 52: This whole useEffect is redundant, implement similar logic like in your previous task `react_dynamic-list-of-todos`,
where you filtered todos in `TodoList`

src/App.tsx
```diff
@@ -1,24 +1,299 @@
...
+  const [sortBy, setSortBy] = useState<string>('all');
+  const [isTodoDeleted, setIsTodoDeleted] = useState(false);
+  const [todosNotCompleted, setTodosNotCompleted] = useState(0);
+  const [isLoading, setIsLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<number[]>([]);
+  const [inputValue, setInputValue] = useState<string>('');
+  const [isLoadingTodo, setIsLoadingTodo] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
```
Radoslaw-Czerniawski, line 22: You could export this logic to a separate hook, for example you could create `useGetTodos` hook where you return an object holding your loading state and data 
image

src/App.tsx
```diff
@@ -1,24 +1,299 @@
...
-const USER_ID = 0;
+const USER_ID = 11361;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [updatedTodos, setUptatedTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [sortBy, setSortBy] = useState<string>('all');
+  const [isTodoDeleted, setIsTodoDeleted] = useState(false);
+  const [todosNotCompleted, setTodosNotCompleted] = useState(0);
```
Radoslaw-Czerniawski, line 15: There is no need for this to be a state, this could be a value memoized with useMemo based on `todos`

src/App.tsx
```diff
@@ -1,24 +1,299 @@
...
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
```
Radoslaw-Czerniawski, line 123: Overall your component here is too long, your components should be max 200 lines long (the shorter the better), split some of the jsx return here into separate components

Radoslaw-Czerniawski (changes requested):
Needs refactoring
- split your jsx into components,
- don't overuse state, some things could be regular or memoized variables
- if you have two or more codependent states / hooks, this suggest creating a custom hook

src/App.tsx
```diff
@@ -1,24 +1,193 @@
...
+
+  const deleteOneTodo = async (todoId: number) => {
+    try {
+      await deleteTodo(USER_ID, todoId);
+    } catch (error) {
+      handleError('Unable to delete a todo');
+    }
+
+    handleIsTodoDeleted(true);
+    setSelectedTodo(prevSelectedTodo => [...prevSelectedTodo, todoId]);
```
carlos-cne, line 35: deleleOneTodo function is setting the deleted todoId within the selectedTodo.

src/App.tsx
```diff
@@ -1,24 +1,193 @@
...
+  } = useGetTodos(USER_ID, tempTodo);
+
+  const deleteOneTodo = async (todoId: number) => {
+    try {
+      await deleteTodo(USER_ID, todoId);
+    } catch (error) {
+      handleError('Unable to delete a todo');
+    }
+
+    handleIsTodoDeleted(true);
```
carlos-cne, line 34: The todo could be removed right after the deleteTodo function resolves. You probably don't need this setState.

src/App.tsx
```diff
@@ -1,24 +1,193 @@
...
+
+  const handleDeleteTodo = (value: number) => deleteOneTodo(value);
+
+  const handleKeyDown = async (
+    event: React.KeyboardEvent<HTMLInputElement>,
+  ) => {
+    if (event.key === 'Enter') {
+      event.preventDefault();
+      if (!inputValue) {
+        handleError("Title can't be empty");
```
carlos-cne, line 46: On this statement, you should use a `return` or `else` statement to finish the function on this point in case there is no inputValue.

src/App.tsx
```diff
@@ -1,24 +1,193 @@
...
+      });
+
+      const todo = {
+        id: 0,
+        userId: 11361,
+        title: inputValue,
+        completed: false,
+      };
+
+      setInputValue('');
```
carlos-cne, line 65: This clear input value statement should be placed after the addTodo function. If the function throws an error the user will lose the input data.

src/App.tsx
```diff
@@ -1,24 +1,193 @@
...
+  ) => {
+    if (event.key === 'Enter') {
+      event.preventDefault();
+      if (!inputValue) {
+        handleError("Title can't be empty");
+      }
+
+      setIsLoadingTodo(true);
+
+      setTempTodo({
```
carlos-cne, line 51: The use of this `setTempTodo` seems you probably want to make some optimistic changes on the Todo list, but it could cause strange behavior on the UI because on the `finally` statement the state is set to null.
Also the state `tempTodo` is used as a parameter in the `useGetTodos` hook and this value is changed twice at least and for every change the `useGetTodos` hook is executed again.

src/api/todos.ts
```diff
@@ -0,0 +1,15 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
```
carlos-cne, line 5: As this function has only one statement, if you want, we could return directly without using the `return` keyword. Like this: 
```js
export const getTodos = (userId: number) => client.get<Todo[]>(`/todos?userId=${userId}`);
```

src/components/Todos.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    const isCompleted = sortBy === SortBy.completed;
+
+    return todos.filter(({ completed }) => completed === isCompleted);
+  }, [sortBy, todos]);
+
+  return (
+    <section className="todoapp__main">
+      {filteredTodos.map(todo => (
```
carlos-cne, line 33: [According to the guidelines](https://mate-academy.github.io/style-guides/htmlcss.html), we should let this component semantic using a `unordered list (ul)` instead of using `div`.

src/hooks.ts
```diff
@@ -0,0 +1,57 @@
+import { useEffect, useMemo, useState } from 'react';
+import { getTodos } from './api/todos';
+import { Todo } from './types';
+
+export const useGetTodos = (USER_ID: number, tempTodo: Todo | null) => {
```
carlos-cne, line 5: This hook is being used as a "logic container" so probably is better to use the `USER_ID` const inside this file and use all other functions that call the API here.
By doing this you won't need the control state anymore (isTodoDelete, handleTodoDelete...)

carlos-cne (changes requested):

src/App.tsx
Luk2asz: this function change "isTodoDeleted" wich is declared in hooks.ts. When user delete todo, then isTodoDeleted is changed to true. 
Later the getTodos function is called and after getting the data from the server "isLoading" is changed to true. Thanks to this, the "loader" is displayed on the removed todos until the server responds

src/App.tsx
Luk2asz: It works like requiments said: 
immediately after sending a request create a todo with id: 0 and save it to the tempTodo variable in the state (NOT to the todos array);
show an independent TodoItem after the list if tempTodo is not null;
temp TodoItem should have the loader (check the original markup);
tempTodo should to set as a new todo and then set to null when finally todo is added to todods, so it causes thath useGetTodos is executed twice

src/hooks.ts
Luk2asz: USER_ID, isTodoDeleted and errorMessage is used in APP.tsx in many places. There are functions in semantic code (onClick itp.) that uses these consts, so it easier to change them via handleError and handleIsTodoDeleted

src/App.tsx
Luk2asz: selectedTodo is a array contains id numbers of todo, wich will be deleted. It is a array, becouse of button "clear completed" which can delete more than one todo.

src/App.tsx
```diff
@@ -1,24 +1,193 @@
...
+    todosNotCompleted,
+    handleError,
+    handleIsTodoDeleted,
+  } = useGetTodos(USER_ID, tempTodo);
+
+  const deleteOneTodo = async (todoId: number) => {
+    try {
+      await deleteTodo(USER_ID, todoId);
+    } catch (error) {
+      handleError('Unable to delete a todo');
```
mykhalenych, line 31: create enum with errorsMessages and use enum instead of string

src/api/todos.ts
```diff
@@ -0,0 +1,17 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+// export const getTodos = (userId: number) => {
```
mykhalenych, line 4: delete commented code

mykhalenych (changes requested):
I can add empty todo, trim value, also you duplicate requests to server and error message after deleting isn't correct

https://github.com/mate-academy/react_todo-app-add-and-delete/assets/53228259/664139ca-b293-4d50-8140-523ce5ee54bc

OleziO (changes requested):

OleziO (changes requested):
Fix all comments.

mykhalenych (changes requested):
https://github.com/mate-academy/react_todo-app-add-and-delete/assets/53228259/4add5958-1f8c-4e60-9c67-65d7c86cb717

I still see duplicate requests after create/delete

Luk2asz:
> I still see duplicate requests after create/delete

I don't know how to fix it. Can you help me?

src/App.tsx
```diff
@@ -1,24 +1,186 @@
...
+      handleError(Errors.delete);
+    }
+
+    handleIsTodoDeleted(true);
+    setSelectedTodo(prevSelectedTodo => [...prevSelectedTodo, todoId]);
+  };
+
+  const handleDeleteTodo = (value: number) => deleteOneTodo(value);
+
+  const handleKeyDown = async (
```
Radoslaw-Czerniawski, line 41: This can be called on form submit rather than on pressing enter in the input
image
image

Radoslaw-Czerniawski (changes requested):
adjust already existing comments

src/App.tsx
```diff
@@ -1,24 +1,193 @@
...
+          type="button"
+          className="delete"
+          onClick={() => handleError('')}
+        />
+
+        {/* show only one message at a time */}
+        {errorMessage}
+      </div>
+    </div>
   );
```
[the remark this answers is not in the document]
OleziO: Not resolved

Luk2asz:
> Resolve all comments

Only unresolved comment is "I still see duplicate requests after create/delete", but in my opinion should be lik that. First request is add/delete todo, second request is generating and uptated todos list.

src/App.tsx
```diff
@@ -1,24 +1,187 @@
...
+          handleDeleteTodo={handleDeleteTodo}
+          isLoading={isLoading}
+          selectedTodo={selectedTodo}
+        />
+
+        <footer className="todoapp__footer">
+          <span className="todo-count">
+            {todosNotCompleted}
+            {' '}
+            items left
```
OleziO, line 118: Use template strings.

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

Radoslaw-Czerniawski (approved):
There is room for improvement, like:
- differentiating between error and no error using null
image
- implementing memoization with `useCallback` and `useMemo` 
- splitting logic into more components

However the functionality is there, and already a great deal of refactoring was done. No need for further blocking, approved.

PR #899, opened 2023-11-10
all tests pass

aniashev:
[DEMO LINK](https://aniashev.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,147 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Error } from './Error';
+import { Header } from './Header';
+import { Todo } from './types/Todo';
+import { TodoList } from './TodoList';
+import { Footer } from './Footer';
+import * as postService from './api/todos';
 
-const USER_ID = 0;
+export const USER_ID = 11894;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [todoTitle, setTodoTitle] = useState('');
+  const [deletingTodoId, setDeletingTodoId] = useState<number | null>(null);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,24 +1,147 @@
...
+import { TodoList } from './TodoList';
+import { Footer } from './Footer';
+import * as postService from './api/todos';
 
-const USER_ID = 0;
+export const USER_ID = 11894;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
```
etojeDenys, line 15: do we need a separate state for this, I think it's redundant

src/TodoList.tsx
```diff
@@ -0,0 +1,217 @@
...
+          >
+            {todo.title}
+          </span>
+
+          {/* Remove button appears only on hover */}
+          <button
+            type="button"
+            key={todo.id}
+            className="todo__remove"
+            data-cy="TodoDelete"
```
etojeDenys, line 61: you can create the TodoItem component and move this code there, it will look better

src/Footer.tsx
```diff
@@ -0,0 +1,98 @@
...
+          className={`filter__link ${filter === Status.ACTIVE ? 'selected' : ''}`}
+          data-cy="FilterLinkActive"
+          onClick={() => handleFilterChange(Status.ACTIVE)}
+        >
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={`filter__link ${filter === Status.COMPLETED ? 'selected' : ''}`}
```
etojeDenys, line 68: we usually use classnames package for such cases

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,158 @@
...
+
+    default:
+      return todos;
+  }
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [todoTitle, setTodoTitle] = useState('');
```
etojeDenys, line 30: looks like we can move this state to Header, and decrease the number of App re-renders

etojeDenys (changes requested):
nit:

anastasiia-tilikina (changes requested):
Good job!

check my comments.
also you need to add loader state for deleting todo while waiting server response for deletion.

VitaliyBondarenko1982 (approved):
well done

PR #903, opened 2023-11-12
React_Todo-App-Add-And-Delete

levchenko-dmytro:
[DEMO LINK](https://levchenko-dmytro.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoApp/TodoApp.tsx
+++ b/src/components/TodoApp/TodoApp.tsx
@@ -1,0 +1,190 @@
+/* eslint-disable react-hooks/exhaustive-deps */
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import classNames from 'classnames';
+import { TodosContext } from '../../TodosContext';
+import * as todoService from '../../api/todos';
+import { Filter } from '../../types/Filter';
+import { UserWarning } from '../../UserWarning';
+import { TodosList } from '../TodosList';
+import { TodosFilter } from '../TododsFilter';
+import { TempTodoContext } from '../../TempTodoContext';
+
+const USER_ID = 11903;
+const ERROR_DELAY = 3000;
+
+export const TodoApp: React.FC = () => {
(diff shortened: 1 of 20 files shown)


etojeDenys (changes requested):
image
let's center filter sections

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,189 @@
...
+        setIsHidden(true);
+      }, ERROR_DELAY);
+    } else {
+      todoService.createTodo({
+        userId: USER_ID,
+        title: todoTitle.trim(),
+        completed: false,
+      })
+        .then(newTodo => {
+          const currentTodos = [...todos];
```
TarasHoliuk, line 73: no need to copy it here
also, it's better to update the state based on the current in this way: https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

FYI

src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,189 @@
...
+        return todo.completed;
+
+      default:
+        return todo;
+    }
+  });
+
+  const completedTodo = todos.some(todo => todo.completed);
+
+  const handleOnBlur = () => {
```
TarasHoliuk, line 57: you don't run this function on blur, so name is wrong - please, fix it in the next task

src/components/TodoLoader/TodoLoader.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useContext } from 'react';
+import { TempTodoContext } from '../../TempTodoContext';
+
+export const TodoLoader: React.FC = () => {
```
TarasHoliuk, line 4: that's implemented fine, but note that in [the working example](https://mate-academy.github.io/react_todo-app-with-api/) we have `loader` on every action - deleting, adding (editing in future)

So in the next task try to do it in the same way

TarasHoliuk (approved):
The previous comment is fixed and generally everything looks good enough - approved

But I left a few comments FYI - please, take them into account in the next task

PR #908, opened 2023-11-13
solved the task

mifyy:
[DEMO LINK](https://mifyy.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoHeader.tsx
+++ b/src/components/TodoHeader.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { Todo } from '../types/Todo';
 
 type Props = {
@@ -19,6 +19,7 @@ export const Header: React.FC<Props> = ({
   updatingTodo,
 }) => {
   const inputRef = useRef<HTMLInputElement | null>(null);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
 
   useEffect(() => {
     if (inputRef.current && !updatingTodo) {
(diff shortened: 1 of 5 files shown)


src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }, [updatingTodo]);
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setNewTodo(e.target.value);
+  };
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+    if (newTodo.trim().length > 0) {
```
superpooperxxx, line 35: I would recommend not checking the `valid` case, but instead check the `invalid` and if it is invalid you do what you need and then return

```
if (newTodo.trim().length === 0) {
  setErrorNotification('Title should not be empty');
  setTimeout(() => {
    setErrorNotification('');
  }, 3000);
  
  return;
}

The rest of the code which handle valid case
```

superpooperxxx, line 35: Let's add empty lines to separate code by logic

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,88 @@
...
+          {
+            id: new Date().getTime(),
+            title: newTodo,
+            completed: false,
+            userId: 0,
+          },
+        ]);
+        setNewTodo('');
+        setUpdatingTodo(false);
+      }, 300);
```
superpooperxxx, line 49: I cannot see where you implement this
image

src/components/TodosList.tsx
```diff
@@ -0,0 +1,105 @@
...
+  };
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {filteredTodos().map((todo) => (
+        <div
+          key={todo.id}
+          data-cy="Todo"
+          className={`todo ${todo.completed ? 'completed' : ''}`}
+        >
```
superpooperxxx, line 53: I would suggest moving it to separate `TodoItem` component

src/components/TodosList.tsx
```diff
@@ -0,0 +1,105 @@
...
+          <span data-cy="TodoTitle" className="todo__title">
+            {newTodo}
+          </span>
+          {/* Remove button appears only on hover */}
+          <div data-cy="TodoLoader" className="modal overlay is-active">
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
+      )}
```
superpooperxxx, line 101: Not sure why you need this

src/components/TodosList.tsx
```diff
@@ -0,0 +1,105 @@
...
+            onClick={() => handleDeleteCurrentTodo(todo.id)}
+          >
+            ×
+          </button>
+          {deletingTodoId === todo.id && (
+            <div data-cy="TodoLoader" className="modal overlay is-active">
+              <div className="modal-background has-background-white-ter" />
+              <div className="loader" />
+            </div>
+          )}
```
superpooperxxx, line 79: You should not render the whole loader conditionally, but add `is-active` class

src/components/TodosList.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const handleDeleteCurrentTodo = (todoId: number) => {
+    setDeletingTodoId(todoId);
+
+    setTimeout(() => {
+      const updatedTodos = currentTodos.filter((todo) => todo.id !== todoId);
+
+      setCurrentTodos(updatedTodos);
+      setDeletingTodoId(null);
+    }, 300);
```
superpooperxxx, line 43: You need to send a request

image

superpooperxxx (changes requested):
Hey, you haven't fully implemented the task requirements, cannot approve😢

src/components/TodosList.tsx
mifyy: My code as it is now won't work properly without it. The item that is being loaded simply won't appear. I will make all the other corrections and try to eliminate this part of the code.

superpooperxxx (approved):
You need to send requests, I expect this in next iteration

PR #910, opened 2023-11-13
add task solution

andrew17752985:
https://andrew17752985.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/TodoItem.tsx
+++ b/src/TodoItem.tsx
@@ -1,0 +1,192 @@
+import {
+  Dispatch,
+  SetStateAction,
+  useEffect, useRef, useState,
+} from 'react';
+import classNames from 'classnames';
+import { Todo } from './types/Todo';
+import { deleteTodos } from './api/todos';
+
+type Props = {
+  todo: Todo,
+  todos: Todo[],
+  setTodos: Dispatch<SetStateAction<Todo[]>>,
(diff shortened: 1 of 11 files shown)


src/TodoList.tsx
```diff
@@ -0,0 +1,53 @@
...
+  todos: Todo[],
+  status: Status,
+  setTodos: Dispatch<SetStateAction<Todo[]>>,
+  setErrorMessage: Dispatch<SetStateAction<string>>,
+};
+
+export const TodoList: React.FC<Props> = ({
+  todos, status, setTodos, setErrorMessage,
+}) => {
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
```
etojeDenys, line 18: we don't need the separate state for visible todos, let's just filter them and return in the variable filtered array of todos

src/TodoItem.tsx
```diff
@@ -0,0 +1,192 @@
...
+        <button
+          type="button"
+          className="todo__remove"
+          data-cy="TodoDelete"
+          onClick={() => removeTodo(todo.id)}
+          aria-label="Delete"
+        >
+          ×
+        </button>
+      )}
```
etojeDenys, line 160: looks like you can combine it in one condition

etojeDenys (changes requested):

PR #916, opened 2023-11-16
Develop

DenysShurubura:


superpooperxxx (changes requested):
Don't forget about demo link

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,94 @@
...
+    await onTodoAdd(title.trim())
+      .then(() => setTitle(''))
+      .catch(() => {
+        onError('Unable to add a todo');
+      })
+      .finally(() => {
+        setIsSubmiting(false);
+        setTimeout(() => {
+          inputRef.current?.focus();
+        }, 0);
```
superpooperxxx, line 50: We usually want to handle side effects is useEffect hook

superpooperxxx (approved):
Looks good to me, just clean up comments and maybe think of handling side effects inside useEffect

PR #919, opened 2023-11-17
Initial commit

yurii-shkrobut-m:


src/components/TodoItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+        className="todo__remove"
+        data-cy="TodoDelete"
+        onClick={() => onDelete(id)}
+      >
+        ×
+      </button>
+
+      <div
+        data-cy="TodoLoader"
+        className={`modal overlay ${id === 0 && 'is-active'}`}
```
etojeDenys, line 47: use classnames for cases, to avoid false class in classes

etojeDenys (changes requested):

anastasiiavorobiova (changes requested):
Could you please provide a demo link in the PR description?

VitaliyBondarenko1982 (approved):
During on add or delete process should be visible loader on todo which in process (or on temp todo when todo is creating.)
Please fix it in next task (when will be finishing todo app).

PR #920, opened 2023-11-18
Develop

ydem1:
[DEMO LINK](https://ydem1.github.io/react_todo-app-add-and-delete/)


Esceype (changes requested):
GJ!
Still need to do:
- text field should be focused by default;
- after edit todo and make it empty, I see the placeholder but even if I save empty that save previous value;
- after edit message and add just spaces at start and end it shows it right but if enable edit mode again I see this spaces, but it shouldn`t be there

anastasiiavorobiova (approved):
Great work!

PR #924, opened 2023-11-20
Develop

kseniia-chepur:
[DEMO LINK](https://kseniia-chepur.github.io/react_todo-app-add-and-delete/) 

1 test failed ( 'should keep an error message for 3s after the last fail') -> has discussed this error with several mentors in the  chat, followed all their advices, however, the test is still failing. As a result, the mentor suggested to create the PR

### DIFF
--- a/src/components/TodoFooter.tsx
+++ b/src/components/TodoFooter.tsx
@@ -62,8 +62,6 @@ export const TodoFooter: React.FC = () => {
         </a>
       </nav>
 
-      {/* don't show this button if there are no completed todos */}
-
       <button
         type="button"
         className="todoapp__clear-completed"


sTorba24 (changes requested):
Well done!

on adding new todo i see duplicate, in 3 sec it disappears, fix please 
image

kseniia-chepur:
fixed

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return <></>;
+  }
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {activeTodoCountMsg}
+      </span>
+
+      {/* Active filter should have a 'selected' class */}
```
loralevitska, line 30: Remove comment. Try it write obvious code that explains all we need without comments)

loralevitska, line 30: Remove comment. Try it write obvious code that explains all we need without comments) Fix all cases

loralevitska (approved):
Good job!

PR #935, opened 2023-11-28
react_todo-app-and-delete

Medaed:
[DEMO LINK](https://Medaed.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoHeader.tsx
+++ b/src/components/TodoHeader.tsx
@@ -26,25 +26,29 @@ const TodoHeader: React.FC = () => {
   const handlePressEnter = async (
     event: React.KeyboardEvent<HTMLInputElement>,
   ) => {
-    if (event.key === 'Enter') {
-      const trimmedValue = inputValue.trim();
+    if (event.key !== 'Enter') {
+      return;
+    }
 
-      if (!trimmedValue) {
-        setErrorMessage('Title should not be empty');
-      } else {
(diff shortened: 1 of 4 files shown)


src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }, []);
+
+  return (
+    <header className="todoapp__header">
+      <form onSubmit={(event) => event.preventDefault()}>
+        <button
+          type="button"
+          className={
+            cn('todoapp__toggle-all',
+              { active: todos.every(todo => todo.completed) })
```
VitaliyBodnar, line 73: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,143 @@
...
+            value={editValue}
+            onChange={handleChangeEditValue}
+            onBlur={handleBlurEdit}
+            ref={inputRef}
+            onKeyPress={handleKeyPressEdit}
+          />
+        </form>
+      )}
+
+      { /* Loader */}
```
VitaliyBodnar, line 134: remove all comments

VitaliyBodnar (changes requested):

lerastarynets (approved):
Great job)

PR #949, opened 2023-12-06
add and delete solution

vadymboichenkoAcupower:
[DEMO LINK](https://vadymboichenkoAcupower.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,232 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
+import { Status } from './types/Status';
+import { Errors } from './types/Errors';
+import { TodosFilter } from './components/TodoFilter/TodoFilter';
 
-const USER_ID = 0;
+const USER_ID = 7023;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredType, setFilteredType] = useState(Status.All);
+  const [todo, setTodo] = useState<Todo | null>(null);
+  const [isInputDisabled, setIsInputDisabled] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [title, setTitle] = useState<string>('');
+
+  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,24 +1,232 @@
...
+    }
+    catch { () => {
+        setErrorMessage(Errors.CAN_NOT_DELETE_TODO);
+      }
+    }
+  }
+
+  useEffect(() => {
+    console.log(todos);
+  }, [todos])
```
etojeDenys, line 115: looks redundant

src/App.tsx
```diff
@@ -1,24 +1,232 @@
...
+      .catch(() => setErrorMessage(Errors.CAN_NOT_DELETE_TODO))
+      .finally(() => {
+        setTodos(current => current.filter(item => item.id !== todo.id));
+      });
+  }
+
+
+  const handleDeleteCompleted = async () => {
+    const deletedTodo = todos.filter(todo => todo.completed);
+    const needToDelete = deletedTodo.map(todo => {
```
etojeDenys, line 99: fix all these linter errors
image

etojeDenys (changes requested):
1. let's focus the input after adding new todo

src/App.tsx
```diff
@@ -1,24 +1,234 @@
...
+    const newTodo: Todo = {
+      title: title.trim(),
+      userId: USER_ID,
+      id: +Date,
+      completed: false,
+    };
+
+    setTodo(newTodo);
+
+    addTodo({ title: title.trim(), userId: USER_ID, completed: false })
```
TarasHoliuk, line 67: you shouldn't send the request if trimmed title is an empty string (anyway you won't add it)

src/App.tsx
```diff
@@ -1,24 +1,234 @@
...
+
+      setTodos(newTodos);
+    } catch {
+      () => {
+        setErrorMessage(Errors.CAN_NOT_DELETE_TODO);
+      };
+    }
+  };
+
+  const filterTodos = useMemo(() => {
```
TarasHoliuk, line 119:
```suggestion
  const filteredTodos = useMemo(() => {
```
`filterTodos` with verb in the name sounds like a function name

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -1,24 +1,237 @@
...
+          return !todo.completed;
+        default:
+          return true;
+      }
+    });
+  }, [filteredType, todos]);
+
+  useEffect(() => {
+    setErrorMessage(null);
+    getTodos(USER_ID).then(todo => setTodos(todo)).catch(() => {
```
etojeDenys, line 139:
```suggestion
    getTodos(USER_ID).then(setTodos).catch(() => {
```
we can simplify it

etojeDenys (changes requested):
1. image
center this section


2. still not fixed
image
 feel free to ask me in the chat, if you have some problems with that

Esceype (approved):
Well done!

PR #951, opened 2023-12-07
result

errisi:
[DEMO LINK](https://errisi.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/api/todos.tsx
+++ b/src/api/todos.tsx
@@ -1,0 +1,27 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const createTodo = ({
+  title,
+  completed,
+  userId,
+}: Omit<Todo, 'id'>) => {
+  return client.post<Todo>('/todos/', { title, userId, completed });
+};
+
+export const updateTodo = ({
+  id,
+  title,
+  completed,
+  userId,
+}: Todo) => {
+  return client.patch<Todo>(`todos/${id}`, { title, completed, userId });
+};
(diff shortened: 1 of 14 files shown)


ArtemSydun (changes requested):
not implemented
Remove all the completed todos after the Clear completed button click:

the button should be enabled only if there is at least 1 completed todo;
the deletion should work as several individual deletions running at the same time;
in case of any error show error message but process success deletions;

src/api/todos.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const createTodo = ({
+  title,
+  completed,
+  userId,
+}: Omit<Todo, 'id'>) => {
+  return client.post<Todo>('/todos/', { title, userId, completed });
+};
```
BudnikOleksii, line 18:
```suggestion
export const createTodo = (newTodo: Omit<Todo, 'id'>) => {
  return client.post<Todo>('/todos/', newTodo);
};
```
In this case I think you don't need destructing

src/components/TodoMain/TodoItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+    </div>
+  );
+};
+
+// I will need this in the next tasks
+
+// <section className="todoapp__main" data-cy="TodoList">
+// {/* This todo is being edited */}
+// <div data-cy="Todo" className="todo">
+//   <label className="todo__status-label">
```
BudnikOleksii, line 70: Get rid of comments, if you need them in the next task at least don't push it

src/components/TodoError/TodoError.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useEffect, useContext } from 'react';
+import cn from 'classnames';
+
+import { TodosContext } from '../TodosContext';
+
+export const TodoError: React.FC = () => {
+  const { todoError, setTodoError } = useContext(TodosContext);
```
BudnikOleksii, line 7: You can create a custom hook for your context and avoid passing every time the same context into `useContext`
```javascript
const useTodos = () => useContext(TodosContext);
```

PR #958, opened 2023-12-13
Task solution

Shostyy:
[DEMO LINK](https://shostyy.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/ErrorMessage/Footer/Footer.tsx
+++ b/src/components/ErrorMessage/Footer/Footer.tsx
@@ -1,0 +1,82 @@
+import React from 'react';
+import cn from 'classnames';
+
+import { TodoListState } from '../../../types/TodoListState';
+
+interface FooterProps {
+  activeTodos: number;
+  completedTodos: number;
+  filter: string;
+  setFilter: (newFilter: TodoListState) => void;
+  clearCompleted: () => void;
+}
+
+export const Footer: React.FC<FooterProps> = ({
+  activeTodos,
+  completedTodos,
+  filter,
+  setFilter,
+  clearCompleted,
+}) => {
+  const handleFilterClick = (
+    event: React.MouseEvent<HTMLAnchorElement>,
+    filterType: TodoListState,
+  ) => {
+    event.preventDefault();
+    setFilter(filterType);
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodos} items left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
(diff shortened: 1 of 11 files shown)


src/types/Errors.ts
```diff
@@ -0,0 +1,8 @@
+export enum Errors {
+  Load = 'Unable to load todos',
+  Empty = 'Title should not be empty',
+  Add = 'Unable to add a todo',
+  Delete = 'Unable to delete a todo',
+  Update = 'Unable to update a todo',
+  NoError = '',
```
l4st1m0za, line 7: Avoid using empty strings for error handling. It is better to use null to show that there is no error. An empty string may cause confusion or unexpected behavior in some cases.

src/components/ErrorMessage/Footer/Footer.tsx
```diff
@@ -0,0 +1,82 @@
...
+    filterType: TodoListState,
+  ) => {
+    event.preventDefault();
+    setFilter(filterType);
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodos} items left`}
```
l4st1m0za, line 32: Consider using a more descriptive message when there is only one item left for better user experience.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+}) => {
+  const [loading, setLoading] = useState(false);
+
+  const handleDeleteTodo = (todoId: number) => {
+    setLoading(true);
+    deleteTodo(todoId)
+      .then(() => {
+        setLoading(false);
+        filterTodoList(todoId);
+      });
```
l4st1m0za, line 25: There's no error handling for the promise returned by deleteTodo function. If the promise fails, the loading state will not change and the user interface will not respond appropriately.

src/components/ErrorMessage/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <button type="button" className="todo__remove" data-cy="TodoDelete">
+            ×
+          </button>
+
+          <div data-cy="TodoLoader" className="modal overlay is-active">
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
+      )}
```
l4st1m0za, line 49: You don't need all this logic. Instead you can use TodoItem but with tempTodo props.

l4st1m0za (changes requested):
Great job, you are very close!

lerastarynets (approved):
You've got this, keep pushing forward!

PR #960, opened 2023-12-14
Develop

MrHunterBoi:
[DEMO](https://mrhunterboi.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -36,13 +36,7 @@ export const App: React.FC = () => {
       .catch(() => errorFound(ErrorType.TodosNotLoaded));
   }, []);
 
-  const createNewTodo = (title: string) => {
-    if (!title) {
-      errorFound(ErrorType.TitleEmpty);
-
-      return;
-    }
-
+  const createNewTodo = (title: string): Promise<void> => {
     const newTodo: NoIdTodo = {
       userId: USER_ID,
       title,
@@ -54,13 +48,17 @@ export const App: React.FC = () => {
       ...newTodo,
     });
 
-    postTodo(newTodo)
+    return postTodo(newTodo)
       .then(
         (response) => setTodosFromServer(
           currentTodos => [...currentTodos, response],
         ),
       )
-      .catch(() => errorFound(ErrorType.NotAddable))
+      .catch(() => {
+        errorFound(ErrorType.NotAddable);
+
+        throw new Error(ErrorType.NotAddable);
+      })
       .finally(() => setTempTodo(null));
   };
 
@@ -120,6 +118,7 @@ export const App: React.FC = () => {
         <Header
           todosFromServer={todosFromServer}
           createNewTodo={createNewTodo}
+          errorFound={errorFound}
         />
 
         <TodoList
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,144 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 
-const USER_ID = 0;
+import { UserWarning } from './UserWarning';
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,24 +1,143 @@
...
+      )
+      .catch(() => {
+        errorFound(ErrorType.NotAddable);
+
+        throw new Error(ErrorType.NotAddable);
+      })
+      .finally(() => setTempTodo(null));
+  };
+
+  const removeTodo = (todoToDelete: Todo) => {
```
ruslanagnatenko, line 65: Loader should be visible also when todo is being deleted, i would recommend create a state for processing todo ids and add todo id every time you delete or add it so loader will be visible on that actions

ruslanagnatenko (changes requested):
Good, but need to add some changes

MrHunterBoi:
Ah, oops, forgot to add it lol

src/App.tsx
```diff
@@ -1,24 +1,130 @@
...
+    });
+  };
+
+  const filteredTodos = useCallback(
+    filterTodos,
+    [selectedFilter],
+  );
+
+  const todosToView = useMemo(
+    () => filteredTodos(todosFromServer, selectedFilter),
```
BudnikOleksii, line 92:
```suggestion
  const todosToView = useMemo(
    () => filterTodos(todosFromServer, selectedFilter),
```
`filteredTodos` looks redundant, as it does nothing, just wrap `filterTodos` function

src/api/todos.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const postTodo = (todo: NoIdTodo) => {
+  return client.post<Todo>('/todos', todo);
+};
+
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+// Add more methods here
```
BudnikOleksii, line 16: Remove comments

src/components/Errorinfo/ErrorInfo.tsx
```diff
@@ -0,0 +1,27 @@
...
+  errorMsg: ErrorType | null;
+  setErrorMsg: (error: ErrorType | null) => void;
+}
+
+export const ErrorInfo = ({ errorMsg, setErrorMsg }: Props) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={cn('notification is-danger is-light has-text-weight-normal', {
+        hidden: errorMsg === null,
```
BudnikOleksii, line 14:
```suggestion
        hidden: !errorMsg,
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,70 @@
...
+    if (!title) {
+      errorFound(ErrorType.TitleEmpty);
+
+      return;
+    }
+
+    setTodoTitleDisabled(true);
+    createNewTodo(title)
+      .then(() => setTodoTitle(''))
+      .catch(() => {})
```
BudnikOleksii, line 39: You should set an error message and show to user it, so they could know that something wrong

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const TodoItem = ({ todo, removeTodo, isLoading }: Props) => {
+  const [loaderStatus, setLoaderStatus] = useState(isLoading);
+
+  const handleDeleteTodo = () => {
+    setLoaderStatus(true);
+    removeTodo(todo)
+      .then(() => {
+        setLoaderStatus(false);
+      });
```
BudnikOleksii, line 19: You need to remove loader in any case, even if error occures

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

src/components/Header/Header.tsx
MrHunterBoi: It's already implemented inside createNewTodo

MrHunterBoi:
Refactored

BudnikOleksii (approved):
Good work 🚀

PR #964, opened 2023-12-18
Develop

Anton-Kuchmasov:
[DEMO LINK]
(https://Anton-Kuchmasov.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,169 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useMemo, useState } from 'react';
+import { UserWarning } from './components/UserWarning/UserWarning';
+import { client } from './utils/fetchClient';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { Header } from './components/Header/Header';
+import { Footer } from './components/Footer/Footer';
+import { Errors } from './components/Errors/Errors';
+import { Status } from './types/Status';
+import { ErrorSpec } from './types/ErrorSpec';
+import { filterTodos } from './helpers/filterTodos';
+import * as TodoService from './services/todo';
 
-const USER_ID = 0;
+const USER_ID = 12021;
+const ADDED_URL = `/todos?userId=${USER_ID}`;
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [hasErrors, setHasErrors] = useState(false);
+  const [error, setError] = useState<ErrorSpec | null>(null);
+  const [status, setStatus] = useState<Status>(Status.ALL);
+  const [uncompletedTodosCount, setUncompletedTodosCount] = useState<number>(0);
+  const [inputDisabled, setInputDisabled] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState<Todo | null>(null);
+
+  const filteredTodos = useMemo(
+    () => filterTodos(todosFromServer, status),
+    [status, todosFromServer],
+  );
+
+  useEffect(() => {
+    client
+      .get<Todo[]>(ADDED_URL)
+      .then((todos) => {
+        setTodosFromServer(todos);
+
+        if (todos.length === 0) {
+          setError(ErrorSpec.EMPTY_TITLE);
+        }
+      })
+      .catch(() => setError(ErrorSpec.NOT_LOADED));
+  }, []);
+
+  useEffect(() => {
+    const completedCount = todosFromServer.filter(
+      (todo) => !todo.completed,
+    ).length;
+
+    setUncompletedTodosCount(completedCount);
+  }, [todosFromServer, status, filteredTodos]);
+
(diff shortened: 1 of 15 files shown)


yefimov77 (approved):
Good Job ! 👍

src/App.tsx
```diff
@@ -1,24 +1,169 @@
...
+const USER_ID = 12021;
+const ADDED_URL = `/todos?userId=${USER_ID}`;
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [hasErrors, setHasErrors] = useState(false);
+  const [error, setError] = useState<ErrorSpec | null>(null);
+  const [status, setStatus] = useState<Status>(Status.ALL);
+  const [uncompletedTodosCount, setUncompletedTodosCount] = useState<number>(0);
+  const [inputDisabled, setInputDisabled] = useState(false);
```
roman-mirzoian, line 23:
```suggestion
  const [isInputDisabled, setIsInputDisabled] = useState(false);
```

src/App.tsx
```diff
@@ -1,24 +1,169 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState<Todo | null>(null);
```
roman-mirzoian, line 25: A state with that name should store only a boolean value for condition checking
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```

src/App.tsx
```diff
@@ -1,24 +1,169 @@
...
+
+    TodoService.createTodo(newTodo)
+      .then((responce) => {
+        setTempTodo(responce);
+        setTodosFromServer([...todosFromServer, responce]);
+      })
+      .catch(() => {
+        setError(ErrorSpec.NOT_ADDED);
+        setInputDisabled(false);
+        throw new Error();
```
roman-mirzoian, line 88: You already set the error text above, if you also want to throw the error into the console - you need to pass the `error` object from `catch` or duplicate the text, leaving the `Error` blank is definitely not an option.
Fix it in all code.

src/App.tsx
```diff
@@ -1,24 +1,169 @@
...
+        )}
+        {todosFromServer.length > 0 && (
+          <Footer
+            onStatus={handleStatus}
+            status={status}
+            completedCount={uncompletedTodosCount}
+            handleClear={clearCompleted}
+            isClearNeeded={uncompletedTodosCount === filteredTodos.length}
+          />
+        )}
```
roman-mirzoian, line 163: The render condition is the same, so there's no need to split it into separate blocks

src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,56 @@
+import { ErrorSpec } from '../../types/ErrorSpec';
+
+type Props = {
+  error: ErrorSpec | null;
+};
+
+export const ErrorMessage: React.FC<Props> = ({ error }) => {
+  const renderErrorMessage = () => {
```
roman-mirzoian, line 8: If you want to keep it as a separate function - it is better to create it outside the component.
Also, all error tags are the same, so you only need to select the text and wrap it once in the correct tags

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+          <button type="button" className="todo__remove" data-cy="TodoDelete">
+            ×
+          </button>
+
+          <div data-cy="TodoLoader" className="modal overlay is-active">
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
```
roman-mirzoian, line 88: This part is very similar to the part above that is used to render normal `todos`, so it's worth bringing it out as a separate `TodoItem` component and reusing it.

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/App.tsx
```diff
@@ -1,24 +1,168 @@
...
+            <Footer
+              onStatus={handleStatus}
+              status={status}
+              completedCount={uncompletedTodosCount}
+              handleClear={clearCompleted}
+              isClearNeeded={uncompletedTodosCount === filteredTodos.length}
+            />
+
+          </>
+        )}
```
roman-mirzoian, line 162:
```suggestion
        {todosFromServer.length > 0 && (
          <TodoList
            todos={filteredTodos}
            tempTodo={tempTodo}
            deleteTodo={removeTodo}
            isProcessing={loadingTodo}
          />
          <Footer
            onStatus={handleStatus}
            status={status}
            completedCount={uncompletedTodosCount}
            handleClear={clearCompleted}
            isClearNeeded={uncompletedTodosCount === filteredTodos.length}
          />
        )}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,39 @@
...
+          deleteTodo={deleteTodo}
+          isProcessing={isProcessing}
+        />
+      ))}
+
+      {!!tempTodo && (
+        <TodoItem
+          key={tempTodo.id}
+          todo={tempTodo}
+          deleteTodo={() => {}}
```
roman-mirzoian, line 33: You can make it an optional props, then you won't need to pass an empty function

roman-mirzoian (approved):
Well done 👍

PR #966, opened 2023-12-19
implement solution

yefimov77:
[DEMO LINK](https://yefimov77.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,0 +1,31 @@
+import { Todo as TodoType } from '../../types/Todo';
+import { Todo } from '../Todo/Todo';
+
+interface Props {
+  todos: TodoType[],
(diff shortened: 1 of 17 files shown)


Yevhen-Kolosov (approved):
Good job 👍

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,31 @@
...
+          todo={todo}
+          onDelete={onDelete}
+        />
+      ))}
+
+      {tempTodo && (
+        <Todo
+          key={tempTodo.id}
+          todo={tempTodo}
+          onDelete={() => {}}
```
roman-mirzoian, line 26: You can make it an optional props, then you won't need to pass an empty function

roman-mirzoian (approved):
Well done 👍

PR #970, opened 2023-12-19
Develop

OstinM:
- [DEMO LINK](https://OstinM.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,6 +21,8 @@ export const App: React.FC = () => {
   const [filterBy, setFilterBy] = useState<FilteredBy>(FilteredBy.DefaultType);
   const [errorType, setErorType] = useState<ErrorType | null>(null);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
 
   const todoToRemove = async (todoId: number) => {
     try {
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,24 +1,140 @@
...
+        completed: false,
+        userId: USER_ID,
+      });
+
+      if (addedTodo.id !== 0) {
+        setTempTodo({
+          ...newTempTodo,
+          id: addedTodo.id,
+        });
+      }
```
BudnikOleksii, line 60: Looks like this step is redundant, as when you have a real todo you just set it  to todos list

src/App.tsx
```diff
@@ -1,24 +1,140 @@
...
+          />
+        )}
+        {todos.length
+          ? (
+            <Footer
+              onFilter={setFilterBy}
+              todos={todosToView}
+              filterBy={filterBy}
+            />
+          ) : null}
```
BudnikOleksii, line 129:
```suggestion
        {todos.length > 0 && (
          <Footer
            onFilter={setFilterBy}
            todos={todosToView}
            filterBy={filterBy}
          />
        )}
```

src/helpers.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Todo } from './types/Todo';
+
+export enum FilteredBy {
+  DefaultType = '',
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
+
+type FilterType = string;
```
BudnikOleksii, line 11: You have `FilteredBy` enum, so you don't need this type, as filter shouldn't be a string

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor fixes are required. Also, check your solution, on demo I don't see the loader when deleting or adding todo

src/App.tsx
```diff
@@ -1,24 +1,147 @@
...
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [isLoadingTodo, setIsLoadingTodo] = useState<Todo | null>(null);
+
+  const todosToView = useMemo(
+    () => filteredTodoList(todos, filterBy),
+    [todos, filterBy],
+  );
+
+  const todoToRemove = async (todoId: number) => {
```
BudnikOleksii, line 33:
```suggestion
  const removeTodo = async (todoId: number) => {
```

src/App.tsx
```diff
@@ -1,24 +1,147 @@
...
+    try {
+      setIsLoadingTodo(todos.find(todo => todo.id === todoId) || null);
```
BudnikOleksii, line 35: I didn't get it, why do you need this state?
For the future as you need to handle loaders for many todos in the same time you can store array of todos ids that currently have the loading state and show loaders for all needed todos

BudnikOleksii (approved):
Good work 👍

PR #972, opened 2023-12-19
add task solution

ViktorAfk:


src/Components/Footer/FooterNav/FooterNav.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <nav className="filter" data-cy="Filter">
+      <a
+        href="#/"
+        className={cn('filter__link',
+          { selected: selectedOption === States.All })}
+        data-cy="FilterLinkAll"
+        onClick={() => setSelectedOption(States.All)}
+      >
+        All
```
polosanya, line 17: You can use Enum instead of the hardcoded value here too

src/Context/Context.tsx
```diff
@@ -0,0 +1,179 @@
...
+        setRenderedTodos(currentRenderedTodos => [...currentRenderedTodos,
+          newTodo]);
+      })
+      .catch((error) => {
+        setErrorMessage(Error.Load);
+        throw error;
+      });
+  };
+
+  const handleDelite = (todoId:number) => {
```
polosanya, line 99:
```suggestion
  const handleDelete = (todoId: number) => {
```

polosanya (approved):
Good job!

PR #973, opened 2023-12-19
task solution

VeDono:
[DEMO LINK](https://VeDono.github.io/react_todo-app-add-and-delete/)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <nav className="filter" data-cy="Filter">
+      <a
+        href="#/"
+        className={cn('filter__link', {
+          selected: filterSelected === FilterOption.All,
+        })}
+        data-cy="FilterLinkAll"
+        onClick={() => setFilterSelected(FilterOption.All)}
+      >
+        All
```
polosanya, line 19: Better use Enum instead of the hardcoded value

polosanya (approved):
Good job!

PR #977, opened 2023-12-20
implemented add and delete

emilionila:
[DEMO LINK](https://emilionila.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,149 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import * as todosService from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+// import { ErrorNotification } from './components/ErrorNotification';
+import { Errors } from './types/Errors';
 
-const USER_ID = 0;
+const USER_ID = 12037;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterValue, setFilterValue] = useState('all');
+  const [errorMessage, setErrorMessage] = useState<Errors>(Errors.NULL);
+  const [tempTodo, setTempTodo] = useState<Omit<Todo, 'id'> | null>(null);
+
+  useEffect(
+    () => {
+      todosService.getTodos(USER_ID)
+        .then(setTodos)
+        .catch(() => setErrorMessage(Errors.LOAD))
+        .finally(() => setTimeout(() => setErrorMessage(Errors.NULL), 3000));
+    }, [],
+  );
+
+  const todosToRender = useMemo(
+    () => {
+      return todos.filter(todo => {
+        return filterValue === 'all'
+          || (filterValue === 'completed' ? todo.completed : !todo.completed);
+      });
+    },
+    [todos, filterValue],
+  );
+
+  const addTodo = async (inputValue: string) => {
+    const data = {
+      id: 0,
+      title: inputValue,
+      completed: false,
+      userId: USER_ID,
+    };
+
+    setTempTodo(data);
+
+    try {
+      const createdTodo = await todosService.createTodo(data);
+
+      return (
+        setTodos(currentTodos => [...currentTodos, createdTodo]));
+    } catch (error) {
+      setErrorMessage(Errors.ADD);
+      throw error;
+    } finally {
+      setTimeout(() => setErrorMessage(Errors.NULL), 3000);
+      setTempTodo(null);
+    }
+  };
+
+  const updateTodo = async (todoToUpdate: Todo) => {
+    try {
+      const updatedTodo = await todosService
+        .updateTodo(todoToUpdate.id, todoToUpdate);
+
+      setTodos(currentTodos => {
+        const newTodos = [...currentTodos];
+
+        const findIndexTodo = [...todos].findIndex(
+          todo => todo.id === todoToUpdate.id,
+        );
+
+        newTodos.splice(findIndexTodo, 1, updatedTodo);
+
+        return newTodos;
+      });
+    } catch (error) {
+      setErrorMessage(Errors.UPDATE);
+      throw error;
+    } finally {
+      setTimeout(() => setErrorMessage(Errors.NULL), 3000);
+    }
+  };
+
+  const deleteTodo = (todoId: number) => {
+    todosService.deleteTodo(todoId)
+      .then(() => setTodos(
+        currentTodos => currentTodos.filter(todo => todo.id !== todoId),
+      ))
+      .catch(() => setErrorMessage(Errors.DELETE))
+      .finally(() => {
+        setTimeout(() => setErrorMessage(Errors.NULL), 3000);
+      });
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <Header
+          updateTodo={updateTodo}
+          todos={todos}
+          addTodo={addTodo}
+          setErrorMessage={setErrorMessage}
+        />
+
+        <TodoList
+          todos={todosToRender}
+          tempTodo={tempTodo}
+          deleteTodo={deleteTodo}
+        />
+
+        {todos.length > 0 && (
+          <Footer
+            todos={todosToRender}
+            setFilterValue={setFilterValue}
+            filterValue={filterValue}
+          />
+        )}
+      </div>
+
+      {errorMessage && (
+        <div
+          className="notification is-danger is-light has-text-weight-normal"
+        >
+          <button
+            type="button"
+            className={cn('delete', {
+              hidden: !errorMessage,
+            })}
+            onClick={() => {
+              setErrorMessage(Errors.NULL);
+            }}
+          />
+          {errorMessage}
+        </div>
+
+      )}
+    </div>
   );
 };
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,72 @@
+import { FC } from 'react';
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+import { FilterValue } from '../../types/FilterValue';
+
+interface Props {
+  todos: Todo[],
+  setFilterValue: (value: string) => void,
+  filterValue: string,
+}
+
+export const Footer: FC<Props> = (props) => {
+  const {
+    todos,
+    setFilterValue,
+    filterValue,
+  } = props;
+
+  const remainingTodosLength = todos.filter(todo => !todo.completed).length;
+  const completedTodosLength = todos.filter(todo => todo.completed).length;
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${remainingTodosLength} items left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={cn('filter__link', { selected: filterValue === 'all' })}
+          data-cy="FilterLinkAll"
+          onClick={() => setFilterValue(FilterValue.ALL)}
+        >
+          All
+        </a>
+
+        <a
+          href="#/active"
+          className={cn('filter__link', { selected: filterValue === 'active' })}
+          data-cy="FilterLinkActive"
+          onClick={() => setFilterValue(FilterValue.ACTIVE)}
+        >
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={cn('filter__link',
+            { selected: filterValue === 'completed' })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setFilterValue(FilterValue.COMPLETED)}
+        >
+          Completed
+        </a>
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed hidden"
+        data-cy="ClearCompletedButton"
+        style={{
+          visibility: completedTodosLength !== 0
+            ? 'visible' : 'hidden',
+        }}
+      >
+        Clear completed
+      </button>
+
+    </footer>
+  );
+};
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,0 +1,93 @@
+import React, {
+  FC, useCallback, useMemo, useState,
+} from 'react';
+import cn from 'classnames';
+import { Errors } from '../../types/Errors';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[],
+  setErrorMessage: (error: Errors) => void,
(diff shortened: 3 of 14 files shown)


src/App.tsx
```diff
@@ -1,24 +1,149 @@
...
+        .catch(() => setErrorMessage(Errors.LOAD))
+        .finally(() => setTimeout(() => setErrorMessage(Errors.NULL), 3000));
+    }, [],
+  );
+
+  const todosToRender = useMemo(
+    () => {
+      return todos.filter(todo => {
+        return filterValue === 'all'
+          || (filterValue === 'completed' ? todo.completed : !todo.completed);
```
MrHunterBoi, line 34: Interesting filter

MrHunterBoi (approved):
Well done

src/App.tsx
```diff
@@ -1,24 +1,149 @@
...
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import * as todosService from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+// import { ErrorNotification } from './components/ErrorNotification';
```
BudnikOleksii, line 10: Uncomment and use your component

src/App.tsx
```diff
@@ -1,24 +1,149 @@
...
+import { Errors } from './types/Errors';
 
-const USER_ID = 0;
+const USER_ID = 12037;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterValue, setFilterValue] = useState('all');
```
BudnikOleksii, line 17: Create an enum for the filter and use it instead of magic values

src/App.tsx
BudnikOleksii, line 34: Use enum instead of 'all'/'completed'

src/App.tsx
```diff
@@ -1,24 +1,149 @@
...
+    try {
+      const createdTodo = await todosService.createTodo(data);
+
+      return (
+        setTodos(currentTodos => [...currentTodos, createdTodo]));
+    } catch (error) {
+      setErrorMessage(Errors.ADD);
+      throw error;
+    } finally {
+      setTimeout(() => setErrorMessage(Errors.NULL), 3000);
```
BudnikOleksii, line 59: In every API call, you set an error and then remove it in 3 seconds. Create a function for it, for example:
```typescript
const handleError = (errorMessage: Errors) => {
  setErrorMessage(errorMessage);
  setTimeout(() => setErrorMessage(Errors.NULL), 3000);
};
```

src/App.tsx
```diff
@@ -1,24 +1,149 @@
...
+        .updateTodo(todoToUpdate.id, todoToUpdate);
+
+      setTodos(currentTodos => {
+        const newTodos = [...currentTodos];
+
+        const findIndexTodo = [...todos].findIndex(
+          todo => todo.id === todoToUpdate.id,
+        );
+
+        newTodos.splice(findIndexTodo, 1, updatedTodo);
```
BudnikOleksii, line 76: Use `map` instead of findIndex + splice

src/App.tsx
```diff
@@ -1,24 +1,149 @@
...
+      setErrorMessage(Errors.UPDATE);
+      throw error;
+    } finally {
+      setTimeout(() => setErrorMessage(Errors.NULL), 3000);
+    }
+  };
+
+  const deleteTodo = (todoId: number) => {
+    todosService.deleteTodo(todoId)
+      .then(() => setTodos(
```
BudnikOleksii, line 90: It's better not to combine async/await and then/catch at the same time. Choose and use one style to handle async operations

src/App.tsx
```diff
@@ -1,24 +1,149 @@
...
+            todos={todosToRender}
+            setFilterValue={setFilterValue}
+            filterValue={filterValue}
+          />
+        )}
+      </div>
+
+      {errorMessage && (
+        <div
+          className="notification is-danger is-light has-text-weight-normal"
```
BudnikOleksii, line 132: Use `ErrorNotification` component instead

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${remainingTodosLength} items left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={cn('filter__link', { selected: filterValue === 'all' })}
+          data-cy="FilterLinkAll"
+          onClick={() => setFilterValue(FilterValue.ALL)}
```
BudnikOleksii, line 33: So, you have the `FilterValue` enum, use it, replace all magic values with it

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    setErrorMessage(Errors.NULL);
+    setIsDisabledInput(true);
+
+    addTodo(inputValue)
+      .then(() => {
+        setIsDisabledInput(false);
+        setInputValue('');
+      })
+      .catch(() => setIsDisabledInput(false));
```
BudnikOleksii, line 46: You do `setIsDisabledInput` in both cases, so do it in finally

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,78 @@
...
+            type="button"
+            className="todo__remove"
+            data-cy="TodoDelete"
+            onClick={() => deleteTodo(todo.id)}
+          >
+            ×
+          </button>
+
+          {/* overlay will cover the todo while it is being updated */}
+          <div data-cy="TodoLoader" className="modal overlay">
```
BudnikOleksii, line 48: Remove comment and show loader when user delete or toggle all todos, so user can understand that they should wait

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <span className="todo__title">
+            {tempTodo.title}
+          </span>
+
+          <div className="modal overlay is-active">
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
+      )}
```
BudnikOleksii, line 75: Remove it, to show the `tempTodo` use the same component

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,78 @@
...
+    todos,
+    tempTodo,
+    deleteTodo,
+  } = props;
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(todo => (
+        <div
+          data-cy="Todo"
```
BudnikOleksii, line 22: Create `Todo` component

src/types/Errors.ts
```diff
@@ -0,0 +1,8 @@
+export enum Errors {
+  NO_TITLE = 'Title can\'t be empty',
+  ADD = 'Unable to add a todo',
+  DELETE = 'Unable to delete a todo',
+  UPDATE = 'Unable to update a todo',
+  LOAD = 'Unable to load todos',
+  NULL = '',
+}
```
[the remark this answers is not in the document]
BudnikOleksii: > `enum` keys should be capitalized, not be in uppercased

Make sense

BudnikOleksii (changes requested):
Great 👍 
But could be even better, check the suggestions. And show the loader when sending requests to API so a user can understand that something happens and they should wait

src/App.tsx
```diff
@@ -1,24 +1,140 @@
...
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import * as todosService from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { Errors } from './types/Errors';
+import { FilterValue } from './types/FilterValue';
+import { ErrorNotification } from './components/ErrorNotification';
+// import {handleError, setErrorMessage} from "./helpers";
```
BudnikOleksii, line 11: Remove comments

src/App.tsx
```diff
@@ -1,24 +1,140 @@
...
+  const [tempTodo, setTempTodo] = useState<Omit<Todo, 'id'> | null>(null);
+  const [errorMessage, setErrorMessage] = useState<Errors>(Errors.Null);
+  const [showLoader, setShowLoader] = useState(false);
+
+  useEffect(
+    () => {
+      todosService.getTodos(USER_ID)
+        .then(setTodos)
+        .catch(() => setErrorMessage(Errors.LoadTodos))
+        .finally(() => setTimeout(() => setErrorMessage(Errors.Null), 3000));
```
BudnikOleksii, line 27: - in the positive case you don't have an error, so you don't need to set timeout.
- you do the same when you call other APIs, let's create a simple handler to do it and then you can just pass the error message in all catch blocks
```typescript
const handleError = (errorMessage: string) => {
  setErrorMessage(errorMessage);
  setTimeout(() => setErrorMessage(Errors.Null), 3000)
};
```

src/App.tsx
```diff
@@ -1,24 +1,140 @@
...
+    setTempTodo(data);
+
+    try {
+      const createdTodo = await todosService.createTodo(data);
+
+      return (
+        setTodos(currentTodos => [...currentTodos, createdTodo]));
+    } catch (error) {
+      setErrorMessage(Errors.AddTodo);
+      throw error;
```
BudnikOleksii, line 59: You don't need to throw an error as it's already set and shown to user

src/App.tsx
BudnikOleksii: > It's better not to combine async/await and then/catch at the same time. Choose and use one style to handle async operations

Not fixed

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const handleFocusOnInput = useCallback((input: HTMLInputElement) => {
+    input?.focus();
+  }, []);
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!inputValue.trim()) {
+      setErrorMessage(Errors.EmptyTitle);
+      setTimeout(() => setErrorMessage(Errors.Null), 3000);
```
BudnikOleksii, line 33: When you created handler for errors you can use it here instead too, not only for API calls

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    addTodo(inputValue)
+      .then(() => {
+        setInputValue('');
+      })
+      .finally(() => setIsDisabledInput(false));
+  };
+
+  const completeAllTodos = async () => {
+    const statusOfTodos = [...todos].every(todo => todo.completed);
```
BudnikOleksii, line 49:
```suggestion
    const statusOfTodos = todos.every(todo => todo.completed);
```
`every` doesn't mutate an array so you don't need to create a copy

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const updateTodos = statusOfTodos
+      ? todos.filter(todo => todo.completed)
+      : todos.filter(todo => !todo.completed);
```
BudnikOleksii, line 53:
```suggestion
    const updateTodos = todos.filter(todo => statusOfTodos);
```
Looks like you can simplify

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,46 @@
...
+        type="button"
+        className="todo__remove"
+        data-cy="TodoDelete"
+        onClick={() => deleteTodo(todo.id)}
+      >
+        ×
+      </button>
+
+      {showLoader && (
+        <div data-cy="TodoLoader" className="modal overlay">
```
BudnikOleksii, line 39: To show the loader add the class 'is-active' to it

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,69 @@
...
+                key={todo.id}
+                timeout={300}
+                classNames="item"
+            >
+              <Todo
+          key={todo.id}
+          todo={todo}
+          deleteTodo={deleteTodo}
+          showLoader={showLoader}
+        />
```
BudnikOleksii, line 35: 👍 
But, please, fix the formatting in the `TodoList` component

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,69 @@
...
+          >
+        <div
+          className="todo"
+        >
+          <label className="todo__status-label">
+            <input
+              type="checkbox"
+              className="todo__status"
+            />
+          </label>
```
BudnikOleksii, line 53: Use the same `Todo` component

src/helpers.ts
```diff
@@ -0,0 +1,9 @@
+// import { useState } from 'react';
+// import { Errors } from './types/Errors';
+//
+// export const [errorMessage, setErrorMessage] = useState<Errors>(Errors.Null);
+//
+// export const handleError = (errorMessage: Errors) => {
+//   setErrorMessage(errorMessage);
+//   setTimeout(() => setErrorMessage(Errors.Null), 3000);
+// };
```
BudnikOleksii, line 9: 👍 
Use this function to handle errors, move it to the component

BudnikOleksii (changes requested):
Good progress 👍 
But, some fixes are still required.
Show loaders as I mentioned previously.
One more suggestion, as you need to show multiple loaders at the same time you can store in the state an array of ids of todos that currently loading. So in the todo component you can check if the current todo id is in the array, and show the loader.

src/App.tsx
```diff
@@ -1,24 +1,136 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterValue, setFilterValue] = useState(FilterValue.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [errorMessage, setErrorMessage] = useState<Errors>(Errors.Null);
+  const [loadingTodoId, setLoadingTodoId] = useState<number[]>([]);
+
+  const handleError = (error: Errors) => {
+    setErrorMessage(error);
+    setTimeout(() => setErrorMessage(error), 3000);
```
BudnikOleksii, line 23:
```suggestion
    setTimeout(() => setErrorMessage(Errors.Null), 3000);
```
In setTimeout, you should reset the error

src/App.tsx
```diff
@@ -1,24 +1,136 @@
...
+  };
+
+  useEffect(
+    () => {
+      todosService.getTodos(USER_ID)
+        .then(setTodos)
+        .catch(() => handleError(Errors.LoadTodos))
+        .finally(() => handleError(Errors.Null));
```
BudnikOleksii, line 31: Finally is redundant in this case, `handleError` should set an error and reset

src/App.tsx
```diff
@@ -1,24 +1,136 @@
...
+
+    setTempTodo(data);
+
+    try {
+      const createdTodo = await todosService.createTodo(data);
+      setTodos(currentTodos => [...currentTodos, createdTodo]);
+    } catch (e) {
+      handleError(Errors.AddTodo);
+    } finally {
+      handleError(Errors.Null);
```
BudnikOleksii, line 62: The same, `handleError(Errors.Null);` is redundant. Fix it in all places

src/App.tsx
BudnikOleksii: > It's better not to combine async/await and then/catch at the same time. Choose and use one style to handle async operations

How about this❓ 🤷‍♂️

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,52 @@
...
+          </CSSTransition>
+        ))}
+
+        {tempTodo && (
+          <CSSTransition
+            key={0}
+            timeout={300}
+            classNames="temp-item"
+          >
+            <TempTodo tempTodo={tempTodo} />
```
BudnikOleksii, line 46: You don't need to create a `TempTodo` component, use `Todo` instead.
As in this case, you have to support two same components

BudnikOleksii (changes requested):
Almost done, just a few minor fixes are required

BudnikOleksii (approved):
Looks great 🔥

PR #982, opened 2023-12-21
solution

ilyapgitt:
[DEMO LINK](https://ilyapgitt.github.io/react_todo-app-add-and-delete/)


Yevhen-Kolosov (changes requested):
Great work! 👍 The task is almost done it's just missing a few things (mostly related to the lack of `catch` blocks in the handlers that would show certain errors on different request failures) 1x

PR #983, opened 2023-12-22
Solution

YuriySobakar:
[DEMO LINK](https://YuriySobakar.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,128 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { createTodo, getTodos } from './api/todos';
+import { TodoList } from './TodoList/TodoList';
+import {
+  applyUncompleted,
+  applySelectedTodos,
+} from './helpers/helpers';
+import { Filter } from './types/Selected-filter-enum';
+import { ErrorType } from './types/errors-enum';
+import { Header } from './Header/Header';
+import { Footer } from './Footer/Footer';
+import { Notification } from './Notification/Notification';
 
-const USER_ID = 0;
+const USER_ID = 12057;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo []>([]);
+  const [error, setError] = useState<ErrorType | null>(null);
+  const [filterType, setFilterType] = useState<Filter>(Filter.all);
+  const [newTitle, setNewTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const hasTodosFromServer = todos.length !== 0;
+
+  const handleFilterType = (type: Filter) => {
+    setFilterType(type);
+  };
+
+  const handleTitle = (e:React.ChangeEvent<HTMLInputElement>) => {
+    setNewTitle(e.target.value);
+  };
+
+  const handleError = (errorMessage: ErrorType) => {
+    setError(errorMessage);
+
+    setTimeout(() => {
+      setError(null);
+    }, 3000);
+  };
+
+  useEffect(() => {
+    setError(null);
+    getTodos(USER_ID)
+      .then((todosFS => setTodos(todosFS)))
+      .catch(() => handleError(ErrorType.didNotGetTodos));
+  }, []);
+
+  const onCreateTodo = () => {
+    const trimmedTitle = newTitle.trim();
+
+    setIsLoading(true);
+    setTempTodo({
+      id: 0,
+      title: trimmedTitle,
+      completed: false,
+      userId: USER_ID,
+    });
+
+    createTodo({
+      title: trimmedTitle,
+      completed: false,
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,24 +1,147 @@
...
+  };
+
+  const onDeleteTodo = (id: number) => {
+    deleteTodo(id)
+      .then(() => {
+        setTodos((currentTodos) => removeTodo(id, currentTodos));
+      })
+      .catch(() => {
+        handleError(ErrorType.cantDeleteTodo);
+      });
```
BudnikOleksii, line 89: You should show loaders for all todos that processing now, for now, I don't see any loaders when deleting one todo or click on the clear completed button
You can create a state with todos ids that currently updating and depends on it show loaders.
Simple example:
```typescript
const [loadingTodosIds, setLoadingTodosIds] = useState<number[]>([]);
```

```suggestion
    setLoadingTodosIds(prev => [...prev, id]);
    
    deleteTodo(id)
      .then(() => {
        setTodos((currentTodos) => removeTodo(id, currentTodos));
      })
      .catch(() => {
        handleError(ErrorType.cantDeleteTodo);
      })
      .finally(() => {
        setLoadingTodosIds(prev => prev.filter(todoId => todoId !== id));
      });
```

src/App.tsx
```diff
@@ -1,24 +1,147 @@
...
+  };
+
+  const clearCompleted = () => {
+    const completedTodos = todos.filter(({ completed }) => completed);
+
+    completedTodos.map(({ id }) => {
+      return onDeleteTodo(id);
+    });
```
BudnikOleksii, line 97: If you don't use returned array there is no sense to use `map` method, you can use `forEach` instead

src/types/errors-enum.ts
```diff
@@ -0,0 +1,6 @@
+export enum ErrorType {
+  emptyInput = 'Title should not be empty', // писать с большой буквы
```
BudnikOleksii, line 2:
```suggestion
  EmptyInput = 'Title should not be empty', // писать с большой буквы
```
Yes, in enum write properties in uppercase

BudnikOleksii (changes requested):
Looks great 👍 
Just one moment has to be implemented, you should show the loaders for users so they know that something happening and they need to wait

src/App.tsx
```diff
@@ -1,24 +1,154 @@
...
+      .finally(() => {
+        setLoadingTodosIds(prev => prev.filter(todoId => todoId !== id));
+      });
+  };
+
+  const clearCompleted = () => {
+    const completedTodos = todos.filter(({ completed }) => completed);
+
+    completedTodos.forEach(({ id }) => {
+      return onDeleteTodo(id);
```
BudnikOleksii, line 102: return doesn't have any effect in forEach

src/TempTodo/TempTodo.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import cn from 'classnames';
+import { Todo } from '../types/Todo';
+
+interface Props {
+  todo: Todo,
+}
+
+export const TempTodo: React.FC<Props> = ({ todo }) => {
```
BudnikOleksii, line 9: You don't need another component for temp todo, use the same `TodoInfo`

src/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onDeleteTodo,
+  hasLoadingStatus,
+}) => {
+  const { title, completed, id } = todo;
+
+  return (
+    <div
+      data-cy="Todo"
+      className={cn('todo', {
+        completed: completed === true,
```
BudnikOleksii, line 22:
```suggestion
        completed,
```
You don't need to compare with true as the value is already boolean, and you can simplify

src/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,59 @@
...
+        value={id}
+        onClick={(e) => onDeleteTodo(+e.currentTarget.value)}
+      >
+        ×
+      </button>
+
+      <div
+        data-cy="TodoLoader"
+        className={cn('modal overlay', {
+          'is-active': hasLoadingStatus === true,
```
BudnikOleksii, line 51:
```suggestion
          'is-active': loadingTodosIds.includes(id), // or pass directly this value to todo info component
```
Now when I delete one todo I see loaders on all todos instead. Loader should be shown only for todos that are actually loading, you can check if current todo id is in the loadingTodosIds array

src/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,34 @@
...
+    {todos.map((todo) => (
+      <TodoInfo
+        todo={todo}
+        key={todo.id}
+        onDeleteTodo={onDeleteTodo}
+        hasLoadingStatus={hasLoadingStatus}
+      />
+    ))}
+
+    {tempTodo && <TempTodo todo={tempTodo} />}
```
BudnikOleksii, line 32: Use `TodoInfo` component instead

BudnikOleksii (changes requested):
Good progress, but some features should be fixed

BudnikOleksii (approved):
It should work, but looks like you forgot to deploy the latest changes.
Approved, but, please, make sure that all works as expected

PR #984, opened 2023-12-22
react_todo-app-add-and-delete

OlegRak86:
https://OlegRak86.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,70 @@
+/* eslint-disable max-len */
+import React, { useMemo } from 'react';
+import cn from 'classnames';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todos: Todo[];
+  filter:string;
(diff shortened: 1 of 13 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+        </a>
+      </nav>
+
+      <button
+        disabled={!todos.some((el) => el.completed === true)}
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        onMouseDown={() => {
+          setFilter('Clear');
```
etojeDenys, line 62: it's not the filter value, you can just delete all todos in some function and call that function here

etojeDenys (changes requested):
you can just send the delete request and then change the todos locally, without get todos request
image

lerastarynets (approved):
lgtm 🚀

PR #985, opened 2023-12-25
implement Todos deliting, Clear Button, decompose Footer

Softjey:
[DEMO LINK](https://softjey.github.io/react_todo-app-add-and-delete/)

Please be strict with my code. I want to write in the best style, but unfortunately I often don't know if my decisions are correct. 

I'm worried about TodosContext, isn't it too big? And when I press the Clear Button, I simply use the forEach method to go through all the todos and delete them. Should we use Promise.all instead?

### DIFF
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -1,0 +1,59 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todo: Todo,
+  isLoading?: boolean,
+}
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  isLoading = false,
+}) => {
+  const [isEditing] = useState(false);
+  const { title, completed } = todo;
+
+  return (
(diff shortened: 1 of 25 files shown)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,85 @@
...
+    if (deleteTodo) {
+      deleteTodo(id)
+        .finally(() => {
+          setIsDeliting(false);
+        });
+    }
+  };
+
+  return (
+    <div data-cy="Todo" className={`todo ${completed && 'completed'}`}>
```
etojeDenys, line 35: better to use classnames package for such cases to avoid it
image

etojeDenys (approved):
good job

etojeDenys:
> [DEMO LINK](https://softjey.github.io/react_todo-app-add-and-delete/)
> 
> Please be strict with my code. I want to write in the best style, but unfortunately I often don't know if my decisions are correct.
> 
> I'm worried about TodosContext, isn't it too big?
> 
> And when I press the Clear Button, I simply use the forEach method to go through all the todos and delete them. Should we use Promise.all instead?

yeah, better to use Promise.all for those requests

PR #996, opened 2024-01-02
Solution

MarcinLigmann:
[DEMO LINK](https://MarcinLigmann.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,8 +6,7 @@ import { TodoErr } from './components/todoerr/todoerr';
 import { TodoForm } from './components/todoform/todoform';
 import { TodoList } from './components/todolist/todolist';
 import { useTodos } from './context/todoProvider';
-
-export const USER_ID = 12075;
+import { USER_ID } from './utils/userID';
 
 export const App: React.FC = () => {
   const { todos } = useTodos();
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,37 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
 import { UserWarning } from './UserWarning';
+import { TodoFooter } from './components/todofooter/todofooter';
+import { TodoErr } from './components/todoerr/todoerr';
+import { TodoForm } from './components/todoform/todoform';
+import { TodoList } from './components/todolist/todolist';
+import { useTodos } from './context/todoProvider';
 
-const USER_ID = 0;
+export const USER_ID = 12075;
 
 export const App: React.FC = () => {
+  const { todos } = useTodos();
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
   return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">React Todo App - Load Todos</a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+      <div className="todoapp__content">
+        <TodoForm />
+
+        <TodoList />
+
+        {todos.length > 0
+          && <TodoFooter />}
+
+      </div>
+
+      <TodoErr />
+
+    </div>
   );
 };
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -1,0 +1,8 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+// Add more methods here
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -6,3 +6,10 @@ export const getTodos = (userId: number) => {
 };
 
 // Add more methods here
+export const AddTodo = ({ title, userId, completed }: any) => {
+  return client.post<Todo>('/todos', { title, userId, completed });
+};
+
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
--- a/src/components/todoerr/todoerr.tsx
+++ b/src/components/todoerr/todoerr.tsx
@@ -1,0 +1,45 @@
+import classNames from 'classnames';
+import { useEffect } from 'react';
+import { useTodos } from '../../context/todoProvider';
+
+export const TodoErr = () => {
+  const { error, setError } = useTodos();
+
+  useEffect(() => {
+    setTimeout(() => {
+      setError(null);
+    }, 3000);
+  }, [error, setError]);
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal', {
+          hidden: !error,
+        },
+      )}
+    >
+      <button
+        data-cy="HideErrorButton"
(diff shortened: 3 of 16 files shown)


src/components/todolist/todolist.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { ErrorType } from '../../types/Error';
+
+export const TodoList = () => {
+  const {
+    visibleTasks, setError, tempTodo, todos,
+    setTodos, deletingTask, setDeletingTask,
+  } = useTodos();
+
+  const handleDeleteClick = (id: number) => {
+    const deletingId = [...deletingTask, id];
```
kate-prakofyeva, line 14: onClick={() => handleDeleteClick(task.id)}
 why you find id again?

src/components/todolist/todolist.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+    setDeletingTask(deletingId);
```
kate-prakofyeva, line 16: if you have a queue todos for deleting, and api has only delete 1 todo, maybe better to map todos, do deleteOne(id) and await Promise.allSettled, for be confident, that we finished process
Make 2 function, one for delete 1 todo, second for delete all completed

src/components/todoerr/todoerr.tsx
```diff
@@ -0,0 +1,36 @@
+import classNames from 'classnames';
+import { useEffect } from 'react';
+import { useTodos } from '../../context/todoProvider';
+
+export const TodoErr = () => {
+  const { error, setError } = useTodos();
+
+  useEffect(() => {
+    setTimeout(() => {
```
kate-prakofyeva, line 9: add return clearInterval, for remouve setTimeout

src/components/todofooter/todofooter.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const TodoFooter = () => {
+  const {
+    filterBy, setFilterBy, countIncompleteTask, todos,
+    setTodos, setError, deletingTask, setDeletingTask,
+  } = useTodos();
+
+  const hiddenBtn = todos.filter(el => el.completed).length === 0;
+
+  const DeleteCompletedTask = () => {
+    const completedTodos = todos.filter(task => !task.completed);
```
kate-prakofyeva, line 15: also hide the notification before any next request;

in the start new request, I add setError(null)

src/api/todos.ts
```diff
@@ -0,0 +1,15 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = (userId: number) => {
+  return client.get<Todo[]>(`/todos?userId=${userId}`);
+};
+
+// Add more methods here
+export const AddTodo = ({ title, userId, completed }: any) => {
```
kate-prakofyeva, line 9: ```export function addTodo(todo: Omit<Todo, 'id'>) {```
you can used Omit type

kate-prakofyeva (approved):
I'm not sure about the process deleting todo, set erorrs and async that function

src/api/todos.ts
mbruhler, line 9: Provide interface for this type, don't type as any

src/components/todoinfo/todoinfo.tsx
```diff
@@ -0,0 +1,112 @@
+// import { useTodos } from '../../context/todoProvider';
```
mbruhler, line 1: clear the comments

src/components/todoform/todoform.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    setIsAddingTask(true);
+    setTempTodo({
+      id: 0,
+      userId: USER_ID,
+      title: taskName.trim(),
+      completed: false,
+    });
+
+    AddTodo({
```
mbruhler, line 39: we shouldn't name functions with first capital letter

src/types/Error.ts
```diff
@@ -0,0 +1,7 @@
+export enum ErrorType {
+  Load = 'Unable to load todos',
```
mbruhler, line 2: this should be uppercase

src/components/todolist/todolist.tsx
mbruhler: it depends on author imagination rather it's an error. I think that this solution works and while it works and is not a problematic thing it should be accepted. Everyone has it's own logic

mbruhler (changes requested):
Please refer to comments and fix the small issues

danon321:
In error message 'text' disappear immediately but background fade out. It will be better if both have this same visual effects

src/components/todolist/todolist.tsx
```diff
@@ -0,0 +1,181 @@
...
+      .catch(() => {
+        setError(ErrorType.delete);
+      })
+      .finally(() => setDeletingTask([]));
+  };
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {visibleTasks.map(task => (
+        <div
```
danon321, line 34: Create seperate component TodoItem

Barosz30 (approved):
Seems to work just fine! Gratz

src/components/todolist/todolist.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const handleDeleteClick = (id: number) => {
+    setError(null);
+
+    setDeletingTask([...deletingTask, id]);
+
+    deleteTodo(id)
+      .then(() => {
+        const filteredTodo = todos.filter(task => task.id !== id);
+
+        setTodos(filteredTodo);
```
mbruhler, line 24: use prevState here as we discussed during call, this is the same situation
So it would go like:
```
setTodos((currentTodos: Todo[]) => {
    return currentTodos.filter(task => task.id !== id);
});
```

mbruhler (changes requested):
As we found out today together, it's best to use prevState while updating state one by one. If this is just one payload to update the state, good old setTodos(todos) will work. But today we saw how it's problematic, so I want you to change it in deleting.

Also please clear the comments.

src/components/todofooter/todofooter.tsx
```diff
@@ -0,0 +1,94 @@
...
+      .finally(() => {
+        setTodos(unCompletedTodos);
+        setDeletingTask([]);
+      });
+  };
+
+  return (
+    <>
+      {todos.length > 0 && (
+        <footer className="todoapp__footer" data-cy="Footer">
```
mpiskadlo, line 39:
```suggestion
if (todos.length <= 0) return null;

  return (
        <footer className="todoapp__footer" data-cy="Footer">
```

src/App.tsx
```diff
@@ -1,24 +1,36 @@
...
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <TodoForm />
+
+        <TodoList />
+
+        {todos.length > 0
```
mpiskadlo, line 27: Check my comment in footer component. You won't need conditional rendering here ;)

src/components/todofooter/todofooter.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export const TodoFooter = () => {
+  const {
+    filterBy, setFilterBy, countIncompleteTask, todos,
+    setTodos, setError, setDeletingTask,
+  } = useTodos();
+
+  const hiddenBtn = todos.filter(el => el.completed).length === 0;
+
+  const DeleteCompletedTask = () => {
```
mpiskadlo, line 14: camelCase
```suggestion
  const deleteCompletedTask = () => {
```

mpiskadlo (changes requested):

mpiskadlo (approved):

PR #1008, opened 2024-01-04
Solution

web3-sorcerer:
[DEMO LINK](https://web3-sorcerer.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Todo/Todo.tsx
+++ b/src/components/Todo/Todo.tsx
@@ -1,0 +1,59 @@
+import { FC } from 'react';
+import cn from 'classnames';
+import { Todo as TodoTypes } from '../../types/Todo';
+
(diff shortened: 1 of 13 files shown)


aureliagorzna (approved):
Looks good

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,59 @@
+import { FC } from 'react';
```
danon321, line 1: TodoItem will be a better name. When i saw 'Todo' i was thinking more about whole Todo implementation

danon321 (changes requested):
if the title is empty show the Title should not be empty notification at the bottom;

that doesnt work

PR #1009, opened 2024-01-04
solution

aureliagorzna:
- [DEMO LINK](https://aureliagorzna.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -22,36 +22,33 @@ export const Header: React.FC<Props> = ({
   const inputRef = useRef<any>();
   const formRef = useRef<any>();
 
+  const handleAddTodo = async () => {
+    if (inputRef.current.value.trim() === '') {
+      handleError(ErrorType.TITLE);
+
+      return;
+    }
+
+    const newTodo: Todo = {
+      id: 0,
+      userId: USER_ID,
+      title: inputRef.current.value.trim(),
+      completed: false,
+    };
+
+    setIsLoading(prev => [...prev, newTodo.id]);
+    setTempTodo(newTodo);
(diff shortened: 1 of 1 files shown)


WiktoriaBielec (approved):
Great code!

danon321 (changes requested):
Dont have;
- if the title is empty show the Title should not be empty notification at the bottom;
- can add empty todo
- immediately after sending a request create a todo with id: 0 and save it to the tempTodo variable in the state (NOT to the todos array);
- covered the todo with the loader while waiting for the API response;

danon321 (changes requested):
Still dont have "Title should not be empty" error info when im trying to add empty todo.

Dont have "tempTodo" after adding new todo its just pop from nowhere.

After delete todo it just disappear instead should show loader:
covered the todo with the loader while waiting for the API response;

Not sure if 'checked' and 'unchecked' should work in this exercise but now it doesn't work

src/components/TodoList.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const deletingTodo = async () => {
+      try {
+        if (todos) {
+          await deleteTodo(id);
+          setTodos(prev => prev.filter(todo => todo.id !== id));
+        }
+      } catch (error) {
+        handleError(ErrorType.DELETE);
+      } finally {
+        setIsLoading(false);
```
danon321, line 38: Now only one item can be in "loading" state. Try to do it in array with ID items that are loading instead of just boolean

danon321 (changes requested):

danon321 (approved):
BEAUTIFUL!

PR #1012, opened 2024-01-08
Solution

AllaSerhiienko:
[DEMO LINK](https://AllaSerhiienko.github.io/react_todo-app-add-and-delete/)


DarkMistyRoom (approved):
Good job! [But for the future I suggest you to read this [article](https://blog.logrocket.com/pitfalls-of-overusing-react-context/) and use Context only where it's really needed](ttps://blog.logrocket.com/pitfalls-of-overusing-react-context/) and use Context only where it's really needed)

PR #1028, opened 2024-01-24
Develop

HokageHEL:
[DEMO LINK](https://HokageHEL.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,155 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Header } from './Components/Header';
+import { TodoList } from './Components/TodoList';
+import { Footer } from './Components/Footer';
+import { Notification } from './Components/Notiifcation';
+import { Todo } from './types/Todo';
+import { client } from './utils/fetchClient';
+import { FilterType } from './types/Filter';
+import { ErrorMessage } from './types/ErrorMessage';
 
-const USER_ID = 0;
+const USER_ID = 13;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState(FilterType.ALL);
+  const [errorMessege, setErrorMessege] = useState(ErrorMessage.NONE);
+  const [selectedId, setSelectedId] = useState(-1);
+  const [isLoadingCompleted, setIsLoadingCompleted] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    client.get<Todo[]>(`/todos?userId=${USER_ID}`)
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessege(ErrorMessage.CANNOT_LOAD_TODOS);
+      });
+  }, []);
+
+  const addTodo = (title: string) => {
+    setIsLoading(true);
+    setErrorMessege(ErrorMessage.NONE);
+    setTempTodo({
+      id: 0,
+      userId: USER_ID,
+      title,
+      completed: false,
+    });
+
+    client.post<Todo>('/todos', {
+      userId: USER_ID,
+      title,
+      completed: false,
+    })
+      .then(newTodo => {
+        setTodos((currentTodos) => [...currentTodos, newTodo]);
+      })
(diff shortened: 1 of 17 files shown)


src/Components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,49 @@
...
+        </CSSTransition>
+      ))}
+
+      {tempTodo !== null && (
+        <CSSTransition key={0} timeout={300} classNames="temp-item">
+          <TodoItem
+            key={tempTodo.id}
+            todo={tempTodo}
+            selectedId={selectedId}
+            isLoading={isLoading}
```
etojeDenys, line 43:
```suggestion
            isLoading
```
temp todo always must has the loader

src/App.tsx
```diff
@@ -1,24 +1,155 @@
...
-const USER_ID = 0;
+const USER_ID = 13;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState(FilterType.ALL);
+  const [errorMessege, setErrorMessege] = useState(ErrorMessage.NONE);
+  const [selectedId, setSelectedId] = useState(-1);
+  const [isLoadingCompleted, setIsLoadingCompleted] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
```
etojeDenys, line 21: looks like, you don't need this state

src/App.tsx
```diff
@@ -1,24 +1,155 @@
...
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <Header
+          todos={todos}
+          addTodo={addTodo}
+          setErrorMessege={setErrorMessege}
+          isLoading={isLoading}
```
etojeDenys, line 125:
```suggestion
          isLoading={!!tempTodo}
```
you can pass here tempTodo

src/App.tsx
```diff
@@ -1,24 +1,155 @@
...
+        />
+
+        <TodoList
+          todos={filteredTodos}
+          deleteTodo={deleteTodo}
+          selectedId={selectedId}
+          isLoadingCompleted={isLoadingCompleted}
+          tempTodo={tempTodo}
+          isLoading={isLoading}
```
etojeDenys, line 134: this prop looks redundant here

etojeDenys (changes requested):

src/App.tsx
HokageHEL: I need them for TodoItem

src/App.tsx
```diff
@@ -1,24 +1,151 @@
...
+
+    const deletePromises = completedTodos.map((todo) => client.delete(`/todos/${todo.id}`));
+
+    Promise.all(deletePromises)
+      .then(() => {
+        setTodos(currentTodo => currentTodo.filter((todo) => !todo.completed));
+      })
+      .catch((error) => {
+        setErrorMessege(ErrorMessage.UNABLE_TO_DELETE_A_TODO);
+        throw error;
```
TarasHoliuk, line 84: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

FYI

src/App.tsx
```diff
@@ -1,24 +1,151 @@
...
+      .then(() => setTodos(todos.filter((todo) => todo.id !== todoId)))
+      .catch(() => {
+        setErrorMessege(ErrorMessage.UNABLE_TO_DELETE_A_TODO);
+      })
+      .finally(() => {
+        setSelectedId(-1);
+      });
+  };
+
+  const deleteAllCompleted = () => {
```
TarasHoliuk, line 71: in the next task try to implement the loader inside `TodoItem` for such actions (deleteAllCompleted, toggle all todos status)

Hint: instead of having `selectedId` state with one id create a state with array of such ids

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+  isLoadingCompleted?: boolean;
+  isLoading?: boolean;
+};
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  deleteTodo,
+  selectedId,
+  isLoadingCompleted,
+  isLoading,
```
TarasHoliuk, line 18: currently, you use it only for tempTodo
But you may set the id of `tempTodo` (0) to `selectedId` and get rid of `isLoading` prop here

TarasHoliuk (approved):
GJ - Approved :heavy_check_mark: 

But added a few suggestions FYI (take them into account in the next task)

PR #1030, opened 2024-01-25
Develop

MarkiMark3:
(https://MarkiMark3.github.io/react_todo-app/)
idk what happened with my reset title. it doesnt work for some reason and idk why. I have like setTitleValue(' ') and it doesn't help

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,287 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { TodoContext } from './todoContext';
+import { TodoList } from './todoList';
 
-const USER_ID = 0;
+const USER_ID = 22;
 
 export const App: React.FC = () => {
+  // #region UseStates
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [count, setCount] = useState(0);
+  const [selectedFilter, setSelectedFilter] = useState('All');
+  const [error, setError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  // #endregion
+
+  let errorTimerId: NodeJS.Timeout;
+
+  const errorHandler = (errorPlace: string) => {
+    clearTimeout(errorTimerId);
+
+    switch (errorPlace) {
+      case 'title':
+        setErrorMessage('Title should not be empty');
+        break;
+      case 'add':
+        setErrorMessage('Unable to add a todo');
+        break;
+      case 'update':
+        setErrorMessage('Unable to update a todo');
+        break;
+      case 'delete':
+        setErrorMessage('Unable to delete a todo');
+        break;
+      default:
+    }
+
+    errorTimerId = setTimeout(() => {
+      setError(false);
+    }, 3000);
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(data => {
+        setTodos(data);
+        setCount(data.length);
+      })
+      .catch(e => {
+        setError(true);
+        errorHandler('add');
+        throw e;
+      });
+  }, []);
+
+  // #region Handlers
+
+  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+  };
+
+  const filteredTodo = todos.filter(todo => {
+    if (selectedFilter === 'Completed') {
+      return todo.completed;
+    }
+
+    if (selectedFilter === 'Active') {
+      return !todo.completed;
+    }
+
+    return true;
+  });
+
+  const handleErrorButton = () => {
+    setError(false);
+    errorHandler('title');
+  };
+
+  const handleFilter = (query: string) => {
+    setSelectedFilter(query);
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    const trimedTitle = title.trim();
+
+    if (!trimedTitle.length) {
+      setError(true);
+      errorHandler('title');
+    }
+
+    const addTodo = (newTodo: Todo) => {
+      setTodos([...todos, newTodo]);
+    };
+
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,24 +1,330 @@
...
+      case 'add':
+        setErrorMessage('Unable to add a todo');
+        break;
+      case 'update':
+        setErrorMessage('Unable to update a todo');
+        break;
+      case 'delete':
+        setErrorMessage('Unable to delete a todo');
+        break;
+      default:
```
TarasHoliuk, line 42:
```suggestion
      default:
        break;
```
or throw an error like `Unknown error type`

src/App.tsx
```diff
@@ -1,24 +1,330 @@
...
+  }, []);
+
+  // #region Handlers
+
+  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitleValue(event.target.value);
+  };
+
+  const filteredTodo = todos.filter(todo => {
+    if (selectedFilter === 'Completed') {
```
TarasHoliuk, line 71: create an enum for filter values
isn't better to use a switch here? WDYT?)

src/App.tsx
```diff
@@ -1,24 +1,330 @@
...
+    // eslint-disable-next-line
+  }, []);
+
+  // #region Handlers
+
+  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitleValue(event.target.value);
+  };
+
+  const filteredTodo = todos.filter(todo => {
```
TarasHoliuk, line 70:
```suggestion
  const filteredTodos = todos.filter(todo => {
```
it's an array of todo**s**

src/App.tsx
```diff
@@ -1,24 +1,330 @@
...
+  });
+
+  const handleErrorButton = () => {
+    setError(false);
+    errorHandler('title');
+  };
+
+  const handleFilter = (query: string) => {
+    setSelectedFilter(query);
+  };
```
TarasHoliuk, line 89: redundant function wrapper - just use `setSelectedFilter` in places you use `handleFilter`

src/App.tsx
```diff
@@ -1,24 +1,330 @@
...
+    errorTimerId = setTimeout(() => {
+      setError(false);
+    }, 3000);
+  };
+
+  useEffect(() => {
+    todoService.getTodos(USER_ID)
+      .then(data => {
+        setTodos(data);
+        setCount(data.length);
```
TarasHoliuk, line 54: you don't need state for it
just create `const todosTotalCount = todos.count`

General rule - do not create a separate state for value that can be calculated from another state value or prop

src/App.tsx
```diff
@@ -1,24 +1,330 @@
...
+const USER_ID = 22;
 
 export const App: React.FC = () => {
+  // #region UseStates
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [titleValue, setTitleValue] = useState('');
+  const [count, setCount] = useState(0);
+  const [selectedFilter, setSelectedFilter] = useState('All');
+  const [error, setError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
```
TarasHoliuk, line 19: you may join these 2 states
if errorMessage is an empty string (falsy value) it means that there is no error, so just get rid of `error` state

TarasHoliuk (changes requested):
Didn't get your comment - ask in the chat for help

Good progress, but there are things that should be fixed, improved

src/App.tsx
```diff
@@ -1,24 +1,330 @@
...
+  useEffect(() => {
+    todoService.getTodos(USER_ID)
+      .then(data => {
+        setTodos(data);
+        setCount(data.length);
+      })
+      .catch(e => {
+        setError(true);
+        errorHandler('add');
+        throw e;
```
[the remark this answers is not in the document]
lerastarynets: not fixed

src/App.tsx
lerastarynets: not fixed

src/App.tsx
```diff
@@ -283,7 +308,7 @@ export const App: React.FC = () => {
                     selected: selectedFilter === 'Completed',
                   })}
                   onClick={() => {
-                    handleFilter('Completed');
+                    setSelectedFilter('Completed');
```
lerastarynets, line 311: values 'Active', 'Completed', 'All' can also be moved to an enum

lerastarynets (changes requested):
Pls fix prev comments
also the following issue is still present
Screenshot at Jan 30 15-42-49

src/todoList/todoList.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { useContext } from 'react';
```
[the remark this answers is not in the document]
etojeDenys: image
image


must be `TodoList.tsx`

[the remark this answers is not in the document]
etojeDenys: fix it everywhere

src/todoItem/todoItem.tsx
```diff
@@ -0,0 +1,163 @@
...
+          </button>
+        )}
+
+        <div
+          data-cy="TodoLoader"
+          className={classNames('modal overlay', {
+            'is-active': (selectedTodo === activeTodo.id && isDeliting)
+              || (selectedTodo === activeTodo.id && isSubmitting)
+              || (clear && todo.completed),
+          })}
```
etojeDenys, line 155: better to move this long condition to some var

src/App.tsx
```diff
@@ -1,24 +1,361 @@
...
+enum FilterOption {
+  'Active',
+  'Completed',
+  'All',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [titleValue, setTitleValue] = useState('');
+  const [count, setCount] = useState(0);
```
etojeDenys, line 29: looks like a redundant state, it can be a simple var that will filter the array and save the array of active elements length

etojeDenys (changes requested):
center this section if we don't have completed todos
image

src/App.tsx
MarkiMark3: idk might be but this way I have countrol over count and my checkbox logic relies on it, so if I take it out everything brakes

etojeDenys (approved):
image

PR #1050, opened 2024-02-02
add task solution

avakiel:
[DEMO LINK](https://avakiel.github.io/react_todo-app-add-and-delete/)
All fixes and improvements allready in 3rd Task!

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,9 +2,7 @@
 /* eslint-disable max-len */
 /* eslint-disable no-console */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, {
-  Dispatch, SetStateAction, useEffect, useState,
-} from 'react';
+import React, { useEffect, useState } from 'react';
 import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
 import {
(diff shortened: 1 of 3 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,34 @@
...
+      case Filter.Completed:
+        return todo.completed;
+
+      case Filter.All:
+        return true;
+
+      default:
+        return true;
+    }
+  });
```
sasholk, line 26: Moving it to `services` would be fine, and I don't know if you really need `case Filter.All:` cuz you get it by default.

src/TodoContext.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable max-len */
+
+import { createContext } from 'react';
+import { TodoContextProps } from './types/interfaces';
+
+export const TodoContext = createContext<TodoContextProps | undefined>(undefined);
```
sasholk, line 6: interesting solution, but is that good, that you have so big `App` component because of it?

sasholk (approved):
Thats really nice work!🖖🤯
And really increadible of using `Context` something new for me.

src/TodoContext.ts
avakiel: its the first task where tried to use 'one' Context. I was lazy to write again useReducer, like in prev Todo App with localStorage.

avakiel: and all in App - just simplified task for myself

sasholk: Ah, I understood, need to see your finished todoapp)

PR #1057, opened 2024-02-04
Solution

VasiliyYarosh:


VasiliyYarosh:
DEMO - https://vasiliyyarosh.github.io/react_todo-app-add-and-delete/

Manankin (approved):
Good job !!!
You can make your code more readable if you create some components as Header, TodoList (and TodoItem inside it) and Footer.
But be careful when drill props inside this components

olya-shyian (approved):
I approve without a code review, I will do it already in the 3rd task.
Do not forget to pay attention to the comments from the PR of the first task (if there were any) before doing a PR for the 3rd task

Also try to split your code in the next task as Dmytro recommened🌼

PR #1065, opened 2024-02-05
Add task solution

dyarechuk:
[DEMO LINK](https://dyarechuk.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,128 @@
-/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import { prepareTodos } from './services/prepareTodos';
+import { wait } from './utils/fetchClient';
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,24 +1,128 @@
...
+    if (!cur.completed) {
+      return acc + 1;
+    }
+
+    return acc;
+  }, 0);
+
+  useEffect(() => {
+    setErrorMessage('');
+
```
nabokok, line 33: Do you need to set error here? You already have this value as initial state

src/App.tsx
```diff
@@ -1,24 +1,128 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (errorMessage) {
+      wait(3000).then(() => {
+        setErrorMessage('');
+      });
+    }
+  }, [errorMessage]);
+
```
nabokok, line 48: You can do it in Error component

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [value, setValue] = useState('');
+
+  const todoInput = useRef<HTMLInputElement>(null);
+
+  const newTodo = {
+    userId: USER_ID,
+    title: value.trim(),
+    completed: false,
+  };
+
```
nabokok, line 27: It is better to move this variable to handleSubmit func

nabokok (approved):
Good job!

PR #1071, opened 2024-02-06
react_todo-app-add-and-delete

NadiiaR:
[DEMO LINK](https://NadiiaR.github.io/react_todo-app-add-and-delete/)


src/components/ContextTodo/Context.tsx
```diff
@@ -0,0 +1,73 @@
...
+};
+
+export const ProviderTodo: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const [filterBy, setFilterBy] = useState(TodoFilter.All);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [updatedTodos, setUpdatedTodos] = useState<Todo[]>([]);
```
etojeDenys, line 36: it can be the simple var instead of the state

etojeDenys (changes requested):
1. hide the footer if we don't have the todos
image

SanyaBratashchuk (approved):

PR #1076, opened 2024-02-09
stage 2 solution

mariuszgit:


src/App.tsx
```diff
@@ -1,24 +1,46 @@
...
+          )}
+
+          <TodoForm />
+        </header>
+
+        <section className="todoapp__main">
+          <TodoList />
+        </section>
+
+        {todos.length > 0 && (
```
yevhenii-pyl, line 38: `todos.length` will be enough for this check.

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import cn from 'classnames';
+import { useTodoContext } from '../../context/TodoContext';
+
+export const TodoFooter = React.memo(() => {
```
yevhenii-pyl, line 5: As a rule of thumb, you can extract it from react in imports 

`import React, {memo} from 'react';`

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+  const removeAllCompleted = useCallback(() => {
+    todos.forEach(todo => {
+      if (todo.completed) {
+        removeTodo(todo.id);
+      }
+    });
+  }, [todos, removeTodo]);
+
+  const nrOfActiveTodos = useMemo(() => {
```
yevhenii-pyl, line 166: Do we really have to memoize it? 
It's just a simple `filter`: memorizing it may more expencive operation than filtering itself.

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,213 @@
...
+  removeTodo: (id: number) => void;
+  updateTodo: (itemId: number, completed: boolean) => void;
+  removeAllCompleted: () => void;
+  changeErrorMessage: (message: string) => void;
+  setError: React.Dispatch<SetStateAction<string | null>>;
+  toggleAll: () => void;
+  setFilter: React.Dispatch<SetStateAction<Filters>>;
+};
+
+export const TodoContext = React.createContext<ContextValue | null>(null);
```
yevhenii-pyl, line 36: Not a mistake here, but as a personal opinion: not a fan of creating a context for such a purpose. 
It may work for a small application, but may quicky become quite painful as the app grows. 

As an alternative for hooks: 
https://www.youtube.com/watch?v=SO8lBVWF2Y8

More on useContext: 
https://medium.com/@msgold/using-usecontext-in-react-a-comprehensive-guide-8a9f5271f7a8

src/App.tsx
```diff
@@ -1,24 +1,46 @@
...
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
+          {(todos.length > 0) && (
```
yevhenii-pyl, line 19: Just `todos.length`

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,213 @@
+import React, {
```
yevhenii-pyl, line 1: Strange style of imports, could you place them in column OR in row?

src/context/TodoContext.tsx
```diff
@@ -0,0 +1,213 @@
...
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(data => {
+        setTodos(data);
+        focusOnInput();
+      })
+      .catch(() => changeErrorMessage(ErrorMessages.DOWNLOAD));
+  }, [changeErrorMessage]);
+
+  const filteredTodos = useMemo(() => {
```
yevhenii-pyl, line 76: useMemo is not necessary, it may only reduce performance here

yevhenii-pyl (changes requested):
Good job!
However, could you check styling advices and revise your using of useMemo and useCallback? 
https://www.joshwcomeau.com/react/usememo-and-usecallback/

src/App.tsx
mariuszgit: need to use !! operator. Otherwise prints "0"

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <a
+          href="#/completed"
+          className={cn('filter__link', { selected: filter === 'completed' })}
+          onClick={() => setFilter('completed')}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {/* don't show this button if there are no completed todos */}
```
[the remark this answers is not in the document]
mariuszgit: disabled

src/App.tsx
yevhenii-pyl: Valid point! Got so used with conditional statements that overlooked the default behaviour here. 

If interested, check this package: https://www.npmjs.com/package/jsx-control-statements
I really encourage using it other that just having conditions on the jsx.

PR #1081, opened 2024-02-12
Solution part #2

kateryna-stemblivska:
[DEMO LINK](https://kateryna-stemblivska.github.io/react_todo-app-add-and-delete/)


IvanFesenko (approved):
Greate work. As recomendation to avoid code duplicating - create custom hook useLoader and move logic of showLoader handler there

PR #1088, opened 2024-02-19
effect todos

BOHDANLIUYUI:
[DEMO LINK](https://BOHDANLIUYUI.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/services/getFilteredTodos.ts
+++ b/src/services/getFilteredTodos.ts
@@ -1,0 +1,15 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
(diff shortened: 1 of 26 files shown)


src/services/getFilteredTodos.ts
```diff
@@ -0,0 +1,15 @@
...
+    switch (status) {
+      case Status.Active:
+        return !completed;
+      case Status.Completed:
+        return completed;
+      default:
+        return true;
+    }
+  });
+}
```
IvanFesenko, line 15: No need to filter in case status is all

IvanFesenko (approved):
Nice job

PR #1108, opened 2024-03-12
react_todo-app-add-and-delete

Hovera-Liubomyr:
https://Hovera-Liubomyr.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -1,0 +1,20 @@
+import { Todo } from '../types/Todo';
+import { USER_ID } from '../utils/USER_ID';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+// Add more methods here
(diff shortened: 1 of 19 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
+import { Todo } from '../types/Todo';
+import { USER_ID } from '../utils/USER_ID';
+import { client } from '../utils/fetchClient';
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+// Add more methods here
```
DarkMistyRoom, line 9: remove the comment

DarkMistyRoom (changes requested):
- the filter links should always be placed in the center of the footer
image

- the counter shouldn't depend on the filter - it should always show the number of uncompleted todos

IvanFesenko (approved):
Fix it on next task:
- in case deleteding loader shoud be on todo that current deleting, not last one 
2024-03-13_15h50_30

PR #1111, opened 2024-03-14
added and deleted gor todos

MartaChobaniuk:
[DEMO LINK](https://martachobaniuk.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/utils/getTodoFilter.ts
+++ b/src/utils/getTodoFilter.ts
@@ -1,0 +1,17 @@
+import { Filter } from '../types/Filter';
+import { Todo } from '../types/Todo';
+
+export const getTodoFilter = (todos: Todo[], filter: Filter) => {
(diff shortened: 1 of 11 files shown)


src/utils/getTodoFilter.ts
```diff
@@ -0,0 +1,17 @@
...
+      case Filter.Active:
+        return !todo.completed;
+
+      case Filter.Completed:
+        return todo.completed;
+
+      default:
+        return true;
+    }
+  });
```
SanyaBratashchuk, line 16: it is better to move filter inside switch case not vice versa to avoid extra iteration in defaul case

SanyaBratashchuk (changes requested):

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1115, opened 2024-03-17
Add task solution

Eater228:


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,65 @@
...
+    deleteTodos(todoId)
+      .then(() => {
+        setTodos(currentTodos =>
+          currentTodos.filter(post => post.id !== todoId),
+        );
+      })
+      .catch(error => {
+        setErrorMessage('Unable to delete a todo');
+        setTodos(todos);
+        throw error;
```
VitaliyBondarenko1982, line 27:
```suggestion
```
it is redundant. setErrorMessage is enough for handling error. and we need to move on finally block.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,65 @@
...
+      .finally(() => {
+        if (todoInput) {
+          todoInput.focus();
+        }
+      });
+  };
+
+  const handleTodoCleaning = () => {
+    todos.map(todo => todo.completed && handleDelete(todo.id));
+    setTodos(todos.filter(todo => !todo.completed));
```
VitaliyBondarenko1982, line 38:
```suggestion
    const completedTodos = todos.filter(todo => todo.completed);
    completedTodos.forEach(todo => handleDelete(todo.id));
```
looks like it is reason of fail this test.
you are already filter todos in then block.
Screenshot from 2024-03-17 15-27-05

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,131 @@
...
+    };
+
+    setTempTodo(newTodo);
+    addTodo(newTodo);
+  };
+
+  const handleAddTodo = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (title.length < 1 || title.trim() === '') {
+      setErrorMessage('Title should not be empty');
```
VitaliyBondarenko1982, line 48: make enum for all possible error messages. and use it here and in all other places.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments and try to pass one failed test.

src/components/Footer/Footer.tsx
Eater228: The problem is that one day it passes and the next it doesn't.

Eater228:
[DEMO LINK](https://Eater228.github.io/react_todo-app-add-and-delete/)

nazarmatsevych (approved):
Awesome 🔥

PR #1121, opened 2024-03-20
Add and delete Todo

nk8bnj:
[DEMO LINK](https://nk8bnj.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoFooter/TodoFooter.tsx
+++ b/src/components/TodoFooter/TodoFooter.tsx
@@ -14,7 +14,7 @@ const Footer: React.FC = () => {
     filter,
     setFilter,
     setIsAllDeleted,
-    handleError,
+    handleError
   } = useTodos();
 
   const isClearButtonVisible = todos.some(todo => todo.completed);
--- a/src/components/TodoFooter/TodoFooter.tsx
+++ b/src/components/TodoFooter/TodoFooter.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+
+import TodoFilter from '../TodoFilter';
+
+import { deleteTodo } from '../../api/todos';
+import { useTodos } from '../../hooks/useTodos';
+import { Todo } from '../../types';
+
+const Footer: React.FC = () => {
+  // eslint-disable-next-line prettier/prettier
+  const {
+    todos,
+    setTodos,
+    filter,
+    setFilter,
+    setIsAllDeleted,
+    handleError,
+  } = useTodos();
+
+  const isClearButtonVisible = todos.some(todo => todo.completed);
+  const amountItemsLeft = todos.filter(todo => !todo.completed).length;
+
+  const handleDeletetodo = (todoId: number) => {
+    setIsAllDeleted(true);
+
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos((prevTodos: Todo[]) =>
+          prevTodos.filter(todo => todo.id !== todoId),
+        );
+      })
+      .catch(() => {
+        handleError('Unable to delete a todo');
+      })
+      .finally(() => {
+        setIsAllDeleted(false);
+      });
+  };
+
+  const handleClearCompleted = () => {
+    todos.forEach(todo => {
+      if (todo.completed) {
+        handleDeletetodo(todo.id);
+      }
+    });
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {amountItemsLeft} items left
+      </span>
+
+      <TodoFilter filter={filter} setFilter={setFilter} />
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        onClick={handleClearCompleted}
+        disabled={!isClearButtonVisible}
+      >
+        Clear completed
+      </button>
+    </footer>
+  );
+};
+
+export default Footer;
--- a/src/components/TodoInput/TodoInput.tsx
+++ b/src/components/TodoInput/TodoInput.tsx
@@ -1,0 +1,110 @@
+import React, { useEffect, useRef } from 'react';
+
+import { useTodos } from '../../hooks/useTodos';
+import { USER_ID, addTodo } from '../../api/todos';
+
+const TodoInput: React.FC = () => {
+  const [isAllCompleted, setIsAllCompleted] = React.useState(false);
+
+  const {
+    todos,
+    setTodos,
+    setError,
+    isLoading,
+    setIsLoading,
+    setTempTodo,
+    handleError,
+    isAllDeleted,
+  } = useTodos();
+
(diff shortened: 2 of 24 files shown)


src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,69 @@
...
+    setFilter,
+    setIsAllDeleted,
+    handleError
+  } = useTodos();
+
+  const isClearButtonVisible = todos.some(todo => todo.completed);
+  const amountItemsLeft = todos.filter(todo => !todo.completed).length;
+
+  const handleDeletetodo = (todoId: number) => {
+    setIsAllDeleted(true);
```
BudnikOleksii, line 24: Do you delete only one todo or all here? Looks like this state should be `isLoading`

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,69 @@
...
+        setIsAllDeleted(false);
+      });
+  };
+
+  const handleClearCompleted = () => {
+    todos.forEach(todo => {
+      if (todo.completed) {
+        handleDeletetodo(todo.id);
+      }
+    });
```
BudnikOleksii, line 45: Ah, I see. You can create a state `loadingTodoIds` or something similar, and add ids to this array when deleting/toggling them and depending on this state show loaders for the corresponding todos

src/components/TodoInput/TodoInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const handleEnterEvent = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      event.preventDefault();
+      handleAddTodo();
+    }
+  };
+
+  const handleToggleAll = () => {
+    setTodos(todos.map(todo => ({ ...todo, completed: !isAllCompleted })));
+    setIsAllCompleted(!isAllCompleted);
```
BudnikOleksii, line 80: When you need to change state depends on the previous one, use functional `setState`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { deleteTodo } from '../../api/todos';
+import { useTodos } from '../../hooks/useTodos';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo;
+};
+
+const TodoItem: React.FC<Props> = ({ todo }) => {
+  const [selectedTodoId, setSelectedTodoId] = React.useState<number>(0);
```
BudnikOleksii, line 13: Why do you store the selected todo id inside the todo item? And why do you set it when you actually delete it

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+      .then(() => {
+        setTodos(todos.filter(prevTodo => prevTodo.id !== todo.id));
+      })
+      .catch(() => {
+        handleError('Unable to delete a todo');
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  };
```
BudnikOleksii, line 48: You have the same handler in `TodoFooter`, maybe you can move it to the context

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const firstCondition = isLoading && todo.id === selectedTodoId;
+  const secondCondition = todo.completed && isAllDeleted;
+  const isLoaderActive = firstCondition || secondCondition;
```
BudnikOleksii, line 52: That is strange and too complicated logic and these conditions will be more. You can create a state where you can store ids of todos that currently loading and depending on it set a loader.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  return (
+    <div
+      data-cy="Todo"
+      className={cn('todo', {
+        completed: todo.completed,
+      })}
+    >
+      <label className="todo__status-label">
```
BudnikOleksii, line 61: A form label must be associated with a control

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <button
+        type="button"
+        className="todo__remove"
+        data-cy="TodoDelete"
+        onClick={handleDeleteTodo}
+      >
+        ×
+      </button>
+
+      {/* overlay will cover the todo while it is being deleted or updated */}
```
BudnikOleksii, line 84: Remove comments

BudnikOleksii (changes requested):
Good job 👍 
Let's fix some issues and improve your solution

src/components/TodoInput/TodoInput.tsx
BudnikOleksii: > When you need to change state depends on the previous one, use functional `setState`

Not fixed
Also, looks like this state is redundant, as this value should be dynamic

src/components/TodoItem/TodoItem.tsx
BudnikOleksii: > Why do you store the selected todo id inside the todo item? And why do you set it when you actually delete it

?

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  const handleDeleteTodo = () => {
+    setSelectedTodoId(todo.id);
+    onDeleteTodo(todo.id);
+  };
+
+  // eslint-disable-next-line prettier/prettier
+  const isLoaderActive = (isLoading && todo.id === selectedTodoId)
+  || (todo.completed && isAllDeleted);
```
BudnikOleksii, line 33: That is strange and too complicated logic and these conditions will be more. You can create a state where you can store ids of todos that currently loading and depending on it set a loader.

BudnikOleksii (changes requested):
Good progress, but some moments still should fixed.
in the next task you need to add functionality to toggle all/remove completed, and need to show loaders for corresponding todos, so let's think about it update the logic of setting loaders

src/context/TodosContext.tsx
```diff
@@ -0,0 +1,107 @@
...
+        setTodos((prevTodos: Todo[]) =>
+          prevTodos.filter(todo => todo.id !== todoId),
+        );
+      })
+      .catch(() => {
+        handleError('Unable to delete a todo');
+      })
+      .finally(() => {
+        setIsLoading(false);
+        setLoadingTodosIDs([]);
```
BudnikOleksii, line 64:
```suggestion
        setLoadingTodosIDs(prevLoadingTodosIDs => (
          LoadingTodosIDs.filter(id => id !==todoId)
        ));
```
You can filter loading todo ids in case you have other loading states

BudnikOleksii (approved):
Looks good 👍

PR #1132, opened 2024-03-25
Develop

NVlasenko:
[DEMO LINK](https://NVlasenko.github.io/react_todo-app-add-and-delete/)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,206 @@
...
+      });
+  };
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      <TransitionGroup>
+        {filter.map(todo => (
+          <CSSTransition key={todo.id} timeout={300} classNames="item">
+            <div
+              key={todo.id}
```
etojeDenys, line 78: redundant here, 'cause you specified it above

src/components/TodoList.tsx
```diff
@@ -0,0 +1,206 @@
...
+              data-cy="Todo"
+              className={`todo ${todo.completed ? 'completed' : ''}`}
```
etojeDenys, line 80: better to use classnames package for conditional classes

etojeDenys (changes requested):
everything looks cool, but you forgot to implement the temp todo logic


feel free to ask for some help with that in the chat

andriy-shymkiv (approved):

PR #1135, opened 2024-03-28
add solutions

shikoryak:
https://shikoryak.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,148 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { USER_ID, addTodo, deleteTodo, getTodos } from './api/todos';
+import { UpdatingStatus, Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Filter, Footer } from './components/Footer';
+import { ErrorMessage, Notification } from './components/Notification';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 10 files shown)


src/components/Header.tsx
```diff
@@ -0,0 +1,53 @@
...
+  status: UpdatingStatus;
+};
+
+export const Header: React.FC<Props> = ({ onAdd, status }) => {
+  const [title, setTitle] = useState('');
+  const titleField = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    titleField.current?.focus();
+    if (status === UpdatingStatus.success) {
```
TarasHoliuk, line 20: you don't use `failed` status. So it can be replaced with a boolean value.
Moreover, it can be `const isInProgess = !!processingIds.length`

Variable name is only example

src/components/Footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+          Completed
+        </a>
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        disabled={completedItemsLeft < 1}
+        onClick={() => onClearCompleted()}
```
TarasHoliuk, line 79:
```suggestion
        onClick={onClearCompleted}
```

src/App.tsx
```diff
@@ -1,26 +1,148 @@
...
+        setUpdatingStatus(UpdatingStatus.failed);
+      })
+      .finally(() => {
+        setTempTodo(null);
+      });
+  };
+
+  const onDeleteHandler = (id: number) => {
+    setUpdatingStatus(UpdatingStatus.inProgres);
+    setProcessingIds(prevIds => [...prevIds, id]);
```
TarasHoliuk, line 88: :fire:

TarasHoliuk (approved):
Great job :+1:

1 comment only - about `updatingStatus`. Fix it in the next task stage

PR #1136, opened 2024-03-29
Develop

olduma:
[DEMO LINK](https://olduma.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/TodoFooter/TodoFooter.tsx
+++ b/src/components/TodoFooter/TodoFooter.tsx
@@ -10,7 +10,11 @@ export const TodoFooter: React.FC = () => {
     todos
       .filter(todo => todo.completed)
       .map(todo => todo.id)
-      .forEach(id => removeTodo(id));
+      .forEach(id => {
+        if (typeof id === 'number') {
+          removeTodo(id);
+        }
+      });
   };
 
   return (
--- a/src/components/TodoFooter/TodoFooter.tsx
+++ b/src/components/TodoFooter/TodoFooter.tsx
@@ -2,12 +2,15 @@ import { useTodos } from '../../utils/TodoContext';
 import { TodoFilter } from '../TodoFilter';
 
 export const TodoFooter: React.FC = () => {
-  const { todos, setTodos } = useTodos();
+  const { todos, removeTodo } = useTodos();
   const isClearButtonActive = todos.some(todo => todo.completed);
   const countActiveTodos = todos.filter(todo => !todo.completed).length;
 
   const clearAllCompleted = () => {
-    setTodos(prevTodos => prevTodos.filter(todo => !todo.completed));
+    todos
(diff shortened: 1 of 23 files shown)


src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  const clearAllCompleted = () => {
+    todos
+      .filter(todo => todo.completed)
+      .map(todo => todo.id)
+      .forEach(id => {
+        if (typeof id === 'number') {
+          removeTodo(id);
+        }
+      });
```
BudnikOleksii, line 17:
```suggestion
      .forEach(todo => {
        removeTodo(todo.id);
      });
```
Why do you need to check the id type?

src/components/TodoHeader/TodoHeader.tsx
```diff
@@ -0,0 +1,87 @@
...
+      completed: false,
+      userId: USER_ID,
+    });
+
+    const newTodo = {
+      id: 0,
+      title: query.trim(),
+      completed: false,
+      userId: USER_ID,
+    };
```
BudnikOleksii, line 47:
```suggestion
    const newTodo = {
      id: 0,
      title: query.trim(),
      completed: false,
      userId: USER_ID,
    };
    
    setTempTodo(newTodo);
```
DRY

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+          prevTodos.map(prev =>
+            prev.id === id ? { ...prev, title: newText } : prev,
+          ),
+        );
+      }
+
+      setEditingTodoId(null);
+    } else if (e.key === 'Escape') {
+      setEditingTodoId(null);
+    }
```
BudnikOleksii, line 53:
```suggestion
      setEditingTodoId(null);
      return;
    }
    
    if (e.key === 'Escape') {
      setEditingTodoId(null);
    }
```

src/utils/TodoContext.tsx
```diff
@@ -0,0 +1,126 @@
...
+        setTodos(prevTodos => [...prevTodos, response]);
+        setQuery('');
+      })
+      .catch(() => {
+        setError(ErrorMessage.ADD_TODO_ERROR);
+      })
+      .finally(() => {
+        setTempTodo(null);
+        setIsLoading(false);
+        setLoadingTodosIDs([]);
```
BudnikOleksii, line 63: In the same way as you add, remove only needed todo ids and keep others, fix it in all places

BudnikOleksii (changes requested):
Good work 👍 
just a few minor fixes are required

BudnikOleksii (approved):
👍

PR #1142, opened 2024-04-01
add task solution

pawelnowicki87:
[DEMO LINK](https://pawelnowicki87.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,187 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { USER_ID, deleteTodo, getTodos, postTodo } from './api/todos';
+import { TodoList } from './components/TodoList';
+import { HeaderInput } from './components/HeaderInput';
+import { Footer } from './components/Footer';
+import classNames from 'classnames';
+import { Errors, SelectedTasks, Todo } from './types/Types';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [selectedTasks, setSelectedTasks] = useState<SelectedTasks>(
+    SelectedTasks.All,
+  );
+  const [tempTodo, setTempTodo] = useState<Todo | null>();
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,187 @@
...
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [selectedTasks, setSelectedTasks] = useState<SelectedTasks>(
+    SelectedTasks.All,
+  );
+  const [tempTodo, setTempTodo] = useState<Todo | null>();
+  const [title, setTitle] = useState('');
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [isAddingTodo, setIsAddingTodo] = useState(false);
+  const [deleteTodoByID, setDeleteTodoByID] = useState<number | null>();
+
```
yevhenii-pyl, line 21: As a matter of code structuring: useStates => useRef, we shouldn't mix them.

src/App.tsx
```diff
@@ -1,26 +1,187 @@
...
+      })
+      .catch(() => setErrorMessage(Errors.Add))
+      .finally(() => {
+        setTempTodo(null);
+        setIsAddingTodo(false);
+      });
+  };
+
+  const deleteCurrentTodo = useCallback((id: number) => {
+    setDeleteTodoByID(id);
```
yevhenii-pyl, line 77: Let's also add a toggling todo loading state logic here.

src/App.tsx
```diff
@@ -1,26 +1,187 @@
...
+        setTodos(prev => prev.filter(todo => todo.id !== id));
+      })
+      .catch(() => {
+        setErrorMessage(Errors.Delete);
+      })
+      .finally(() => setDeleteTodoByID(null));
+  }, []);
+
+  const clearCompleted = async () => {
+    const todoToClear: Todo[] = [];
```
yevhenii-pyl, line 89: Hmmm, let's try to simplify it a bit here: you can create an array with todoToClear by filtering current todos on completed status. 

Then, you can use Promise.all to run all deletion operations at once by mapping through todoToClear.

Also, it's worth having a toggling loading state logic here.

yevhenii-pyl (changes requested):
Solid! 
Let's include adjustments from the previous PR + new comments and this will be a  decent amount of work here!

src/App.tsx
yevhenii-pyl: Ah, I guess my comment was misleading, my bad. You did a good job by using the useRef, I only wanted to point your attention to the hooks order in your file, it should be:

```
useStates
useRefs
useEffects
```
OR
```
useRefs
useStates
useEffects
```

So the structure is always the same, not 

```
useState
useState
useRef
useEffect
useState
```

PR #1145, opened 2024-04-01
solution#1

ChrisHryts:
https://chrishryts.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -1,0 +1,25 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 332;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+export const addTodo = (data: Omit<Todo, 'id'>) => {
+  return client.post<Todo>(`/todos`, data);
+};
+
+export const deleteTodo = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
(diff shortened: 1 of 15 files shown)


src/api/todos.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+// export const editTodos = ({ id, userId, title, completed }: Todo) => {
+//   return client.patch<Todo>(`/todos/${id}?userId=${USER_ID}`, {
+//     id,
+//     userId,
+//     title,
+//     completed,
+//   });
+// };
```
DenNice-cloud, line 25: Remove comments or use this part if it's needed

DenNice-cloud (approved):
Amazing code, interesting realisation with context!
Dont know what to add.., wow!🤯

src/api/todos.ts
polosanya: ➕ , don't forget to remove the redundant comments

src/components/Footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+    todos,
+    setTodos,
+    query,
+    setQuery,
+    setLoadingTodosIds,
+    handleError,
+    setIsInputFocused,
+  } = useTodosContext();
+  const completedIds = useMemo(() => {
+    return todos.filter(item => item.completed === true).map(item => item.id);
```
polosanya, line 19:
```suggestion
    return todos
      .filter(item => item.completed)
      .map(item => item.id);
```
Can be simplified. Btw, you can use reduce method instead of two methods

polosanya (approved):
Great job! I'm approving it, but please pay attention to the comments)

PR #1155, opened 2024-04-03
add task solution

kshepetska:
[DEMO LINK](https://kshepetska.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,208 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useState, useEffect } from 'react';
+import cn from 'classnames';
 
-const USER_ID = 0;
+import { UserWarning } from './UserWarning';
+import { USER_ID } from './api/todos';
+import Header from './components/Header';
+import Footer from './components/Footer';
+import TodoList from './components/TodoList';
+import { Status } from './types/Status';
+import * as postService from './api/todos';
+import { Todo } from './types/Todo';
+import { handleError } from './components/Error';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<Status>(Status.All);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [inputDisabled, setInputDisabled] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [shouldFocusInput, setShouldFocusInput] = useState(false);
+
+  useEffect(() => {
+    setLoading(true);
+    postService
+      .getTodos()
+      .then(fetchedTodos => {
+        setTodos(fetchedTodos);
+      })
+      .catch(() => {
+        handleError('Unable to load todos', setErrorMessage);
+      })
+      .finally(() => {
+        setLoading(false);
+      });
+  }, []);
+
+  useEffect(() => {
+    if (shouldFocusInput) {
+      setShouldFocusInput(false);
+    }
+  }, [shouldFocusInput]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
-  return (
-    <section className="section container">
-      <p className="title is-4">
-        Copy all you need from the prev task:
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">
-          React Todo App - Load Todos
-        </a>
-      </p>
+  const itemsLeft = todos.filter(({ completed }) => !completed).length;
+  const haveCompletedTodos = todos.some(({ completed }) => completed);
+  const allCompleted = todos.every(({ completed }) => completed);
 
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+  const filteredTodos = todos.filter(task => {
+    switch (status) {
+      case Status.Active:
+        return !task.completed;
+      case Status.Completed:
+        return task.completed;
+      case Status.All:
+      default:
+        return true;
+    }
+  });
+
+  const addTodo = (title: string) => {
+    const trimmedTitle = title.trim();
+
+    if (!trimmedTitle) {
+      handleError('Title should not be empty', setErrorMessage);
+
+      return;
+    }
+
+    setInputDisabled(true);
+
+    const newTodo: Todo = {
+      id: Math.max(0, Math.max(...todos.map(todo => todo.id))) + 1,
+      title: trimmedTitle,
+      completed: false,
+      userId: USER_ID,
+      status: status,
+    };
+
+    const fakeTodo: Todo = {
+      id: 0,
+      title: trimmedTitle,
+      completed: false,
+      userId: USER_ID,
+      status: status,
+    };
+
+    setTempTodo(fakeTodo);
+
+    postService
+      .postTodo(newTodo)
+      .then(() => {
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,26 +1,208 @@
...
+    if (!trimmedTitle) {
+      handleError('Title should not be empty', setErrorMessage);
+
+      return;
+    }
+
+    setInputDisabled(true);
+
+    const newTodo: Todo = {
+      id: Math.max(0, Math.max(...todos.map(todo => todo.id))) + 1,
```
vladarskyi, line 76: This may not be a problem, but the server itself adds unique IDs

vladarskyi (changes requested):
Good work! Some points need to be corrected:
- After error `Title should not be empty` (with whitespaces) you don't need to clear the input field;
- The text field should be in focus after removing the individual todo and after `Clear completed`.

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+import { Todo } from './types/Todo';
+import { handleError } from './components/Error';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<Status>(Status.All);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [inputDisabled, setInputDisabled] = useState(false);
+  const [loading, setLoading] = useState(false);
```
BudnikOleksii, line 20: - Boolean variables should start from `is` or `has`.

Hint: instead you can create a state with an array of loading todo ids. So you can add and remove ids of todos that currently loading and show loaders for them. It can help with the next task

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+        setTodos(fetchedTodos);
+      })
+      .catch(() => {
+        handleError('Unable to load todos', setErrorMessage);
+      })
+      .finally(() => {
+        setLoading(false);
+      });
+  }, []);
+  useEffect(() => {
```
BudnikOleksii, line 37: Add space between `useEffect`

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+    switch (status) {
+      case Status.Active:
+        return !task.completed;
+      case Status.Completed:
+        return task.completed;
+      case Status.All:
+      default:
+        return true;
+    }
+  });
```
BudnikOleksii, line 61: Create a function that will do it and move this function to utils/helpers file/folder

src/App.tsx
BudnikOleksii: > This may not be a problem, but the server itself adds unique IDs

Yes, you don't need id

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+      status: status,
+    };
+
+    const fakeTodo: Todo = {
+      id: 0,
+      title: trimmedTitle,
+      completed: false,
+      userId: USER_ID,
+      status: status,
+    };
```
BudnikOleksii, line 88:
```suggestion
    const fakeTodo: Todo = {
      id: 0,
      ...newTodo,
    };
```

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+  const updateTodo = (patchTodo: Todo) => {
+    postService.updateTodo(patchTodo).then(task => {
+      setTodos(currentTodos => {
+        const newTodos = [...currentTodos];
+        const index = newTodos.findIndex(({ id }) => id === task.id);
+
+        newTodos[index] = task;
+
+        return newTodos;
+      });
```
BudnikOleksii, line 117:
```suggestion
      setTodos(currentTodos => (
        currentTodos.map((todo) => todo.id === task.id ? task : todo)
      ));
```
Use map instead

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+  };
+
+  const handleClearCompleted = () => {
+    const completedTodoIds = todos
+      .filter(duty => duty.completed)
+      .map(duty => duty.id);
+
+    completedTodoIds.forEach(id => {
+      postService
+        .deleteTodo(`/todos/${id}`)
```
BudnikOleksii, line 144:
```suggestion
        .deleteTodo(id)
```
`handleClearCompleted` shouldn't know anything about URL, pass only id and manage URL in the service function

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const updatedTodo: Todo = {
+      ...todo,
+      completed: !todo.completed,
+    };
+
+    if (todo.status === Status.Active) {
+      updatedTodo.status = Status.Completed;
+    } else if (todo.status === Status.Completed) {
+      updatedTodo.status = Status.Active;
+    }
```
BudnikOleksii, line 52: You don't need to add a new field `status` for todos

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+          className={cn('todo__title', { hidden: isEditing })}
+          onClick={handleEditToggle}
+        >
+          {todo.title}
+        </span>
+      )}
+
+      <div
+        data-cy="TodoLoader"
+        className={`modal overlay ${!todo.id || (loading && todo.id === selectedTodo?.id) ? 'is-active' : ''}`}
```
BudnikOleksii, line 93: Use classnames library to set classes conditionally

BudnikOleksii (changes requested):
Good work 👍 
Left some suggestions in comments, check them

src/App.tsx
StasSokolov1: not fixed

src/App.tsx
```diff
@@ -1,26 +1,192 @@
...
+          errorMessage={errorMessage}
+        />
+        <TodoList
+          todos={filteredTodos}
+          onDeleteTodo={deleteTodo}
+          updateTodo={updateTodo}
+          tempTodo={tempTodo}
+          isLoading={isLoading}
+        />
+        {todos.length > 0 && (
```
StasSokolov1, line 160:
```suggestion
        {!!todos.length && (
```

src/components/Filter.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { Status } from '../types/Status';
+import cn from 'classnames';
+
+interface FilterProps {
+  status: Status;
+  onStatusChange: (status: Status) => void;
+}
+
+const Filter: React.FC<FilterProps> = ({ status, onStatusChange }) => {
+  return (
```
StasSokolov1, line 11:
```suggestion
const Filter: React.FC<FilterProps> = ({ status, onStatusChange }) => (
```

check in all places

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <div className="loader" />
+      </div>
+
+      <button
+        type="button"
+        className="todo__remove"
+        data-cy="TodoDelete"
+        onClick={() => {
+          onDeleteTodo(todo.id);
+          setSelectedTodo(todo);
```
StasSokolov1, line 100: let's move this logic into separated function

src/components/TodoList.tsx
```diff
@@ -0,0 +1,45 @@
...
+          isLoading={isLoading}
+        />
+      ))}
+
+      {tempTodo && (
+        <TodoItem
+          todo={tempTodo}
+          key={tempTodo.id}
+          onDeleteTodo={() => {}}
+          onUpdateTodo={() => {}}
```
StasSokolov1, line 37: let's remove this and set this parameter optional

StasSokolov1 (changes requested):
Almost done, let's improve your code a bit more🔥

PR #1157, opened 2024-04-03
Add second part

winniepuhh:


src/App.tsx
```diff
@@ -1,26 +1,151 @@
...
+          inputRef={inputRef}
+        />
+        <TodoList
+          filteredTodos={filteredTodos}
+          TodoDeleteButton={TodoDeleteButton}
+          isLoading={isLoading}
+          tempTodo={tempTodo}
+        />
+
+        {todos.length > 0 && (
```
vlesia, line 139: You can use only todos.length, if todos.length === 0 it will be false

vlesia (approved):
Good job!

src/App.tsx
```diff
@@ -1,26 +1,151 @@
...
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos(prevTodos => prevTodos.filter(todo => todo.id !== todoId));
+        setIsLoading(0);
+      })
+      .catch(() => {
+        setError(ErrorTypes.UnableToDelete);
+        setTimeout(() => {
+          setError('');
+        }, 3000);
```
BudnikOleksii, line 50: Create a handler that sets the error message and then removes it in 3 seconds as you do it for all requests
```typescript
const handleError = (errorMessage: ErrorTypes) => {
  setError(errorMessage);
  setTimeout(() => {
    setError('');
  }, 3000);
}; 
```

src/App.tsx
```diff
@@ -1,26 +1,151 @@
...
+      .catch(() => {
+        setError(ErrorTypes.OneMessage);
+        setTimeout(() => {
+          setError('');
+        }, 3000);
+      });
+  }, []);
+
+  const removeTodo = (todoId: number) => {
+    setIsLoading(todoId);
```
BudnikOleksii, line 40: Instead of setting a single id you can create a state with an array of loading todo ids. So you can add and remove ids of todos that currently loading and show loaders for them. It can help with the next task

src/Components/ErrorNotification.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+type Props = {
+  error: string;
+  onDismiss: () => void;
+};
+
+export const ErrorNotification: React.FC<Props> = ({ error, onDismiss }) => {
+  const handleDismiss = () => {
+    onDismiss();
+  };
```
BudnikOleksii, line 11: This function is redundant, just pass `onDismiss` to `onClick`

src/additional/additional.tsx
BudnikOleksii, line 1: Remove unnecessary file

BudnikOleksii (changes requested):
Good job 👍 
Left a few suggestions, check them

src/App.tsx
```diff
@@ -1,26 +1,158 @@
...
+    }, 3000);
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .then(() => {
+        if (isLoading.length > 0) {
+          setIsLoading([]);
+        }
```
BudnikOleksii, line 41: You should remove loaders in the finally block as you have to do it in both cases

src/App.tsx
```diff
@@ -1,26 +1,158 @@
...
+    setIsLoading(prev => [...prev, todoId]);
+    deleteTodo(todoId)
+      .then(() => {
+        setTodos(prevTodos => prevTodos.filter(todo => todo.id !== todoId));
+        setIsLoading([]);
+      })
+      .catch(() => {
+        setIsLoading([]);
+        handleError(ErrorTypes.UnableToDelete);
+      });
```
BudnikOleksii, line 59:
```suggestion
        setTodos(prevTodos => prevTodos.filter(todo => todo.id !== todoId));
      })
      .catch(() => {
        handleError(ErrorTypes.UnableToDelete);
      })
      .finally(() => {
        setIsLoading(prevIds => prevIds.filter(id => id !== todoId));
      });
```
And remove loaders only what you actually need

src/App.tsx
```diff
@@ -1,26 +1,158 @@
...
+        setTodos(prevTodos => [...prevTodos, savedTodo]);
+        setValue('');
+      })
+      .catch(() => {
+        handleError(ErrorTypes.UnableToAdd);
+      })
+      .finally(() => {
+        setInputDisabled(false);
+        setTempTodo(null);
+        setIsLoading([]);
```
BudnikOleksii, line 83: And the same in all places

src/App.tsx
```diff
@@ -1,26 +1,158 @@
...
+  const handleSubmit: (
+    event: React.FormEvent<HTMLFormElement>,
+  ) => void = event => {
+    event.preventDefault();
+
+    if (!value.trim()) {
+      setError(ErrorTypes.TitleNotEmpty);
+      setTimeout(() => {
+        setError('');
+      }, 3000);
```
BudnikOleksii, line 96: You have `handleError` function so use it

BudnikOleksii (approved):
Good job 👍 
Approved, but fix what I've mentioned in the next task

PR #1158, opened 2024-04-03
React_TODO-APP-ADD-AND-DELETE

TetianaMD:
[DEMO LINK](https://TetianaMD.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID, createTodos, getTodos } from './api/todos';
+import { USER_ID, createTodos, deleteTodo, getTodos } from './api/todos';
 import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
@@ -10,6 +10,7 @@ export const App: React.FC = () => {
   const [isTitle, setTitle] = useState('');
   const [loading, setLoading] = useState(false);
   const [filterStatus, setFilterStatus] = useState('all');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
 
   const filteredTodos = todos.filter(todo => {
     switch (filterStatus) {
@@ -54,6 +55,19 @@ export const App: React.FC = () => {
     });
   };
 
+  const deleteTodos = async (userId: number) => {
+    try {
+      setLoading(true);
+
+      await deleteTodo(userId);
+      setTodos(currentTodo => currentTodo.filter(todo => todo.id !== userId));
+    } catch (errors) {
+      setError('Unable to delete a todo');
+    } finally {
+      setLoading(false);
+    }
+  };
+
   // Відповідає за обробку події форми.
   const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
@@ -65,11 +79,23 @@ export const App: React.FC = () => {
 
     try {
       setLoading(true);
-      await addTodo({
+
+      // create a todo with id: 0
+      setTempTodo({
+        id: 0,
         title: isTitle.trim(),
         completed: false,
         userId: USER_ID,
       });
+      // викликаємо функцію addTodo, яка відправляє запит на сервер для створення нового todo
+      const newTodo = await addTodo({
+        title: isTitle.trim(),
+        completed: false,
+        userId: USER_ID,
+      });
+
+      setTodos(prevTodos => [...prevTodos, newTodo] as Todo[]); // додаємо нове todo до масиву todos
+      setTempTodo(null); // Сховати tempTodo
       setTitle('');
       handleRequest(); // Оновлюємо список завдань після додавання нового завдання
     } catch (errors) {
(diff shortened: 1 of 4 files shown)


andriy-shymkiv (changes requested):
There is some problems with UI, please pay attention to it:
1. when adding a new todo there is a 1sec glitch
2. when deleting an existing todo you shouldn't hide all of them for a second

Screen recording attached

https://github.com/mate-academy/react_todo-app-add-and-delete/assets/102289654/201195b1-d46b-4029-b172-e2a655d979ca

src/App.tsx
```diff
@@ -1,26 +1,222 @@
...
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
+  const [error, setError] = useState<string | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isTitle, setTitle] = useState('');
+  const [loading, setLoading] = useState(false);
+  const [filterStatus, setFilterStatus] = useState('all');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoadingTodo, setIsLoadingTodo] = useState<number[]>([]);
```
TarasHoliuk, line 18:
```suggestion
  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
```
since it's not a boolean variable but array of IDs let's rename it accordingly

src/App.tsx
```diff
@@ -1,26 +1,222 @@
...
+    }
+  });
+
+  useEffect(() => {
+    if (USER_ID) {
+      handleRequest();
+    }
+
+    setTimeout(() => {
+      setError(''); // Очищаємо помилку після схову повідомлення
```
TarasHoliuk, line 37: that's nice you wanted to add an explanation, but the code is quite self-explanatory. Comments in JS code are needed in case you want to explain difficult logic or there is something "unusual", exceptions, etc.

I'd recommend you to remove all such comments

src/App.tsx
```diff
@@ -1,26 +1,222 @@
...
+      setTodos(prevTodos => [...prevTodos, newTodo]);
+      setTitle('');
+    } catch (errors) {
+      setError('Unable to add a todo');
+    } finally {
+      setTempTodo(null); // Сховати tempTodo
+    }
+  };
+
+  const deleteTodos = async (userId: number) => {
```
TarasHoliuk, line 99:
```suggestion
  const deleteTodo = async (userId: number) => {
```
it deletes one todo
you can `import { deleteTodo as deleteTodoApi } from 'path/example'` to avoid variable name duplication

src/App.tsx
```diff
@@ -1,26 +1,222 @@
...
+    } finally {
+      // setLoading(false);
+    }
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const handleInputTodo = (value: string) => {
```
TarasHoliuk, line 168: `handleInputTodo` function wrapper is redundant - instead you can use  `setTitle` everywhere

src/App.tsx
```diff
@@ -1,26 +1,222 @@
...
+import { Todo } from './types/Todo';
+import { Header } from './components/Header';
+import { Main } from './components/Main';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
+  const [error, setError] = useState<string | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isTitle, setTitle] = useState('');
+  const [loading, setLoading] = useState(false);
```
TarasHoliuk, line 15: you can get rid of this state
`const isLoading = !!loadingTodoIds.length;`

image

TarasHoliuk (changes requested):
Hi! App is look and works way better now 👍 
Let's improve your code a bit :)

src/App.tsx
```diff
@@ -1,26 +1,219 @@
...
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filterStatus) {
+      case TodoStatus.Active:
+        return !todo.completed;
+      case TodoStatus.Completed:
+        return todo.completed;
+      default:
+        return true;
+    }
```
volodymyr-soltys97, line 34:
```suggestion
    switch (filterStatus) {
      case TodoStatus.Active:
        return !todo.completed;
        
      case TodoStatus.Completed:
        return todo.completed;
        
      default:
        return true;
    }
```

src/App.tsx
```diff
@@ -1,26 +1,219 @@
...
+      setError(null);
+    } catch (errors) {
+      setError('Unable to load todos');
+    }
+  };
+
+  const isLoading = !!loadingTodoIds.length;
+
+  const inputRef = useRef<HTMLInputElement>(null);
+  // додавання фокусу на input
```
volodymyr-soltys97, line 62: If you need comments in the code, they should be in English only. Fix it everywhere

src/App.tsx
```diff
@@ -1,26 +1,219 @@
...
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [todos]);
+
+  // Відповідає за обробку події форми.
+  const handleSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+    if (title.trim() === '') {
```
volodymyr-soltys97, line 72:
```suggestion
    if (!title.trim()) {
```

src/App.tsx
```diff
@@ -1,26 +1,219 @@
...
+            clearCompletedTodos={clearCompletedTodos}
+          />
+        )}
+      </div>
+
+      {/* DON'T use conditional rendering to hide the notification */}
+      {/* Add the 'hidden' class to hide the message smoothly */}
+      <div
+        data-cy="ErrorNotification"
+        className={`notification is-danger is-light has-text-weight-normal ${!isLoading && error ? '' : 'hidden'}`}
```
volodymyr-soltys97, line 206: Use the classnames library for add classes with condition everywhere

src/components/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${todosCounter} ${todosCounter === 1 ? 'item' : 'items'} left`}
+      </span>
+
+      {/* Active link should have the 'selected' class */}
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={`filter__link ${filterStatus === TodoStatus.All ? 'selected' : ''}`} // use the selected class to highlight a selected link;
```
volodymyr-soltys97, line 32: The same as above, fix it everywhere

src/components/Main.tsx
```diff
@@ -0,0 +1,72 @@
...
+            <input
+              data-cy="TodoStatus"
+              type="checkbox"
+              className="todo__status"
+              checked={todo.completed}
+              onChange={() => toggleTodoCompletion(todo.id)}
+            />
+          </label>
+          <span data-cy="TodoTitle" className="todo__title">
+            {todo.title}
```
volodymyr-soltys97, line 40: Use destruction for `todo`

volodymyr-soltys97 (changes requested):
Almost done!
Let's make your code better

l4st1m0za (commented):
Good job!

PR #1173, opened 2024-04-06
finished part2

julwer1k:
[DEMO](https://julwer1k.github.io/react_todo-app-add-and-delete/)

The only bug I can't fix on my own. When we delete multiple Todos at the same time. We click on the 3rd, 2nd, and wait. **Loader = true**, the 2nd Todo ID is added to ### ProcessTodoIds, then when one is deleted **Loader = false** and because of this, the style of the loader disappears on the Todo we clicked later. There may be a problem in changing the **ProcessTodoIds state** or **implementing the Loader = true/false logic** or with the implementation of **isActive - on line 24 of the code**

### DIFF
--- a/cypress/integration/page.spec.js
+++ b/cypress/integration/page.spec.js
@@ -38,14 +38,17 @@ const page = {
   mockLoad: (response = { fixture: 'todos' }) => {
     return cy.intercept('**/todos?userId=*', response);
   },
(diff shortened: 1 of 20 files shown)


etojeDenys (approved):
good job, feel free to ask for some help in the chat how to fix that bug

PR #1176, opened 2024-04-08
add solution

Dmitrel:
https://Dmitrel.github.io/react_todo-app-add-and-delete/


Oleksandr2108 (changes requested):
demo link does not work

PR #1190, opened 2024-04-17
solution

Miishca:
[DEMO LINK](https://miishca.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/component/Header.tsx
+++ b/src/component/Header.tsx
@@ -1,0 +1,77 @@
+import React, { useEffect, useRef, useState } from 'react';
+import { useTodos } from '../utils/TodoContext';
+import classNames from 'classnames';
+import { Errors } from '../types/ErrorsTodo';
+import { USER_ID } from '../api/todos';
+
+export const Header: React.FC = () => {
+  const { todos, addTodo, loading, setLoading, setErrorMessage } = useTodos();
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const [inputTodo, setInputTodo] = useState('');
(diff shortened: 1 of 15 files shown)


src/component/Header.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { useTodos } from '../utils/TodoContext';
+import classNames from 'classnames';
+import { Errors } from '../types/ErrorsTodo';
+import { USER_ID } from '../api/todos';
+
+export const Header: React.FC = () => {
+  const { todos, addTodo, loading, setLoading, setErrorMessage } = useTodos();
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const [inputTodo, setInputTodo] = useState('');
+  const [submitting, setSubmitting] = useState(false);
```
etojeDenys, line 11: looks like it's redundant and you can use the isLoading state

etojeDenys (changes requested):
feel free to ask for some help about that in the chat

src/component/Header.tsx
Miishca: I remove the submitting state. 
in code test passed, but after the deployed test 'should remove todos with success responses and keep todos with errors' did not pass.
Could you suggest where is the problem?

Miishca: Screenshot 2024-04-17 at 13 23 38

Miishca: now tests pass, but I remove setTodos(todos) and [todos].
Won't this cause errors later on?

IvanFesenko (approved):
Well done

PR #1191, opened 2024-04-17
solution

BodyaKutsyk:
[DEMO LINK](https://BodyaKutsyk.github.io/react_todo-app-add-and-delete/)


anastasiia-tilikina (changes requested):
Good job!
left some comments here

IvanFesenko (approved):
Pay attention not all comments were fixed. They should be fixed in next task

PR #1192, opened 2024-04-17
Develop

Patryk-Buczkowski:
[DEMO LINK](https://Patryk-Buczkowski.github.io/react_todo-app-loading-todos/)


Patryk-Buczkowski:
> [DEMO LINK](https://Patryk-Buczkowski.github.io/react_todo-app-loading-todos/)

    Adding a todo

      √ should focus text field by default

      if title is empty

        √ should not send a request

        √ should keep text field focused

        1) should display an error message => but is visible?!

BogdanMaliuta (changes requested):
Fix test with empty title error

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+
+    if (trimedTitle !== '') {
+      const newTodo: Todo = {
+        id: 0,
+        userId: 472,
+        title: trimedTitle,
+        completed: false,
+      };
+
+      const temp = {
```
yevhenii-pyl, line 124: Why do you need newTodo and temp? 
Looks like they are pretty much the same items

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
-  }
+  const [isLoading, setIsLoading] = useState<number | null>(null);
+  const [isEdited, setIsEdited] = useState<number | null>(null);
+  // const [todos, setTodos] = useState<Todo[] | []>([]);
+  const [editedTitle, setEditedTitle] = useState('');
+  const [newTitle, setNewTitle] = useState('');
+  const [errMessage, setErrMessage] = useState('');
+  const [stat, setStat] = useState(Status.all);
+  const [visibleErr, setVisibleErr] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo[] | null>(null);
```
yevhenii-pyl, line 21: No need in that extra state, it just complicates things here.

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+
+      try {
+        setIsLoading(0);
+
+        await postTodo(newTodo).then(respond => {
+          setTempTodo(null);
+          setTodos(prevTodos => [...prevTodos, respond]);
+        });
+
+        await getTodos().then(setTodos);
```
yevhenii-pyl, line 141: Bad practice! 
You never wants to refresh your list in that manner: having 2 API calls instead of 1 will quickly create performance issues. 

See, the logic here should be as simple as possible: 
user added a todo => you received a 200 response => you are 100% sure that the list is updated on the sever => no need to fetch it, you can change the UI already. 

Set a temp todo in order to make that loader work before success => replace it with the actual todo from response on success, that will pretty much do the job.

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+        newTodo = { ...todoToUpdate, completed: !updatedTodo.completed };
+      }
+
+      if (option === 'title' && todoToUpdate) {
+        newTodo = { ...todoToUpdate, title: editedTitle };
+      }
+
+      await patchTodo(newTodo);
+
+      await getTodos().then(setTodos);
```
yevhenii-pyl, line 85: Same here, you are sending unnecessary API calls to your backend. 
Backend dev will not be happy :)

yevhenii-pyl (changes requested):
You are getting close to it, but still we have to remove unnecessary API calls 💪

yevhenii-pyl (approved):
You did it, let's move forward :) 
However, to make your code even cleaner: 
1. Respect the folder structure, there should be the 'components' folder for all your components, except of App.tsx 
2. Don't forget to delete commented redundant code (you have that big chunk at ContextTodos.tsx)
3. Try not to create useless variables: you have newTodo and temp that are exactly the same at TodoContext.tsx).
4. Try not to disable eslint rules, almost always there is a better solution.

PR #1193, opened 2024-04-18
add task solution

Danylo-Chumakov:
[DEMO LINK](https://Danylo-Chumakov.github.io/react_todo-app-add-and-delete/)

I have a problem with 1 test (Adding a todo. if title is empty, should keep text field focused )
how to make it stay in focus?


sTorba24 (changes requested):
Good work!

Generally the problem is that i'm not able to focus on input absolutely after getting error, when i try to add empty todo, fix it please

If you have any issues with it - feel free to text your fe_chat

sTorba24 (approved):

PR #1201, opened 2024-04-22
solution

Anothar:
[DEMO LINK](https://Anothar.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/component/Footer/Footer.tsx
+++ b/src/component/Footer/Footer.tsx
@@ -1,0 +1,98 @@
+import { useContext } from 'react';
+import { DispatchContext, StateContext } from '../../store/store';
+import cn from 'classnames';
+import { IsUseTodos } from '../../types/IsUseTodos';
+import { deleteTodo } from '../../api/todos';
+
+export const Footer = () => {
+  const dispatch = useContext(DispatchContext);
+  const { todos, useTodos } = useContext(StateContext);
+
+  const todosLeft = todos.filter(todo => !todo.completed).length;
+
+  const handleDeleteCompletedTodo = () => {
+    const allCompleteId = todos
+      .filter(todo => todo.completed)
+      .map(todo => todo.id);
+
+    dispatch({ type: 'setVaitTodoId', id: allCompleteId });
+
+    for (const id of allCompleteId) {
+      deleteTodo(id)
+        .then(() => {
+          dispatch({ type: 'removeTodo', id: id });
+        })
+        .catch(() => {
+          dispatch({ type: 'setError', error: 'Unable to delete a todo' });
+        })
+        .finally(() => {
+          dispatch({ type: 'deleteVaitTodoId', id: id });
+        });
+    }
+  };
+
+  return (
+    <>
+      {!!todos.length && (
+        <footer className="todoapp__footer" data-cy="Footer">
+          <span className="todo-count" data-cy="TodosCounter">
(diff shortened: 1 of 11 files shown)


src/component/Footer/Footer.tsx
```diff
@@ -0,0 +1,98 @@
...
+              Completed
+            </a>
+          </nav>
+
+          {/* this button should be disabled if there are no completed todos */}
+          <button
+            type="button"
+            className="todoapp__clear-completed"
+            data-cy="ClearCompletedButton"
+            onClick={() => handleDeleteCompletedTodo()}
```
lerastarynets, line 89:
```suggestion
            onClick={handleDeleteCompletedTodo}
```

src/component/Header/Header.tsx
```diff
@@ -0,0 +1,87 @@
...
+      inputRef.current?.focus();
+    } else {
+      inputRef.current?.blur();
+    }
+  }, [focusNewTodo]);
+
+  return (
+    <header className="todoapp__header">
+      {/* this button should have `active` class only if all todos are completed */}
+      {todos.length > 0 && (
```
lerastarynets, line 57:
```suggestion
      {!!todos.length && (
```

src/component/Header/Header.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+      {/* Add a todo on form submit */}
+      <form onSubmit={addNewTodo}>
+        <input
+          ref={inputRef}
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          disabled={idTodoSubmitting !== 0}
```
lerastarynets, line 78:
```suggestion
          disabled={!!idTodoSubmitting}
```

src/component/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,128 @@
...
+              {title}
+            </span>
+          )}
+
+          {id === changerId && (
+            <form
+              onSubmit={e => {
+                e.preventDefault();
+                dispatch({ type: 'setChangedTodoId', id: 0 });
+              }}
```
lerastarynets, line 76: Pls move the function inside `onSubmit` handler to a separate const and use as a callback here
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/component/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,128 @@
...
+                type="text"
+                className="todo__title-field"
+                placeholder="Empty todo will be deleted"
+                value={title}
+                onKeyUp={e => {
+                  if (e.key === 'Escape') {
+                    dispatch({ type: 'escapeChangedText', id: id });
+                    dispatch({ type: 'setChangedTodoId', id: 0 });
+                  }
+                }}
```
lerastarynets, line 89: Pls move the function inside `onKeyUp` handler to a separate const and use as a callback here
If a mentor points out a mistake, it means that you need to fix ALL occurrences of the mistake, not only in the place where the comment was left.

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #1202, opened 2024-04-24
Implement deleting todos functionality

olafchuszno:
[DEMO LINK](https://olafchuszno.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Add a todo with the entered title on the form submit:
 
 Remove a todo on the `TodoDeleteButton` click:
 
-- covered the todo with the loader while waiting for the API response;
+- cover the todo with the loader while waiting for the API response;
 - remove the todo from the list on success;
 - in case of API error show `Unable to delete a todo` notification at the bottom (the todo must stay in the list);
 


src/App.tsx
```diff
@@ -1,26 +1,181 @@
...
+    newTodoInput.current?.focus();
+  }, [todos, newTodoInput]);
+
+  useEffect(() => {
+    if (deleteTodoId) {
+      setCurrentError(null);
+
+      deleteTodo(deleteTodoId)
+        .then(() => {
+          getTodos()
```
yevhenii-pyl, line 59: This API call is pretty much redundant: you never want to send an API call to 'get the whole list of items once again'  after delete/put/post request. You backend will be overloaded and users with slow connection will be irritated. 

If it was successful, knowing the item properties, you can easily update your UI. 

Also, notice how quickly your code becomes flooded with .catch statements. Not cool.

src/App.tsx
```diff
@@ -1,26 +1,181 @@
...
+            .catch(() => {
+              setCurrentError(Error.CannotLoad);
+            });
+        })
+        .catch(() => {
+          setCurrentError(Error.CannotDelete);
+        })
+        .finally(() => setDeleteTodoId(null));
+    }
+  }, [deleteTodoId]);
```
yevhenii-pyl, line 70: This is strange: why do you want to use useEffect here at all? 
Can't you just create a function for deleting the item? 
The looks overcomplicated: you have a useState to set the id for deletion and create a useEffect hook that is triggered whenever the id for deletion is changed. It should never be like that, all you need here is a simple async function to delete an item.

src/App.tsx
```diff
@@ -1,26 +1,181 @@
...
+  const createTodo = useCallback(() => {
+    const todoTitle = addTodoTitle?.trim();
+
+    if (!todoTitle.length) {
+      setCurrentError(Error.EmptyTitle);
+
+      return;
+    }
+
+    setTempTodo({
```
yevhenii-pyl, line 96: The idea of having a tempTodo is working, but don't you find it overcomplicated? 
Like, you have a separate state for it and exception in your jsx to display that one item. 
And all the effort - just to display an item with loader. 

Wouldn't it be easier to add a temporary item in the list before the callback, like:

` setTodos(current => [...current, tempTodo])`

trigger the API call
then, on success you can remove your temporary item with the real one(received from the server)
or, on fail, catch the error and remove temp item from the list.

src/App.tsx
```diff
@@ -1,26 +1,181 @@
...
+      <div className="todoapp__content">
+        <TodoHeader
+          handleChangeTitle={setAddTodoTitle}
+          addTodoTitle={addTodoTitle}
+          createTodo={createTodo}
+          newTodoInput={newTodoInput}
+          isNewTodoLoading={isNewTodoLoading}
+        />
+
+        {todos.length > 0 && (
```
yevhenii-pyl, line 149: `!!todos?.length ` if you want to go a bit fancy

src/App.tsx
```diff
@@ -1,26 +1,181 @@
...
+          <TodoList
+            todos={getFilteredTodos(todos, currentFilter)}
+            handleDeleteTodo={deleteTodoById}
+          />
+        )}
+
+        {tempTodo && (
```
yevhenii-pyl, line 156: If you use my advice on how to work with tempTodo => this chunk can be completely removed

src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 467;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
```
yevhenii-pyl, line 7: Let's extract that `/todos` into a const, like BASE_TODOS_PATH or something

src/App.tsx
olafchuszno: In the readme it is specified that there has to be state for the tempTodo and to NOT add it to the todos array.
However if you want of course I can try to do it this way:)

src/App.tsx
```diff
@@ -1,26 +1,158 @@
...
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [currentError, setCurrentError] = useState<Error | null>(null);
+  const [currentFilter, setCurrentFilter] = useState<Filter>(Filter.All);
+  const [addTodoTitle, setAddTodoTitle] = useState<string>('');
```
BogdanMaliuta, line 18: this naming looks like func, maybe better name for this is newTodoTitle

src/App.tsx
```diff
@@ -1,26 +1,158 @@
...
+    if (currentError) {
+      newTodoInput.current?.focus();
+
+      setTimeout(() => setCurrentError(null), 3000);
+    }
+  }, [currentError]);
+
+  useEffect(() => {
+    newTodoInput.current?.focus();
+  }, [todos, newTodoInput]);
```
BogdanMaliuta, line 50: cant understand this useEffect, it means if list of todos changes, or newTodoInput changes, focus input? I think here smthng wrong in logic

BogdanMaliuta (approved):
Nice work, code works good.But sometimes its hard to understand, because of naming

PR #1203, opened 2024-04-24
2 part

noraaudonnet:
[DEMO LINK](https://noraaudonnet.github.io/react_todo-app-add-and-delete/)

That's the first try. I know it doesn't pass all of the tests but I'm a bit confused in terms of my next moves. 
I don't understand why useRef doesn't work so it focuses the input by default.

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -95,16 +95,15 @@ export const Footer: React.FC<Props> = ({
           {Status.Completed}
         </a>
       </nav>
-      {!!complitedTodosCount && (
-        <button
-          type="button"
-          className="todoapp__clear-completed"
-          data-cy="ClearCompletedButton"
-          onClick={handleClearButton}
-        >
-          Clear completed
-        </button>
-      )}
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        onClick={handleClearButton}
+        disabled={!complitedTodosCount}
+      >
+        Clear completed
+      </button>
     </footer>
   );
 };


TarasHoliuk:
The filter should always be centered:
image

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,100 @@
...
+  setTodoTitle,
+  setErrorMessage,
+  loading,
+  setLoading,
+}) => {
+  const titleField = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (titleField.current) {
+      titleField.current.focus();
```
TarasHoliuk, line 29: it doesn't work (and tests also fails because of it) - try to debug and find the reason. Ask for help in the chat if you need

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+    return todosService
+      .deleteTodo(todoId)
+      .then(() => {})
+      .catch(error => {
+        setErrorMessage(`Unable to delete a todo`);
+        throw error;
+      });
+  }
+
+  const handleDelete = (todoId: number) => {
```
TarasHoliuk, line 44: redundant function wrapper - just use `removeTodo`

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+        className="todo__remove"
+        data-cy="TodoDelete"
+        onClick={() => handleDelete(id)}
+      >
+        ×
+      </button>
+
+      <div data-cy="TodoLoader" className="modal overlay">
+        <div className="modal-background has-background-white-ter" />
+        <div className="loader" />
```
TarasHoliuk, line 79: you have to show loader when deleting an item. Note that on click `clear completed` you should show the loader for many items. To do it, you can create `loadingTodoIds` array state and push the id every time you possess some action with this todo. Finally, remove this id from the array.
If you need more help feel free to ask in the chat

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <section className="todoapp__main" data-cy="TodoList">
+      {filteredTodos.map(todo => (
+        <TodoItem
+          todo={todo}
+          key={todo.id}
+          todos={todos}
+          setTodos={setTodos}
+          setErrorMessage={setErrorMessage}
+        />
+      ))}
```
TarasHoliuk, line 42: you should also display `tempTodo` here with loader. Read task description for more info

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  function removeTodo(todoId: number) {
+    setTodos(currentTodos => currentTodos.filter(todo => todo.id !== todoId));
+
+    return todosService
+      .deleteTodo(todoId)
+      .then(() => {})
+      .catch(error => {
+        setErrorMessage(`Unable to delete a todo`);
+        throw error;
```
TarasHoliuk, line 40: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    setTodos(updatedTodos);
+  };
+
+  function removeTodo(todoId: number) {
+    setTodos(currentTodos => currentTodos.filter(todo => todo.id !== todoId));
+
+    return todosService
+      .deleteTodo(todoId)
+      .then(() => {})
```
TarasHoliuk, line 37: only after successful answer from server change the state

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }, []);
+  const allTodosCompleted = todos.every(todo => todo.completed);
```
TarasHoliuk, line 32:
```suggestion
  const completed Todos= todos.every(todo => todo.completed);
```

TarasHoliuk (changes requested):
GJ 👍 
Let's improve it a bit (input focus, loader). Pay attention to tests - try to fix them

etojeDenys (changes requested):
1. fix this mismatch
image
image

anastasiiavorobiova (approved):
Great work!

PR #1205, opened 2024-04-30
Solution

Davwidos:
[DEMO LINK](https://davwidos.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/providers/TodosProvider.tsx
+++ b/src/providers/TodosProvider.tsx
@@ -1,0 +1,145 @@
+import {
+  FC,
+  PropsWithChildren,
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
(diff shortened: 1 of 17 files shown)


src/providers/TodosProvider.tsx
```diff
@@ -0,0 +1,145 @@
...
+        return activeTodos;
+      case 'all':
+      default:
+        return todos;
+    }
+  };
+
+  const onAddTodo = useCallback((title: string) => {
+    if (!title) {
+      setError('empty');
```
nataliaklonowska, line 98: define all types of errors

nataliaklonowska (changes requested):
when API error occurs during todo deletion you have endless loader

nataliaklonowska (approved):

PR #1206, opened 2024-04-30
Develop2

ArtemZakhar:
Hi)

Have a nice day!)

[DEMO LINK](https://ArtemZakhar.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,56 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useContext } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { USER_ID, getTodos } from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { DispatchContext } from './store/todoReducer';
+import { Action } from './types/actions';
+import { ErrorMessage } from './components/ErrorMessage';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [errorMessage, setErrorMessage] = useState('');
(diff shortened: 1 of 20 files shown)


ArtemZakhar:
I finally made it!) 
I have made the opportunity to pass only partial data to the server instead of full object) 
But I had problems with merging branches.
And so I am afraid to move something)

Will show all changes at the next stage and visit qna to cover the merging issue)

src/App.tsx
```diff
@@ -1,26 +1,56 @@
...
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { DispatchContext } from './store/todoReducer';
+import { Action } from './types/actions';
+import { ErrorMessage } from './components/ErrorMessage';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [errorMessage, setErrorMessage] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
```
alexander-ignatow, line 16: Not an issue at all - but possibly you'll find more usable `undefined`, not `null`

```suggestion
  const [tempTodo, setTempTodo] = useState<Todo>();
```
`tempTodo` is Todo | undefined

IMO this looks slightly better, though some prefer using `null`s

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (data) {
+    // We add body and Content-Type only for the requests with data
+    options.body = JSON.stringify(data);
+    options.headers = {
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // DON'T change the delay it is required for tests
+  return wait(100)
```
alexander-ignatow, line 30: In the real world you don't want this `wait` in production, I guess :)

alexander-ignatow (approved):
Very good
Clean project structure  👍 

And your "poorman's redux" is cool too :)

src/App.tsx
ArtemZakhar: I suppose previously I was required using null))

PR #1213, opened 2024-05-08
React todo app add and delete

itsXenthis:
[DEMO](https://itsXenthis.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/utils/fetchClient.ts
+++ b/src/utils/fetchClient.ts
@@ -12,7 +12,7 @@ type RequestMethod = 'GET' | 'POST' | 'PATCH' | 'DELETE';
 function request<T>(
   url: string,
   method: RequestMethod = 'GET',
-  data: any = null, // we can send any data to the server
+  data: any = null,
 ): Promise<T> {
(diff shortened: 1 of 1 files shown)


src/components/TodoApp.tsx
```diff
@@ -0,0 +1,120 @@
...
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage(ErrorMessage.load);
+      });
+  }, [setTodos, setErrorMessage]);
+
+  useEffect(() => {
+    setErrorMessage(ErrorMessage.noError);
+  }, [todos, setErrorMessage]);
```
nataliaklonowska, line 56: why do you need this useEffect?

nataliaklonowska (changes requested):
fix errors in tests

nataliaklonowska (changes requested):
in your contexts you only create states but you can also move there functions that change those states

nataliaklonowska (approved):

PR #1214, opened 2024-05-08
Solution proposed

bartKarb:
[DEMO LINK](https://bartKarb.github.io/react_todo-app-add-and-delete/)


nataliaklonowska (changes requested):
when you have error that prevents deleting todo this todo stays in loading state
your notifications errors (empty title) can't handle several errors in a row

nataliaklonowska (changes requested):
when handling multiple todo deletions only first one really delete others stay in loading state and after refresh they still in todo list

nataliaklonowska (approved):

PR #1215, opened 2024-05-08
solution part2

etuzimek:
[DEMO LINK](https://e.tuzimek.github.io/react_todo-app-add-and-delete/)


nataliaklonowska (changes requested):
you should be able to handle multiple deletions in the same time

nataliaklonowska (changes requested):
you should be able to delete multiple todos at once. Now you are deleting them but loader stays only on one of them not all

nataliaklonowska (approved):

PR #1217, opened 2024-05-08
solution

weswojciech:
https://weswojciech.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/components/TodoItem.tsx
+++ b/src/components/TodoItem.tsx
@@ -1,0 +1,84 @@
+import { FC, useState } from 'react';
+import { Todo } from '../types/Todo';
+import React from 'react';
+
+type Props = {
+  todo: Todo;
+  deleteTodo: (todoId: number) => Promise<boolean>;
+  isTemp?: boolean;
+  // handleCompleteTodo: (todoId: number) => void;
+};
+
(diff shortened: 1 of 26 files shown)


nataliaklonowska (changes requested):
after todo addition fails you should focus on text field (input)

src/components/TodoItem.tsx
```diff
@@ -21,8 +21,10 @@ export const TodoItem: FC<Props> = ({
   const handleClick = () => {
     setIsBeingDeleted(true);
 
-    deleteTodo(todo.id).catch(() => {
-      setIsBeingDeleted(false);
+    deleteTodo(todo.id).then(didSucceed => {
+      if (!didSucceed) {
+        setIsBeingDeleted(false);
+      }
```
nataliaklonowska, line 27: deleteTodo don't return promise so you can't use than, catch or finally (these methods are called on a promise to handle the resolution or rejection of the promise. If the deleteTodo function returns a value other than a promise, you won't be able to handle it using promise-related methods)

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #1220, opened 2024-05-10
Develop

marinatea:
[DEMO LINK](https://marinatea.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,258 @@
...
+        errorAll = true;
+      }
+    }
+
+    if (errorAll) {
+      setError(true);
+      setErrorType('delete');
+
+      return;
+    }
```
nataliaklonowska, line 128: if any of todos fail you have return and don't filter any of them in setTodos (thats why test fail)

nataliaklonowska (changes requested):

PR #1224, opened 2024-05-13
second part of todo_app

iamgranovskaya:
[DEMO](https://iamgranovskaya.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,0 +1,87 @@
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import { ActionNames, Errors, TodoContext } from './TodoContext';
+import { Todo } from '../types/Todo';
+import { USER_ID, createTodo } from '../api/todos';
+
+export const TEMP_USER_ID = -1;
+
+export const Header: React.FC = () => {
+  const { todos, dispatch, handleError } = useContext(TodoContext);
+  const [value, setValue] = useState('');
+  const [disabled, setDisabled] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    event.preventDefault();
+    setValue(event.target.value);
+  };
+
+  const handleKeyDown = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (value.trim() === '') {
+      handleError(Errors.EmptyTitle);
+
+      return;
+    }
+
+    const newTodo: Todo = {
+      id: TEMP_USER_ID,
+      userId: USER_ID,
+      completed: false,
+      title: value.trim(),
+    };
+
+    setDisabled(true);
+
+    dispatch({ type: ActionNames.Add, payload: newTodo });
+
+    createTodo(newTodo)
+      .then(nt => {
+        dispatch({ type: ActionNames.Delete, payload: newTodo.id });
+        dispatch({ type: ActionNames.Add, payload: nt });
+        setValue('');
+      })
+      .catch(() => {
+        dispatch({ type: ActionNames.Delete, payload: newTodo.id });
+        handleError(Errors.AddTodo);
+      })
+      .finally(() => setDisabled(false));
+  };
+
+  const focusInput = () => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  };
+
+  useEffect(() => {
+    focusInput();
+  }, [todos]);
+
+  return (
+    <header className="todoapp__header">
+      {/* this button should have `active` class only if all todos are completed */}
+      <button
+        type="button"
+        className="todoapp__toggle-all"
+        data-cy="ToggleAllButton"
+      />
+
+      {/* Add a todo on form submit */}
+      <form onSubmit={handleKeyDown}>
+        <input
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          autoFocus
+          value={value}
+          ref={inputRef}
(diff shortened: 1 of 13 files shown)


src/components/Header.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const TEMP_USER_ID = -1;
+
+export const Header: React.FC = () => {
+  const { todos, dispatch, handleError } = useContext(TodoContext);
+  const [value, setValue] = useState('');
+  const [disabled, setDisabled] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
TarasHoliuk, line 14:
```suggestion
  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```

src/components/Header.tsx
```diff
@@ -0,0 +1,87 @@
...
+    event.preventDefault();
+    setValue(event.target.value);
+  };
+
+  const handleKeyDown = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (value.trim() === '') {
```
TarasHoliuk, line 22:
```suggestion
    if (!value.trim()) {
```

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,169 @@
...
+    setDeleteCandidates(prev => [...prev, ...ids]);
+  };
+
+  const value = useMemo(
+    () => ({
+      todos: filter(todos, filteredBy),
+      originalTodos: todos,
+      dispatch,
+      filteredBy,
+      handleFilterBy,
```
TarasHoliuk, line 137:
```suggestion
      setFilteredBy,
```

`handleFilterBy` function wrapper is redundant - you can delete it

src/components/Header.tsx
```diff
@@ -0,0 +1,87 @@
...
+      title: value.trim(),
+    };
+
+    setDisabled(true);
+
+    dispatch({ type: ActionNames.Add, payload: newTodo });
+
+    createTodo(newTodo)
+      .then(nt => {
+        dispatch({ type: ActionNames.Delete, payload: newTodo.id });
```
TarasHoliuk, line 41: It works, but let's do it as described in the task (with `tempTodo`) instead of adding and deleting this temporary todo in todos array

src/components/Header.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import { ActionNames, Errors, TodoContext } from './TodoContext';
+import { Todo } from '../types/Todo';
+import { USER_ID, createTodo } from '../api/todos';
+
+export const TEMP_USER_ID = -1;
```
TarasHoliuk, line 6:
```suggestion
export const TEMP_TODO_ID = 0;
```

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const handleError = (message: string) => {
+    setError(message);
+    setTimeout(() => setError(null), 3000);
+  };
+
+  const handleFilterBy = useCallback((type: FilterBy) => {
+    setFilteredBy(type);
+  }, []);
+
+  const handleDeleteCandidates = (ids: number[]) => {
```
TarasHoliuk, line 127:
```suggestion
  const handleDeleteTodos = (ids: number[]) => {
```
or `handleClearCompleted`

Instead of se

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,169 @@
...
+  Load = 'LOAD',
+  Add = 'ADD',
+  Delete = 'DELETE',
+  Update = 'UPDATE',
+  ToggleCompleted = 'TOGGLE_COMPLETED',
+  ToggleAllCompleted = 'TOGGLE_ALL_COMPLETED',
+  ClearCompleted = 'CLEAR_COMPLETED',
+}
+
+type T = {
```
TarasHoliuk, line 57: why is this action type declared separately? Anyway, if you need to keep it separately it should have meaningful name

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,169 @@
...
+export type Action =
+  | { type: ActionNames.Load; payload: Todo[] }
+  | { type: ActionNames.Add; payload: Todo }
+  | { type: ActionNames.Delete; payload: number }
+  | { type: ActionNames.Update; payload: { id: number; title: string } }
+  | T
+  | { type: ActionNames.ToggleAllCompleted; payload: Todo[] }
+  | { type: ActionNames.ClearCompleted };
+
+export const Errors = {
```
TarasHoliuk, line 71:
```suggestion
export const errors = {
```
it's just object, not a type (enum)

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,169 @@
...
+      error,
+      handleError,
+      deleteCandidates,
+      handleDeleteCandidates,
+    }),
+    [todos, dispatch, filteredBy, handleFilterBy, error, deleteCandidates],
+  );
+
+  useEffect(() => {
+    if (deleteCandidates.length) {
```
TarasHoliuk, line 147: interesting approach)
but in the next task you will need to change it 😌

TarasHoliuk (changes requested):
Nicely done 👍 

Added a few small comments to make your code better 🙂

src/components/TodoContext.tsx
iamgranovskaya: I'll fix it in the next task)

anastasiiavorobiova (approved):
Great work!

PR #1231, opened 2024-05-14
Developbr

Allanpt:
[DEMO LINK](https://Allanpt.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import {
   USER_ID,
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,26 +1,304 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectSort, setSelectSort] = useState<SortSelectOptions>('all');
+  const [error, setError] = useState<ErrorMessages>('');
+  const [inputTodo, setInputTodo] = useState('');
+  const [loading, setLoading] = useState(false);
+  const inputRefAddTodo = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    inputRefAddTodo.current?.focus();
+  }, [loading]);
```
EdPirro, line 34: If `loading` is present in the dependency array it should be used!
The `ref` props accepts a callback, maybe this behavior should be moved there:
```js
<Component
  ref={element => {
    if(element) element.focus;
  }}
/>
```

https://react.dev/reference/react-dom/components/common#ref-callback

src/App.tsx
```diff
@@ -1,26 +1,304 @@
...
+    setError('');
+    setLoading(true);
+    todos
+      .filter(el => el.completed)
+      .forEach(todoCompleted => {
+        deleteTodo(todoCompleted.id)
+          .then(() => {
+            deletedID.push(todoCompleted.id);
+
+            setTodos(tds => tds.filter(el => !deletedID.includes(el.id)));
```
EdPirro, line 185: TIP: When using a callback inside the `setState`, `tds` will already have the updated value, meaning every other item in `deletedID` will already have been deleted; take care of deleting only the current item:
`tds => tds.filter(el => el.id !== todoCompleted.id)`

EdPirro (changes requested):
Great job, @Allanpt! You're almost there! 👏
I pointed out a few improvements which I'll ask you to implement before continuing to the next part;

Also, you have several functions that depend on reactive values. Remember to wrap them in `useCallback`s or `useMemo`s to avoid unnecessary re-renders

src/App.tsx
Allanpt: great

Allanpt:
I've tried to implement useCallback in useful places, Later, i'll want a best approach to implement it.

src/App.tsx
```diff
@@ -1,26 +1,302 @@
...
+          type="button"
+          className="delete"
+          onClick={() => setError('')}
+        />
+        {/* show only one message at a time */}
+        {error === 'Unable to load todos' && 'Unable to load todos'}
+        {error === 'Title should not be empty' && 'Title should not be empty'}
+        {error === 'Unable to delete a todo' && 'Unable to delete a todo'}
+        {error === 'Unable to add a todo' && 'Unable to add a todo'}
+        {error === 'Unable to update a todo' && 'Unable to update a todo'}
```
jv-aquino, line 298: like this here, this string comparation will probably slow down your code

PR #1233, opened 2024-05-15
Develop

ordila:
Hi ! 
I would really thankful if you will find a problem because i can't , i tried also in Q&A session  - not successful ;

[DEMO LINK](https://ordila.github.io/react_todo-app-add-and-delete/)


metrakonon (approved):
Interesting solution! :+1:

mrkovaliv (changes requested):
Good work 👍 

Added some improvements)))

PR #1247, opened 2024-05-22
add solution

D-Kondratiev:
[DEMO LINK](https://D-Kondratiev.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -116,7 +116,7 @@ export const App: React.FC = () => {
           inputRef={inputRef}
         />
 
-        {todos.length > 0 && (
+        {!!todos.length && (
           <>
             <TodoList
               todos={visibleTodos}
--- a/src/ErrorNotification.tsx
+++ b/src/ErrorNotification.tsx
@@ -18,6 +18,7 @@ export const ErrorNotification: React.FC<Props> = ({
     const timerID = setTimeout(() => onDeleteError(), 3000);
 
     return () => clearTimeout(timerID);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [errorMessage]);
 
   return (
--- a/src/Header.tsx
+++ b/src/Header.tsx
@@ -9,6 +9,7 @@ interface Props {
   titleText: string;
   onType: (event: React.ChangeEvent<HTMLInputElement>) => void;
 }
+
 export const Header: React.FC<Props> = ({
   inputRef,
   onSubmit,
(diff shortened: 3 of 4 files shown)


luyx0 (approved):
Good Job)

Maybe in Header don't pass it to tempTodo and just take that inputRef to App

src/utils/getFilteredTodos.ts
```diff
@@ -0,0 +1,24 @@
+import { Errors } from '../types/Errors';
+import { FilterBy } from '../types/FiilterBy';
+import { Todo } from '../types/Todo';
+
+type FilterTheTodos = (todos: Todo[], filterBy: FilterBy) => Todo[];
+
+export const getFilteredTodos: FilterTheTodos = (todos, filterBy) => {
+  let filteredTodos = todos;
```
DanilWeda, line 8: You don't need let in this case. You can use const and then return action with the array.

src/Header.tsx
```diff
@@ -0,0 +1,46 @@
...
+  inputRef,
+  onSubmit,
+  tempTodo,
+  isDisabled,
+  titleText,
+  onType,
+}) => {
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, [tempTodo]);
```
DanilWeda, line 22: disable this rule for this case

src/ErrorNotification.tsx
```diff
@@ -0,0 +1,39 @@
...
+}) => {
+  useEffect(() => {
+    if (!errorMessage) {
+      return;
+    }
+
+    const timerID = setTimeout(() => onDeleteError(), 3000);
+
+    return () => clearTimeout(timerID);
+  }, [errorMessage]);
```
DanilWeda, line 21: disable eslint rule for this case

src/App.tsx
```diff
@@ -1,26 +1,145 @@
...
+        <Header
+          tempTodo={tempTodo}
+          isDisabled={isDisabled}
+          titleText={titleText}
+          onSubmit={handleSubmit}
+          onType={handleTyping}
+          inputRef={inputRef}
+        />
+
+        {todos.length > 0 && (
```
DanilWeda, line 119:
```suggestion
        {!!todos.length  && (
```
as alternative

DanilWeda (changes requested):
Good job. 
Check some small comments. 
Thanks)

src/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { useEffect } from 'react';
+import cn from 'classnames';
+
+interface Props {
```
undeletable, line 4: [Not critical]
Here and in the similar places for other components:
The better name for this interface would be something like `ErrorNotificationProps`, so that it precisely describes its contents and cannot be confused with other types.

undeletable (approved):
Looks good, I just have a small advice on improving your code.

PR #1255, opened 2024-05-23
solution

denyssheremeta:
[DEMO LINK](https://denyssheremeta.github.io/react_todo-app-add-and-delete/)
extract functionality to custom hooks coming soon =)


sevenlays (approved):
Really Good Job!:)

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,72 @@
...
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={cn('filter__link', {
+            selected: status === 'all',
+          })}
+          data-cy="FilterLinkAll"
+          onClick={() => setStatus('all')}
```
DanilWeda, line 32: all statuses need to move in constants or Enum ( best way Enum), then use in all places in your application.

src/utils/getFilteredTodos.ts
```diff
@@ -0,0 +1,18 @@
+import { Statuses } from '../types/Statuses';
+import { Todo } from '../types/Todo';
+
+interface Filter {
+  status: Statuses;
+}
+
+export const getFilteredTodos = (todos: Todo[], { status }: Filter) => {
+  let filteredTodos = [...todos];
```
DanilWeda, line 9: In this case, you don't need let . You can use const and then return operations with the array.

DanilWeda (changes requested):
Good job! 
Check some small comments! 
Thanks)

PR #1257, opened 2024-05-23
Added 2/3 part of todos App

PrytulaAlexandr:
[DEMO LINK](https://PrytulaAlexandr.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/utils/FilteredTodos.tsx
+++ b/src/utils/FilteredTodos.tsx
@@ -1,0 +1,22 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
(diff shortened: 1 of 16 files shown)


src/utils/FilteredTodos.tsx
```diff
@@ -0,0 +1,22 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
+export const FilteredTodos = (todos: Todo[], filterBy: Status): Todo[] => {
+  let filteredTodos = [...todos];
```
DanilWeda, line 5: In this case, you don't need let . You can use const and then return operations with the array.

DanilWeda (approved):
Great. 
Approved with a small comment) 
Thanks)

PR #1259, opened 2024-05-23
add task solution

AgnieszkaWw:
(https://AgnieszkaWw.github.io/react_todo-app-add-and-delete/) and add it to the PR description.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,170 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+import React, { useEffect, useMemo, useState } from 'react';
+import { UserWarning } from './UserWarning';
+import {
+  USER_ID,
+  getTodos,
+  createTodo,
+  deleteTodo,
+  updateToDo,
+} from './api/todos'; // Ensure createTodo is imported
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import { Header } from './components/Header/Header';
+import { ToDoList } from './components/ToDoList/ToDoList';
+import { Error } from './components/Error/Error';
+import { Footer } from './components/Footer/Footer';
+function getVisibleToDos(newTodos: Todo[], newStatus: Status) {
+  switch (newStatus) {
+    case Status.Active:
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,170 @@
...
+
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  USER_ID,
+  getTodos,
+  createTodo,
+  deleteTodo,
+  updateToDo,
+} from './api/todos'; // Ensure createTodo is imported
```
BogdanMaliuta, line 12: remove comments from code

src/App.tsx
```diff
@@ -1,26 +1,170 @@
...
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import { Header } from './components/Header/Header';
+import { ToDoList } from './components/ToDoList/ToDoList';
+import { Error } from './components/Error/Error';
+import { Footer } from './components/Footer/Footer';
+function getVisibleToDos(newTodos: Todo[], newStatus: Status) {
```
BogdanMaliuta, line 19: add empty line here for better code style

src/App.tsx
```diff
@@ -1,26 +1,170 @@
...
+
+  useEffect(() => {
+    let timeoutId: NodeJS.Timeout;
+
+    if (error) {
+      timeoutId = setTimeout(() => setError(''), 3000);
+    }
+
+    return () => clearTimeout(timeoutId);
+  }, [error]);
```
BogdanMaliuta, line 56: you dont need this use effect, you can remove error on finally when geting todos

src/App.tsx
```diff
@@ -1,26 +1,170 @@
...
+      .then(resultTodo => {
+        setTodos([...todos, resultTodo]);
+        setTitle('');
+      })
+      .catch(() => setError('Unable to add a todo'))
+      .finally(() => {
+        setTempTodo(null);
+        setIsLoading(false);
+      });
+  };
```
BogdanMaliuta, line 88: you are using it only in header so i think thus func should be in header, also your another functions needs to be in another files

BogdanMaliuta (changes requested):
Hello Agnieszka.
Here is some things to fix:
fix styles on editing todo
image
chage in some places logic of your code, for example using second useEffect for remove error after 3 seconds its to much, you can do this in your finally
lift state up only when you need this, dont save all methods and state in your app.tsx

PR #1265, opened 2024-05-24
Solution

OTsiolkovskiy:
[DEMO LINK](https://OTsiolkovskiy.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/utils/filteredTodos.tsx
+++ b/src/utils/filteredTodos.tsx
@@ -1,0 +1,22 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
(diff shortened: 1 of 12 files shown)


src/utils/filteredTodos.tsx
```diff
@@ -0,0 +1,22 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
+export const FilteredTodos = (todos: Todo[], filterBy: Status): Todo[] => {
+  let filteredTodos = [...todos];
```
DanilWeda, line 5: You don't need let in this case. You can use const and return operations with an array.

DanilWeda (approved):
Good job. 
Approved with a small comments! 
Thanks)

PR #1266, opened 2024-05-24
Solution

ivankovbohdan:
[DEMO LINK](https://ivankovbohdan.github.io/react_todo-app-add-and-delete/)


lerastarynets (changes requested):
Your todo bar disappears for a few seconds, if user click on delete item, pls fix

SanyaBratashchuk (approved):
tempTodo should have the same title as todo that is adding 
image

PR #1268, opened 2024-05-26
aksdhj

yurii-filiushkin:


src/App.tsx
```diff
@@ -1,26 +1,257 @@
...
+
+    const timeout = setTimeout(() => setErrorMessage(''), 3000);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [errorMessage]);
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
```
kinqbert, line 40: [Non critical] You can use reducer to create both active and completed todos lists in one iteration through an array

kinqbert (changes requested):
Hey! You did a good job, but still I have to reject your solution as tests are not fully passing (there must be 96 tests for this part of a task) and there is no demo link, so I couldn't check if your solution works fine :(

I'm sorry for rejection, but I also didn't have possibility to check it fully,

yurii-filiushkin:
[DEMO LINK](https://yurii-filiushkin.github.io/react_todo-app-add-and-delete/)

I can`t understand why some tests don`t passed

kinqbert (approved):
Thanks for the demo link! Everything seems to work as expected there. If everything's fine on your end, than it must be a problem with GitHub or something. 

But still, if you have enough time, try to play around with focusing on main input field, it seems linke something might be wrong with it so tests don't pass (check screenshot below). 

image

However everything still works fine on demo, so approved by me. Have a good day :)

PR #1280, opened 2024-06-04
does not display the loader on tap when the TodoDeleteButton is pressed

AnnBusya:
[Demo](https://AnnBusya.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,133 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useState, useEffect, useRef } from 'react';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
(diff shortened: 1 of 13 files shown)


l4st1m0za (changes requested):
Please, fix the issue with loader before sending another request. Feel free to ask in FE chat

src/App.tsx
```diff
@@ -1,26 +1,165 @@
...
+      >
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => setError(null)}
+          disabled={isSubmitting}
+        />
+        {error}
+      </div>
```
vadiimvooo, line 162: Create error component and move layout of it inside.

src/App.tsx
```diff
@@ -1,26 +1,165 @@
...
+    if (error !== '') {
+      const timer = setTimeout(() => {
+        setError('');
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+
+    return;
+  }, [error]);
```
vadiimvooo, line 33: Create error component and move logic of it inside.

vadiimvooo (changes requested):
Almost done, just one suggestion.

PR #1296, opened 2024-06-18
add task solution

Esther-Guimaraes:
[DEMO LINK](https://Esther-Guimaraes.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,139 @@
...
+        setTodosList(currentTodos =>
+          currentTodos.filter(todo => todo.id !== todoId),
+        );
+        setFilteredTodoList(currentTodos =>
+          currentTodos.filter(todo => todo.id !== todoId),
+        );
+      })
+      .catch(() => {
+        setError(ErrorsEnum.UANBLE_DELETE_TODO);
+        setTimeout(() => setError(null), 3000);
```
pedro-ruas, line 54: This _setTimeOut_ could be used inside a an _useEffect_ hook that tracks _error_.
Why? For now, you only have this error, but in the future, you can have multiple errors being fired everywhere. (one for deleting a Todo, another for creating one, etc...)
Using a hook to track the error and apply the timeOut will shorten your code and keep all appearances of this timeout in a single place

src/App.tsx
```diff
@@ -1,26 +1,139 @@
...
+import { ErrorsEnum } from './utils/ErrorsEnum';
+import { Error } from './components/Error/Error';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { Header } from './components/Header/Header';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todoList, setTodosList] = useState<Todo[]>([]);
+  const [filteredTodoList, setFilteredTodoList] = useState<Todo[]>([]);
```
pedro-ruas, line 15: You don't have to save _filteredTodoList_ in your state.
Instead you can use an _useMemo_ hook to create a variable outside of the state that filters the result based on other state aspects.
The way your code is working, you have one state handling the filter type, another handling the filter result, when in fact, whenever the filter type changes, the filter result will also change, so they can be used in a single state.

src/App.tsx
```diff
@@ -1,26 +1,139 @@
...
+  }, [todoList]);
+
+  if (!todosService.USER_ID) {
     return <UserWarning />;
   }
 
+  const addTodo = () => {
+    if (inputValue.trim() === '') {
+      setError(ErrorsEnum.TITLE_IS_EMPTY);
+      setTimeout(() => setError(null), 3000);
```
pedro-ruas, line 76: Here goes the same advice from the above _setTimeout_

src/App.tsx
```diff
@@ -1,26 +1,139 @@
...
+
+    setTempTodo({
+      id: 0,
+      userId: 0,
+      title: inputValue.trim(),
+      completed: false,
+    });
+
+    todosService
+      .createTodos(inputValue.trim())
```
pedro-ruas, line 90: You have used _inputValue.trim()_ three times in inside this function, maybe it's a good idea to save it on a variable at the function beginning.
```
const title = inputValue.trim()
```

src/App.tsx
```diff
@@ -1,26 +1,139 @@
...
+      .then(newTodo => {
+        setTodosList(currentTodos => [...currentTodos, newTodo]);
+        setFilteredTodoList(currentTodos => [...currentTodos, newTodo]);
```
pedro-ruas, line 93: Here is another example of why you shouldn't save the filtered todos in the state.
Everytime you update your TodoList, you also need to update the FilteredTodoList

src/App.tsx
```diff
@@ -1,26 +1,139 @@
...
+        setInputValue('');
+      })
+      .catch(() => {
+        setError(ErrorsEnum.UNABLE_ADD_TODO);
+        setTimeout(() => setError(null), 3000);
```
pedro-ruas, line 98: And the same advice for  _setTimeout_

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { TodoFilter } from '../TodoFilter/TodoFilter';
+
+export const Footer: React.FC<TodoFooterProps> = ({
+  todoList,
+  isClearDisabled,
+  filterTodoList,
+  clearCompleted,
+  updateClearDisabled,
+}) => {
+  const getItemsLeft = useMemo(() => {
```
pedro-ruas, line 12: Nice!

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,57 @@
...
+        className={`todo ${todo.completed ? 'completed' : ''}`}
+        key={todo.id}
+      >
+        <label className="todo__status-label" htmlFor={`todo-${todo.id}`}>
+          <input
+            id={`todo-${todo.id}`}
+            data-cy="TodoStatus"
+            type="checkbox"
+            className="todo__status"
+            checked={todo.completed}
```
pedro-ruas, line 31: To prevent the above lint error pointed by GitHub, instead of using _id_ you should use the _name_ property for this input
(htmlFor on the label tracks the name property on an input)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const TodoFilter: React.FC<TodoFilterProps> = ({
+  todoList,
+  filterTodoList,
+  updateClearDisabled,
+}) => {
+  const [selectedFilter, setSelecetedFilter] = useState<FiltersEnum>(
+    FiltersEnum.All,
+  );
+
+  useEffect(() => {
```
pedro-ruas, line 14: Instead of doing it here, you could do it on the parent component.
In this way, you would not need to pass _todoList_, _filterTodoList_, and _updateClearDisabled_ as props.
You could also pass this state to te parent
The only prop you would need is _setSelectedFilter_ and _selectedFilter_.

src/types/ComponentsProps.ts
pedro-ruas, line 1: If you have a folder for each component, you should also keep that component's types in the same folder.
For example, on Todo folder, you could either have a file "/Todo/todo.types.ts"with the interface TodoProps, or just move the interface to the same Todo.tsx

pedro-ruas (changes requested):

src/App.tsx
```diff
@@ -1,26 +1,156 @@
...
+    return todoList.filter(todo => !todo.completed).length;
+  }, [todoList]);
+
+  const updateClearDisabled = useCallback(() => {
+    setIsClearDisabled(!todoList.some(todo => todo.completed));
+  }, [todoList]);
+
+  useEffect(() => {
+    setTimeout(() => setError(null), 3000);
+  }, [error]);
```
jv-aquino, line 48: nice!

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export const TodoList: React.FC<TodoListProps> = ({
+  todoList,
+  tempTodo,
+  deleteTodo,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {/* This is a completed todo */}
```
jv-aquino, line 18: i think you forgot to remove some comments or something

src/App.tsx
```diff
@@ -1,26 +1,156 @@
...
+  const handleClearCompleted = useCallback(() => {
+    todoList.map(todo => {
+      if (todo.completed) {
+        handleDeleteTodo(todo.id);
+      }
+    });
+  }, [todoList]);
+
+  const addTodo = () => {
+    const title = inputValue.trim();
```
jv-aquino, line 88: nice improvement

src/App.tsx
```diff
@@ -1,26 +1,156 @@
...
+        setTodosList(currentTodos =>
+          currentTodos.filter(todo => todo.id !== todoId),
+        );
+      })
+      .catch(() => {
+        setError(ErrorsEnum.UANBLE_DELETE_TODO);
+      })
+      .finally(() => {
+        setIsEnabled(true);
+      });
```
jv-aquino, line 76: nice error handling!

jv-aquino (approved):
great job esther!

PR #1309, opened 2024-07-03
save progress

Oleksii-Bidiak:
[DEMO LINK](https://Oleksii-Bidiak.github.io/react_todo-app-add-and-delete/)


volodymyr-soltys97 (changes requested):
Hi, It's bad practice to write so much code in one place, you need to separate it into different components, for example TodoItem, TodoList, Header, Footer

volodymyr-soltys97 (approved):
Well done! 🔥

PR #1310, opened 2024-07-05
add task solution

dima91020:
- [DEMO LINK](https://dima91020.github.io/react_todo-app-add-and-delete/)


volodymyr-soltys97 (changes requested):
Hi, you need to fix the failed tests, feel free in the fe_chat if you need help 
image

PR #1315, opened 2024-07-09
Add delete and add todos

GoVolodya:
[DEMO LINK](https://GoVolodya.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,335 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useEffect, useRef, useState } from 'react';
+import { addTodo, deleteTodo, getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { Errors } from './types/Errors';
+import { Filter } from './types/Filter';
+import classNames from 'classnames';
+import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterBy, setFilterBy] = useState(Filter.all);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [isSending, setIsSending] = useState(false);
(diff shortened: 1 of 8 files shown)


src/App.tsx
anastasiiavorobiova, line 1: Consider splitting the logic of this component into different components to avoid God-components

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/App.tsx
GoVolodya: I have splatted App component for some smaller parts.

src/types/FooterProps.ts
```diff
@@ -0,0 +1,9 @@
+import { Todo } from './Todo';
+import { Filter } from './Filter';
+
+export interface FooterProps {
+  todos: Todo[];
+  filterBy: string;
+  setFilterBy: (filterBy: Filter) => void;
+  handleClearCompleted: () => void;
+}
```
etojeDenys, line 9: it's redundant to create separate files for props, better to keep them close to the component

src/App.tsx
```diff
@@ -1,26 +1,209 @@
...
+    loadTodos();
+  }, []);
+
+  useEffect(() => {
+    focusNewTodoInput();
+  }, []);
+
+  useEffect(() => {
+    focusNewTodoInput();
+  }, [todos]);
```
etojeDenys, line 169: combine that in one useEffect

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,76 @@
...
+          tempTodo={tempTodo}
+          onDelete={handleTodoDelete}
+          deletingCompleted={deletingCompleted}
+        />
+      ))}
+
+      {tempTodo && (
+        <TodoItem
+          todo={tempTodo}
+          tempTodo={tempTodo}
```
etojeDenys, line 39: looks strange, let's just pass into TodoItem optional prop isLoading, so here you need to add
```suggestion
          todo={tempTodo}
          isLoading
```

etojeDenys (changes requested):

src/types/FooterProps.ts
GoVolodya: Sure. It's way more easier to and new props or change existing now.

src/App.tsx
GoVolodya: Improved.

src/components/Main/Main.tsx
GoVolodya: Oh, that is right we can remove one prop, thanks for that.

GoVolodya:
Have done some small fixes for components props.

lerastarynets (approved):
lgtm 🚀

PR #1320, opened 2024-07-09
solution

yaroslav1177:
[DEMO LINK](https://yaroslav1177.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,139 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { createTodo, getTodos, USER_ID, deleteTodo } from './api/todos';
+import { Filter } from './types/Filter';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { TodoList } from './components/TodoList';
+import classNames from 'classnames';
+import { getPreparedTodos } from './utils/preparedTodos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<Filter>(Filter.All);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [isLoading, setIsLoading] = useState<number[]>([]);
+  const [isDisabledInput, setIsDisabledInput] = useState(false);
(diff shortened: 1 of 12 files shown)


yuliaPel (approved):
Good job)

src/App.tsx
```diff
@@ -1,26 +1,139 @@
...
+        />
+
+        <TodoList
+          todos={filteredTodos}
+          onDelete={deleteTodoId}
+          tempTodo={tempTodo}
+          isLoading={isLoading}
+        />
+
+        {todos.length && (
```
dinaminahi, line 110: This way when you don't have todos it shows 0
image

Just make it boolean
```suggestion
        {!!todos.length && (
```

src/App.tsx
```diff
@@ -1,26 +1,139 @@
...
+      >
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => setErrorMessage('')}
+        />
+        {errorMessage}
+      </div>
+    </div>
```
dinaminahi, line 137: Let's create a component for this

src/components/Footer.tsx
```diff
@@ -0,0 +1,79 @@
...
+export const Footer: React.FC<Props> = ({
+  todos,
+  filterBy,
+  handleSelect,
+  onDelete,
+}) => {
+  const clearCompleted = () => {
+    const allCompleted = getCompletedTodos(todos);
+
+    return allCompleted.map(todo => onDelete(todo.id));
```
dinaminahi, line 22: Should this function return something? If no let's just use forEach

And you are calling `getCompletedTodos` funtion 2 times in the component, let's just declare a variable for it out of this funtion

``

dinaminahi, line 22:
```suggestion
  const allCompleted = getCompletedTodos(todos);

  const clearCompleted = () => {
     allCompleted.forEach(todo => onDelete(todo.id));
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,79 @@
...
+          Completed
+        </a>
+      </nav>
+
+      {/* this button should be disabled if there are no completed todos */}
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        disabled={!getCompletedTodos(todos).length}
```
dinaminahi, line 72:
```suggestion
        disabled={!allCompleted.length}
```

src/components/Header.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onDisabled,
+  onError,
+}) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  });
```
dinaminahi, line 24: This useEffect hook  lacks a dependency array, which means it will run on every render. This could lead to performance issues or unwanted behavior, such as repeatedly focusing the input on every render

dinaminahi (approved):
Good job 👍 Let's fix all this comments in the next task

PR #1323, opened 2024-07-10
add task solution

demkoveronika:
[DEMO LINK](https://demkoveronika.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,134 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import * as todoService from './api/todos';
 
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
+import { FilterBy } from './types/FilterBy';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [valueError, setValueError] = useState('');
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.All);
+  const [isNotificationHidden, setIsNotificationHidden] = useState(true);
+  const [isLoading, setIsLoading] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const autoHideNotification = () => {
+    setTimeout(() => {
+      setIsNotificationHidden(true);
+    }, 3000);
+  };
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,26 +1,134 @@
...
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const todosFromServer = await todoService.getTodos();
+
+        setTodos(todosFromServer);
+      } catch (error) {
+        setValueError('Unable to load todos');
+        setIsNotificationHidden(false);
```
dinaminahi, line 38: I think we can have only one state to control error,  so when the error is not empty string we show notification
` { hidden: !valueError.length }`

src/components/Footer.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const allTodosCompleted = todos.filter(todo => todo.completed);
+
+    allTodosCompleted.map(todo => onDelete(todo.id));
+  };
+
+  return (
+    <>
+      <footer className="todoapp__footer" data-cy="Footer">
+        <span className="todo-count" data-cy="TodosCounter">
+          {todosCounter} items left
```
dinaminahi, line 31: Let's minimize array methods, as they can be heavy

```suggestion
  const allTodosCompleted = todos.filter(todo => todo.completed);

  const handleDeleteCompletedTodos = () => {
    allTodosCompleted.forEach(todo => onDelete(todo.id));
  };

  return (
    <>
      <footer className="todoapp__footer" data-cy="Footer">
        <span className="todo-count" data-cy="TodosCounter">
          {todos.length - allTodosCompleted.length} items left
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,79 @@
...
+            Completed
+          </a>
+        </nav>
+
+        <button
+          type="button"
+          className="todoapp__clear-completed"
+          data-cy="ClearCompletedButton"
+          onClick={handleDeleteCompletedTodos}
+          disabled={!todosCompleted}
```
dinaminahi, line 72:
```suggestion
          disabled={!allTodosCompleted.length}
```

src/components/Header.tsx
```diff
@@ -0,0 +1,103 @@
...
+        title: trimmedTitle,
+        userId: newTodo.userId,
+        completed: newTodo.completed,
+      });
+
+      onAdd(createdTodo);
+
+      setValueTitle('');
+      setTempTodo(null);
+      inputRef.current?.focus();
```
dinaminahi, line 63: We can create useEffect in App.tsx which will focus input when there are changes and you won't repeat so many times

src/components/TodoList.tsx
```diff
@@ -0,0 +1,48 @@
...
+              title={title}
+              completed={completed}
+              onDelete={onDelete}
+              isLoading={isLoading}
+            />
+          </CSSTransition>
+        ))}
+
+        {tempTodo && (
+          <CSSTransition key={tempTodo.id} timeout={300} classNames="item">
```
dinaminahi, line 35: there should be different class for temp todo animation
```suggestion
          <CSSTransition key={tempTodo.id} timeout={300} classNames="temp-item">
```

dinaminahi (approved):
Good job 👍 
Few fixes needed, please include them in your next task
Loader should be only on loading item not on all of them

PR #1340, opened 2024-07-15
TempTodo should be add before submit

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,10 +5,15 @@ import { Errors } from './components/Errors/Errors';
 import { TodoContent } from './components/TodoContent/TodoContent';
 import { Footer } from './components/Footer/Footer';
 import { TodoList } from './components/TodoList/TodoList';
+import { useClearCompleted } from './hooks/useClearCompleted';
 
 const AppContent: React.FC = () => {
   const { todos, filteredTodos, error, setFilter, setError, filter } =
     useTodos();
+  const { handleClearCompleted, error: clearCompletedError } =
+    useClearCompleted();
+
+  const combinedError = error || clearCompletedError;
 
   if (!todos) {
     return <UserWarning />;
@@ -17,10 +22,15 @@ const AppContent: React.FC = () => {
   return (
     <div className="todoapp">
       <TodoContent onErrorChange={setError}>
-        <TodoList todos={filteredTodos} />
-        <Footer todos={todos} filter={filter} onFilterChange={setFilter} />
+        <TodoList todos={filteredTodos} onErrorChange={setError} />
+        <Footer
+          todos={todos}
+          filter={filter}
+          onFilterChange={setFilter}
+          onClearCompleted={handleClearCompleted}
+        />
       </TodoContent>
-      <Errors error={error} onErrorChange={setError} />
+      <Errors error={combinedError} onErrorChange={setError} />
     </div>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { TodoProvider, useTodos } from './utils/TodoContext'; // Poprawny import TodoProvider
+import { TodoProvider, useTodos } from './utils/TodoContext';
 import { UserWarning } from './UserWarning';
 import { Errors } from './components/Errors/Errors';
 import { TodoContent } from './components/TodoContent/TodoContent';
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,8 +8,7 @@ import { TodoList } from './components/TodoList/TodoList';
 import { useClearCompleted } from './hooks/useClearCompleted';
 
 const AppContent: React.FC = () => {
-  const { todos, filteredTodos, error, setFilter, setError, filter } =
-    useTodos();
+  const { todos, filteredTodos, error, filter } = useTodos();
   const { handleClearCompleted, error: clearCompletedError } =
     useClearCompleted();
 
@@ -21,16 +20,15 @@ const AppContent: React.FC = () => {
 
   return (
     <div className="todoapp">
-      <TodoContent onErrorChange={setError}>
-        <TodoList todos={filteredTodos} onErrorChange={setError} />
+      <TodoContent>
+        <TodoList todos={filteredTodos} />
         <Footer
           todos={todos}
           filter={filter}
-          onFilterChange={setFilter}
           onClearCompleted={handleClearCompleted}
         />
       </TodoContent>
-      <Errors error={combinedError} onErrorChange={setError} />
+      <Errors error={combinedError} />
     </div>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
+import { TodoProvider, useTodos } from './utils/TodoContext'; // Poprawny import TodoProvider
 import { UserWarning } from './UserWarning';
-import { TodoProvider, useTodos } from './utils/TodoContext';
 import { Errors } from './components/Errors/Errors';
 import { TodoContent } from './components/TodoContent/TodoContent';
 import { Footer } from './components/Footer/Footer';
@@ -8,7 +8,7 @@ import { TodoList } from './components/TodoList/TodoList';
 import { useClearCompleted } from './hooks/useClearCompleted';
 
 const AppContent: React.FC = () => {
-  const { todos, filteredTodos, error, filter } = useTodos();
+  const { todos, error } = useTodos();
   const { handleClearCompleted, error: clearCompletedError } =
     useClearCompleted();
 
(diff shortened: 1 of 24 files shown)


natalia-klonowska (changes requested):
make deploy and add demo link in pr description 

the problem is that you're adding temp todo to todos array and then replace it with createdTodo 
image

przwojwwp:
nie mogę wyświetlić błędu przy usuwaniu todo

Zibi95 (changes requested):
Hey. There is infinite loading spinner on every todo. I can't test functionalities. Nice that you are using custom hooks but there is so many of them that its hard to follow what comes from what.

przwojwwp:
indeed, the names help a bit but it's hard to follow. Do you suggest that I should modify them so that everything is in components instead of hooks? At this stage of the task it will be very difficult

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        disabled={completedTodosCount === 0}
+        onClick={onClearCompleted} // Add onClick handler
+      >
+        Clear completed
```
Zibi95, line 67: There is clearly the same pattern for every anchor. What about creating array with objects that keep the values needed to properly create every anchor with map.

Zibi95 (changes requested):
Hey. You don't have too change anything about custom hooks. I'm just saying that its powerful tool but like with everything if it is overused it becomes bad. There is problem with logic of handling errors. Every error displays only once. When i try to add todo without title i get notified only once also for every TodoItem i get notified only once when i try to delete it.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {activeTodosCount} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        {filters.map(({ type, label, cy, href }) => (
+          <a
+            key={type}
+            href={href}
+            className={`filter__link ${filter === type ? 'selected' : ''}`}
```
natalia-klonowska, line 52: use classnames library to improve readability of conditional classes

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+  onErrorChange: (error: ErrorType | null) => void; // Add onErrorChange prop
+};
+
+export const TodoItem: React.FC<TodoItemProps> = ({
+  todo,
+  isTemp,
+  onErrorChange,
+}) => {
+  const { deleteTodo, isDeleting, error } = useDeleteTodo(); // Get error from useDeleteTodo
+  const [showLoader, setShowLoader] = useState<boolean>(false);
```
natalia-klonowska, line 19: you don't need to specify type because TypeScript can determine it from initial value

src/App.tsx
```diff
@@ -1,26 +1,46 @@
...
-          React Todo App - Load Todos
-        </a>
-      </p>
+    <div className="todoapp">
+      <TodoContent onErrorChange={setError}>
+        <TodoList todos={filteredTodos} onErrorChange={setError} />
+        <Footer
+          todos={todos}
+          filter={filter}
+          onFilterChange={setFilter}
```
natalia-klonowska, line 29: you shouldn't pass state setters like setFilter and setError as props and if you're using Context you can get access these props directly inside your components instead of passing them

natalia-klonowska (changes requested):

src/App.tsx
```diff
@@ -8,8 +8,7 @@ import { TodoList } from './components/TodoList/TodoList';
 import { useClearCompleted } from './hooks/useClearCompleted';
 
 const AppContent: React.FC = () => {
-  const { todos, filteredTodos, error, setFilter, setError, filter } =
-    useTodos();
+  const { todos, filteredTodos, error, filter } = useTodos();
```
natalia-klonowska, line 11: you don't use filter and filteredTodos you just pass it as props to other components

src/App.tsx
```diff
@@ -21,16 +20,15 @@ const AppContent: React.FC = () => {
 
   return (
     <div className="todoapp">
-      <TodoContent onErrorChange={setError}>
-        <TodoList todos={filteredTodos} onErrorChange={setError} />
+      <TodoContent>
+        <TodoList todos={filteredTodos} />
         <Footer
           todos={todos}
           filter={filter}
```
natalia-klonowska, line 27: don't pass filteredTodos, todos and filter because you have it inside Context

src/App.tsx
```diff
@@ -1,26 +1,44 @@
...
 
-const USER_ID = 0;
+const AppContent: React.FC = () => {
+  const { todos, filteredTodos, error, filter } = useTodos();
+  const { handleClearCompleted, error: clearCompletedError } =
+    useClearCompleted();
 
-export const App: React.FC = () => {
-  if (!USER_ID) {
+  const combinedError = error || clearCompletedError;
```
natalia-klonowska, line 15: you can move it inside Errors component

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,49 @@
...
+      }
+    };
+  }, [error, setError]);
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={`notification is-danger is-light has-text-weight-normal ${
+        !error ? 'hidden' : ''
+      }`}
```
natalia-klonowska, line 38: use classnames everywhere you use conditional classes

src/components/TodoContent/TodoContent.tsx
```diff
@@ -0,0 +1,61 @@
+import * as React from 'react';
+import { ChangeEvent, FormEvent, useEffect, useState } from 'react';
+import { usePostTodos } from '../../hooks/usePostTodos';
+import { useTodos } from '../../utils/TodoContext';
+
+export const TodoContent: React.FC<{ children: React.ReactNode }> = ({
+  children,
+}) => {
+  const [input, setInput] = useState<string>('');
```
natalia-klonowska, line 9: don't need type cos it's the same as initial value (find and fix all instances where it's not needed)

natalia-klonowska (changes requested):

src/App.tsx
przwojwwp: if I change it code stop working

przwojwwp:
never been on the theory about classnames and its syntax. If there are so many things to reorganize, I should probably do the task from the beginning. I look through other people's pull requests and these things are not implemented in them

przwojwwp:
I cant make it myself. I need help with it

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+export const Footer: React.FC<FooterProps> = ({
+  todos,
+  filter,
+  onClearCompleted,
+}) => {
+  const { setFilter } = useTodos();
+
+  if (todos.length === 0) {
```
BogdanMaliuta, line 20:
```suggestion
  if (!todos?.length) {
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return null;
+  }
+
+  const activeTodosCount = todos.filter(todo => !todo.completed).length;
+  const completedTodosCount = todos.filter(todo => todo.completed).length;
+
+  const filters = [
+    { type: FilterType.All, label: 'All', cy: 'FilterLinkAll', href: '#/' },
```
BogdanMaliuta, line 28: split it

BogdanMaliuta (changes requested):
Hello, fix what Natalia described, use your hook useTodos in components where you needs todos, if it`s stops working, just fix your hook, always remove comments from code, take a look on your types, remove it where you have initial value, and check your code style, you can improve it in some places

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,58 @@
...
+        setClearCompletedError(null);
+        timerRef.current = null;
+      }, 3000);
+    }
+
+    return () => {
+      if (timerRef.current) {
+        clearTimeout(timerRef.current);
+        timerRef.current = null;
+      }
```
Zibi95, line 29: You don't need to have ref to keep timer id. You can simply define variable using `let` inside body of `effect`

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const Footer: React.FC = () => {
+  const { todos, filter, setFilter } = useTodos();
+  const { handleClearCompleted } = useClearCompleted();
+
+  if (!todos?.length) {
+    return null;
+  }
+
+  const activeTodosCount = todos.filter(todo => !todo.completed).length;
+  const completedTodosCount = todos.filter(todo => todo.completed).length;
```
Zibi95, line 16: I guess you don't have to filter out completedTodos. You can just now do some maths with activeTodosCount and all todos so 
`todos.length - activeTodosCount = completedTodosCount`

Zibi95 (approved):
Good job! However, I feel that you've overused custom hooks. While it might look clean because there is less code in the components, it can become really hard to understand when you need to debug the logic. Using custom hooks within other custom hooks can also lead to complex code. What I tend to do is create the logic without using a custom hook initially. Then, if I find a place where I need to reuse the logic, or if a significant amount of code is required to achieve the desired result, I create a custom hook. I can approve this task as you have met the requirements for this task, but try to rethink the logic and simplify the code for other task in this series. Don't skip react go as the roadmap goes. Redux is some extra and its useful if you build rly big apps.

PR #1345, opened 2024-07-17
solution

Kozubowicz:
[DEMO LINK](https://Kozubowicz.github.io/react_todo-app-add-and-delete/)


src/types/FilterType.ts
```diff
@@ -0,0 +1,5 @@
+export enum FilterType {
+  All,
+  Active,
+  Complited,
```
Zibi95, line 4: *Completed

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+        onClick={() => removeOneTodo(id)}
+      >
+        ×
+      </button>
+
+      {/* overlay will cover the todo while it is being deleted or updated */}
+      <div
+        data-cy="TodoLoader"
+        className={classNames('modal overlay', {
+          'is-active': id < 0 || updatedTodosId.includes(id),
```
Zibi95, line 48: You could just pass `isUpdating` here as prop and check if it is updating while mapping list of todos

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+  updatedTodosId: number[];
+  removeOneTodo: (id: number) => void;
+};
+
+export const TodoItem = ({
+  id,
+  completed,
+  title,
+  updatedTodosId,
+  removeOneTodo,
```
Zibi95, line 18: There is a convention of passing a handler through props by calling it `onRemove`.

src/components/Header.tsx
```diff
@@ -0,0 +1,30 @@
...
+  onSubmit: (event: React.FormEvent<HTMLFormElement>) => void;
+};
+
+export const Header = ({ inputRef, onSubmit }: HeaderProps) => {
+  return (
+    <header className="todoapp__header">
+      {/* this button should have `active` class only if all todos are completed */}
+      <button
+        type="button"
+        className="todoapp__toggle-all active"
```
Zibi95, line 14: As the comment above says, this button should have `active` class only if all todos are completed

src/components/Header.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+type HeaderProps = {
+  inputRef: React.Ref<HTMLInputElement>;
```
Zibi95, line 4: The ref should be passed to child component by using `forwardRef`
`const MyIChildComponent= forwardRef(function MyInput(props, ref)`

here is documentation https://react.dev/reference/react/forwardRef

src/App.tsx
```diff
@@ -1,26 +1,186 @@
...
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [error, setError] = useState<ErrorType | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
Zibi95, line 17: The value of `filteredTodos` can be easily computed. This state is unnecessary and also requires an additional `useEffect`, which we always strive to minimize.

src/App.tsx
```diff
@@ -1,26 +1,186 @@
...
+      })
+      .catch(() => setError(ErrorType.UnableToDelete))
+      .finally(() => {
+        setUpdatedTodosId(prev => prev.filter(el => el != id));
+        inputRef.current?.focus();
+      });
+  };
+
+  const removeAllComplited = () => {
+    const tmp = todos.filter(todo => todo.completed).map(todo => todo.id);
```
Zibi95, line 136: Use proper names for variables for example here weould be good `todosToRemoveIds`. Give names some meaning.

src/App.tsx
```diff
@@ -1,26 +1,186 @@
...
+    setUpdatedTodosId(tmp);
+
+    tmp.map(id => {
+      removeOneTodo(id);
+    });
+  };
+
+  const isClearAllCompletedActive = () => {
+    return !todos.some(todo => todo.completed);
+  };
```
Zibi95, line 147: This names makes me think that it keeps value of type `boolean`. Change the name or remove this function and compute to `boolean` value here

src/App.tsx
```diff
@@ -1,26 +1,186 @@
...
+      case ErrorType.UnableToLoad:
+        return 'Unable to load todos';
+      case ErrorType.UnableToUpdate:
+        return 'Unable to update a todo';
+      case ErrorType.UnableToAdd:
+        return 'Unable to add a todo';
+      default:
+        return null;
+    }
+  };
```
Zibi95, line 69: This can be colocated down to component that shows error as you already pass error with props

src/App.tsx
```diff
@@ -1,26 +1,186 @@
...
+
+  useEffect(() => {
+    const timeout = setTimeout(() => {
+      setError(null);
+    }, 3000);
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [error]);
```
Zibi95, line 79: This one also could be colocated to the component that receives it through props.

src/App.tsx
```diff
@@ -1,26 +1,186 @@
...
+          }
+
+          setTmpTodo(null);
+        });
+    } else {
+      setError(ErrorType.EmptyTitle);
+    }
+  };
+
+  const removeOneTodo = (id: number) => {
```
Zibi95, line 121: More meaningful name for this function would be `handleRemoveTodoById`

Zibi95 (changes requested):
Good job. I left some comments to improve the code. Always use names that describe what value is kept inside or what function is doing. The button to check all active todos should be inactive when there are still some active todos.

src/components/Header.tsx
Kozubowicz: image

src/App.tsx
```diff
@@ -1,26 +1,157 @@
...
+    setUpdatedTodosId(todosToRemoveIds);
+
+    todosToRemoveIds.map(id => {
+      handleRemoveTodoById(id);
+    });
+  };
+
+  const isClearAllCompletedActive = () => {
+    return !todos.every(todo => todo.completed);
+  };
```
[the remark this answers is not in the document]
Kozubowicz: I know that it return now false only when all todos are completed, because I typed that in coment to your friend who said that button should be active only when all todos are completed couse the code comment saying that 
image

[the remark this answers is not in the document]
natalia-klonowska: we are taking about two different buttons 😅 
toggle-all:
image
and clear all:
image

src/App.tsx
```diff
@@ -1,26 +1,191 @@
...
+          break;
+        case FilterType.Completed:
+          setFilteredTodos(todos.filter(todo => todo.completed));
+          break;
+        default:
+          setFilteredTodos(todos);
+          break;
+      }
+    }
+  }, [todos, query]);
```
Zibi95, line 53:
```suggestion
const filterTodos = (query, todos) => {
  if (query === FilterType.Active) {
    return todos.filter(todo => !todo.completed);
  } else if (query === FilterType.Completed) {
    return todos.filter(todo => todo.completed);
  } else {
    return todos;
  }
};

const filteredTodos = filterTodos(todos, query)
```

src/App.tsx
```diff
@@ -1,26 +1,191 @@
...
+import { OptionaAtributs } from './types/OptionaAtributs';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [error, setError] = useState<ErrorType | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
Zibi95, line 18: Then this can be removed

```suggestion
```

Zibi95 (changes requested):
Hey,
There are still some issues that need fixing:

- The "Toggle All" button should not be visible when there are no todos in the list.
image

- The button currently only changes todos to active; it should also be able to mark todos as completed.
- If there are both completed and active todos, the "Toggle All" button should first mark active todos as completed. If all todos are already completed, it should unmark them, making them active again.

natalia-klonowska (approved):
Great work 😃 you can start last part of this project ^^

PR #1348, opened 2024-07-18
First Solution

DimaK123:
[DEMO LINK](https://dimak123.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,244 @@
...
+  const [filterStatus, setFilterStatus] = useState<Filter>(Filter.All);
+  const [inputValue, setInputValue] = useState('');
+  const [tempTodo, setTempTodo] = useState<null | string>(null);
+  const [lockInput, setLockInput] = useState(false);
+  const [deleteAll, setDeleteAll] = useState(false);
+
+  const [loadingError, setLoadingError] = useState(false);
+  const [titleError, setTitleError] = useState(false);
+  const [todoAddError, setTodoAddError] = useState(false);
+  const [deleteError, setDeleteError] = useState(false);
```
VitaliyBondarenko1982, line 30: it can be one error with appropriate error text string or empty string in case no error

src/App.tsx
```diff
@@ -1,26 +1,244 @@
...
+    try {
+      const deleteAllActive = todosFromServer.map(async todo => {
+        const { id, completed } = todo;
+
+        if (completed) {
+          await handleDeleteTodo(id);
+        }
+      });
+
+      await Promise.all(deleteAllActive);
```
VitaliyBondarenko1982, line 129: test will expect handling each separate delete request. so better in forEach loop send delete request.
in success cases todo should be deleted.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments and fix demo link.
Deploy was unsuccesful
https://github.com/DimaK123/react_todo-app-add-and-delete/actions

VitaliyBondarenko1982 (changes requested):
.

src/App.tsx
DimaK123: Fixed

src/App.tsx
DimaK123: I didn't fully understand how to use forEach since I need to pass an array into Promise.allSettled(), which doesn't work with this array iteration method. Therefore, I used the map method.

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,66 @@
...
+        setFilterStatus(Filter.Active);
+        break;
+
+      case Filter.Completed:
+        setFilterStatus(Filter.Completed);
+        break;
+
+      default:
+        setFilterStatus(Filter.All);
+    }
```
etojeDenys, line 27:
```suggestion
setFilterStatus(text);
```
we can simplify that

etojeDenys (approved):
good job

PR #1353, opened 2024-07-22
develop

akryto:
https://akryto.github.io/react_todo-app-add-and-delete/

Add a todo with the entered title on the form submit:

- text field should be focused by default;
- if the title is empty show the `Title should not be empty` notification at the bottom;
- trim the title when checked or saved;
- use your `userId` for the new todo;
- send a POST request to the API (check the [API Documentation](https://mate-academy.github.io/fe-students-api/))
- disable the input until receiving a response from the API;
- immediately after sending a request create a todo with `id: 0` and save it to the `tempTodo` variable in the state (NOT to the `todos` array);
- show an independent `TodoItem` **after** the list if `tempTodo` is not `null`;
- temp TodoItem should have the loader (check the original markup);
- in case of success add the todo created by the API to the array (take it from the POST response);
- in case of an API error showing `Unable to add a todo` notification at the bottom;
- set `tempTodo` to `null` to hide the extra `TodoItem`;
- focus the text field after receiving a response;
- clear the text in case of success;
- keep the text in case of error;

> Don't try to implement animations for adding or removing Todos (at least until you finish everything else).
> If you really feel confident to try, there is a hint at the end of the description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Remove all the completed todos after the `Clear completed` button click:
 - Install Prettier Extention and use this [VSCode settings](https://mate-academy.github.io/fe-program/tools/vscode/settings.json) to enable format on save.
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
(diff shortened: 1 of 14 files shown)


DSB-18 (commented):
Good work!
But there are extra comments, that should be deleted

DSB-18 (approved):
Good work!
But there are extra comments, that should be deleted

PR #1355, opened 2024-07-22
feat: add solution

mfcastilho:
[DEMO LINK](https://mfcastilho.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,170 @@
...
+
+    setIsLoading(true);
+
+    const newTodo = { id: 0, title: newTitle, completed: false };
+
+    setTempTodo(newTodo);
+
+    return createTodo(newTodo)
+      .then(resultTodo => {
+        setTodos([...todos, resultTodo]);
```
pedro-ruas, line 80: It's better to access the state inside the function to prevent call overlapping:
```
setTodos((tds) => [...tds, resultTodo])
```

src/App.tsx
```diff
@@ -1,26 +1,170 @@
...
+        ),
+      )
+      .catch(() => setError('Unable to update a Todo'));
+  };
+
+  const handleDeleteCompleted = () => {
+    setIsLoading(true);
+    const promises: Promise<void>[] = [];
+
+    todos.forEach(todo => {
```
pedro-ruas, line 121: This whole part can be changed to a simpler .map:
```
const promises: Promise<void>[] = todos.filter(td => td.completed)
    .map((todo) => deleteTodo(todo.id)
    .then(...)
```

src/api/todos.ts
```diff
@@ -0,0 +1,20 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 760;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+export const createTodo = (todo: Partial<Todo>) => {
```
pedro-ruas, line 10: `todo: Omit<Todo, "id">`
I believe you must have everything, but the id, right?

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <form onSubmit={handleSubmit}>
+        <input
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          disabled={isLoading}
+          placeholder="What needs to be done?"
+          ref={inputRef}
+          value={initialTitle}
+          onChange={event => onTitleChange(event.target.value)}
```
pedro-ruas, line 46: Save this function to a callback using useCallback to better memory usage:
```
const handleTitleChange = useCallback((event) => onTitleChange(event.target.value), [onTitleChange]);

onChange={handleTitleChange}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({
+  visibleToDos,
+  onDelete,
+  onUpdate,
+  tempTodo,
+}) => {
+  const renderToDos = useMemo(
+    () => (tempTodo ? [...visibleToDos, tempTodo] : visibleToDos),
```
pedro-ruas, line 19: Instead, you could add a conditional rendering for a <TodoItem> after the .map method.
In that way, you don't have to re-render the entire list because the tempTodo has changed.

pedro-ruas (changes requested):
Minimal improvemetns

PR #1361, opened 2024-07-26
second part

shelyaa:
[DEMO LINK](https://shelyaa.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,185 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
+  if (filter !== Filter.all) {
+    filteredTodos = todos.filter(todo => {
+      switch (filter) {
+        case Filter.active:
+          return !todo.completed;
+        case Filter.completed:
+          return todo.completed;
+        default:
+          return true;
+      }
```
SanyaBratashchuk, line 66: move filter inside switch case to avoid extra iteration in default case

src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
+    });
+  }
```
SanyaBratashchuk, line 68: create a separate function to filter todos

SanyaBratashchuk (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #1363, opened 2024-07-27
solution

getchards:
[DEMO LINK](https://getchards.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,237 @@
...
+      >
+        {emptyTitle && <p>Title should not be empty</p>}
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={closeErrorHandler}
+        />
+        {errorMessage}
+      </div>
```
vadiimvooo, line 234: Create error component and move template inside of it.

vadiimvooo (changes requested):
Good job, but you have to pass all tests. Please feel free to ask for help in fe_chat

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export const ErrorNotification: React.FC<Props> = ({
+  errorMessage,
+  onClose,
+}) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={`notification is-danger is-light has-text-weight-normal ${!errorMessage && 'hidden'}`}
```
etojeDenys, line 16: better to use classnames for conditional classes

etojeDenys (changes requested):
1. do not make request on filter option change, just change the array
2. add the footer in this case
https://github.com/user-attachments/assets/79057764-1eb9-43d2-928d-8d72755b2779

src/App.tsx
```diff
@@ -1,26 +1,142 @@
...
+
+          default:
+            setTodos(todosFromServer);
+        }
+      })
+      .catch(() => setErrorMessage('Unable to load todos'))
+      .finally(() => {
+        setTimeout(() => setErrorMessage(''), 3000);
+      });
+  }, [setTodos, status]);
```
TarasHoliuk, line 43: Not fixed

> do not make request on filter option change, just change the array

create variable for filteredTodos:
```
// some filtering logic
  return todos.filter()
}, [status])
```

src/App.tsx
```diff
@@ -1,26 +1,142 @@
...
+
+    return addNewTodo(trimmedTodo)
+      .then(todo => {
+        setTodos(todos);
+        setTodos(currentTodos => [...currentTodos, todo]);
+      })
+      .catch(error => {
+        setErrorMessage('Unable to add a todo');
+        setTodos(todos);
+        throw error;
```
TarasHoliuk, line 74: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

src/App.tsx
```diff
@@ -1,26 +1,142 @@
...
+import { Header } from './components/Header';
+import { Todo } from './types/Todo';
+import { TodoStatus } from './types/SortTypes';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [status, setStatus] = useState('');
+  const [idTodo, setIdTodo] = useState(0);
+  const [newTodo, setNewTodo] = useState({
```
TarasHoliuk, line 18: userId and completed will always be the same
So you can simplify it and save only newTitle string in state

src/components/ErrorNotification.tsx
TarasHoliuk: Not fixed

Meant `classnames` library (or any similar):
image

src/App.tsx
```diff
@@ -1,26 +1,142 @@
...
+import { Footer } from './components/Footer';
+import { TodoList } from './components/TodoList';
+import { Header } from './components/Header';
+import { Todo } from './types/Todo';
+import { TodoStatus } from './types/SortTypes';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [status, setStatus] = useState('');
```
TarasHoliuk, line 16:
```suggestion
  const [status, setStatus] = useState(TodoStatus.All);
```
You can use typecasting in Footer if you need

src/components/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {leftItems} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={cn('filter__link', { selected: !status })}
+          data-cy="FilterLinkAll"
+          onClick={() => onClick('')}
```
TarasHoliuk, line 32:
```suggestion
          onClick={() => onClick(TodoStatus.All)}
```

src/components/Header.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+      <form onSubmit={handleSubmit}>
+        <input
+          value={todo.title}
+          ref={titleField}
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          onChange={event => onChange(event.target.value)}
```
TarasHoliuk, line 60: it's better to create handler for it und use here

src/components/Header.tsx
```diff
@@ -0,0 +1,66 @@
...
+  onError,
+}) => {
+  const titleField = useRef<HTMLInputElement>(null);
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    if (titleField.current) {
+      titleField.current.focus();
+    }
+  }, [onReset]);
```
TarasHoliuk, line 26: Why this dependency?
```suggestion
  }, []);
```

src/App.tsx
```diff
@@ -125,15 +125,13 @@ export const App: React.FC = () => {
...
-        {todos.length > 0 && (
-          <Footer
-            onClick={setStatus}
-            status={status}
-            leftItems={leftItemsCount}
-            completedItems={completedItems}
-            onDelete={clearCompletedTodo}
-          />
-        )}
+        <Footer
```
TarasHoliuk, line 128: You should add conditional rendering here
But you should hide footer if there is no todos on server at all

Not when there are no todos with selected status:
image

TarasHoliuk (changes requested):

src/components/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <a
+          href="#/completed"
+          className={cn('filter__link', {
+            selected: status === TodoStatus.Completed,
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => onClick(TodoStatus.Completed)}
+        >
+          Completed
+        </a>
```
volodymyr-soltys97, line 57: Use `Object.values()` and render these options with `map()` method

src/App.tsx
```diff
@@ -1,26 +1,141 @@
...
+        setTodos(todos);
+        setTodos(currentTodos => [...currentTodos, todo]);
+      })
+      .catch(() => {
+        setErrorMessage('Unable to add a todo');
+        setTodos(todos);
+      });
+  }
+
+  function handleChangeTitle(value: string) {
```
volodymyr-soltys97, line 76: It is important to follow the same way of creating functions everywhere
```suggestion
  const handleChangeTitle = (value: string) => {
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,56 @@
...
+          key={id}
+        >
+          {/* eslint-disable jsx-a11y/label-has-associated-control  */}
+          <label className="todo__status-label" htmlFor={'' + id}>
+            <input
+              id={'' + id}
+              data-cy="TodoStatus"
+              type="checkbox"
+              className="todo__status"
+              checked={completed ? true : false}
```
volodymyr-soltys97, line 26:
```suggestion
              checked={completed}
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {list.map(({ title, id, completed }) => (
+        <div
+          data-cy="Todo"
+          className={`todo ${completed && 'completed'}`}
+          key={id}
+        >
+          {/* eslint-disable jsx-a11y/label-has-associated-control  */}
+          <label className="todo__status-label" htmlFor={'' + id}>
```
volodymyr-soltys97, line 20:
```suggestion
          <label className="todo__status-label" htmlFor={id}>
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,56 @@
...
+  idTodo: number;
+};
+
+export const TodoList: React.FC<Props> = ({ list, onDelete, idTodo }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {list.map(({ title, id, completed }) => (
+        <div
+          data-cy="Todo"
+          className={`todo ${completed && 'completed'}`}
```
volodymyr-soltys97, line 16: Use the classnames library rules for add classes with condition

src/components/Header.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <form onSubmit={handleSubmit}>
+        <input
+          value={title}
+          ref={titleField}
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          onChange={handleChange}
+          disabled={loading ? true : false}
```
volodymyr-soltys97, line 76:
```suggestion
          disabled={loading}
```

volodymyr-soltys97 (changes requested):
GJ!
Let's improve your code
Try to fix the failed test, if you need help feel free in the fe_chat

anastasiiavorobiova (approved):
Great work!

PR #1364, opened 2024-07-29
Todo app add and delete

OMedvid:
[DEMO LINK](https://OMedvid.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,59 @@
+import { FC } from 'react';
+import { FilterTupes } from '../types/Filters';
+import React from 'react';
+import cn from 'classnames';
+
+interface Props {
+  onFilter: (filter: FilterTupes) => void;
+  onClear: () => void;
+  selectedFilter: FilterTupes;
+  todosCount: {
+    active: number;
+    completed: number;
+  };
+}
+
+export const Footer: FC<Props> = ({
+  onFilter,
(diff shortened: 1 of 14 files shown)


src/components/todoList.tsx
```diff
@@ -0,0 +1,109 @@
...
+  onDelete,
+  onEdit,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(todo => (
+        <Todo key={todo.id} todo={todo} onDelete={onDelete} onEdit={onEdit} />
+      ))}
+
+      {tempTodoTitle && <TempTodo title={tempTodoTitle} />}
```
etojeDenys, line 29: it can be the same Todo component

for that you need to save the tempTodo instead of tempTodoTitle

src/components/Footer.tsx
```diff
@@ -0,0 +1,59 @@
...
+  };
+}
+
+export const Footer: FC<Props> = ({
+  onFilter,
+  onClear,
+  todosCount,
+  selectedFilter,
+}) => {
+  const filters = [FilterTupes.All, FilterTupes.Active, FilterTupes.Completed];
```
etojeDenys, line 22: you can move it outside of the component as the constant

etojeDenys (approved):
good job

PR #1366, opened 2024-07-29
Add task solution

luishenriquetp:
[DEMO LINK](https://luishenriquetp.github.io/react_todo-app-add-and-delete/)


jv-aquino (approved):
hey luis, there are some minor logic mistakes that I think you should watch out, but as this is a part of the final project, i won't reject it and instead go to the final part

btw the "complete" didnt need to be finished by this part, but thats even better kkkkkk

PR #1369, opened 2024-07-30
solution need help with tests

opalahecha:
[DEMO LINK](https://opalahecha.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/todoFilter.tsx
+++ b/src/components/todoFilter.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+import { Filters } from '../types/Filters';
+import { Todo } from '../types/Todo';
+
+interface Props {
+  currentFilter: Filters;
+  onFilterChange: (filter: Filters) => void;
+  todos: Todo[];
+  activeTodosCount: number;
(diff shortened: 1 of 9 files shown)


etojeDenys (changes requested):
1. image
center filter section like in the working example

2. show loader on todo deleting

src/components/todoFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {activeTodosCount} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={`filter__link ${currentFilter === Filters.All ? 'selected' : ''}`}
```
TarasHoliuk, line 31: FYI
It's better to use library like `classnames` for conditional classes

TarasHoliuk (approved):
Well done - approved ✔️

PR #1374, opened 2024-08-02
task solution

BoGoBooM:
[DEMO LINK](https://bogoboom.github.io/react_todo-app-add-and-delete/)

I just copied the previous task & change 2 lines to solve this task, but this method is works


SanyaBratashchuk (changes requested):
temp todo is not implemented 
image

src/components/errorNotification.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useEffect, useState } from 'react';
+
+interface ErrorNotificationProps {
+  error: string | null;
+}
+
+export const ErrorNotification: React.FC<ErrorNotificationProps> = ({
```
TarasHoliuk, line 7: file name with component should be in PascalCase - `ErrorNotification.tsx` in this case

Fix everywhere

src/components/errorNotification.tsx
```diff
@@ -0,0 +1,39 @@
...
+      return () => clearTimeout(timer);
+    }
+
+    return undefined;
+  }, [error]);
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={`notification is-danger is-light has-text-weight-normal ${hidden ? 'hidden' : ''}`}
```
TarasHoliuk, line 28: Consider using classnames library for conditional classNames

src/components/footer.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { Todo } from '../types/Todo';
+
+interface FooterProps {
+  todos: Todo[];
+  filter: 'all' | 'active' | 'completed';
```
TarasHoliuk, line 6:
```suggestion
  filter: Filter;
```
You should use enum everywhere for it

src/components/footer.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {todos.filter(todo => !todo.completed).length} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={`filter__link ${filter === 'all' ? 'selected' : ''}`}
+          data-cy="FilterLinkAll"
+          onClick={() => onSetFilter('all')}
```
TarasHoliuk, line 28:
```suggestion
          onClick={() => onSetFilter(Filter.All)}
```

src/App.tsx
```diff
@@ -1,26 +1,241 @@
...
-
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [newTodo, setNewTodo] = useState('');
+  const [error, setError] = useState<string | null>(null);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
TarasHoliuk, line 21:
```suggestion
  const [filter, setFilter] = useState<Filter>(Filter.All);
```
You have created enum for it
Actually, you can use union type here instead of enum. but instead of repeating `'all' | 'active' | 'completed'` everywhere you should create such type and reuse

But I'd prefer enum anyway)

src/App.tsx
```diff
@@ -1,26 +1,241 @@
...
+  const [loading, setLoading] = useState<boolean>(false);
```
TarasHoliuk, line 22:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
for boolean variables: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

Generic is redundant here - type will be automatically defined based on initial value (false)

src/App.tsx
```diff
@@ -1,26 +1,241 @@
...
+
+    createTodo(newTodo)
+      .then((newTodoItem: Todo) => {
+        setTodos(prevTodos => [...prevTodos, newTodoItem]);
+        setNewTodo('');
+        setTempTodo(null);
+      })
+      .catch(() => {
+        setError('Unable to add a todo');
+        setTempTodo(null);
```
TarasHoliuk, line 63: do it in finally instead of in both callbacks (then, catch)

src/App.tsx
```diff
@@ -1,26 +1,241 @@
...
+        }),
+      ),
+    )
+      .then(() => {
+        setTodos(updatedTodos);
+      })
+      .finally(() => setLoading(false));
+  };
+
+  const handleDeleteTodo = (todoId: number) => {
```
TarasHoliuk, line 109: You should show the loader when adding/deleting (editing in the next task) todo
See working example and description in the task instructions

P.S. it also related to multiple todo actions such as `Clear completed` button

src/App.tsx
```diff
@@ -1,26 +1,241 @@
...
+      })
+      .finally(() => setLoading(false));
+  };
+
+  const handleEditTodo = (todo: Todo) => {
+    setEditingTodoId(todo.id);
+    setEditingTodoTitle(todo.title);
+  };
+
+  const handleUpdateTodo = (event: React.FormEvent) => {
```
TarasHoliuk, line 133: I'd pass `todoId` as an argument (you can get rid of the editingTodoId state here and just create isEditing state in TodoItem component) when calling it in the TodoItem component

also, you can create a wrapper function there (in TodoItem) to prevent default event

src/App.tsx
```diff
@@ -1,26 +1,241 @@
...
+        setError('Unable to update a todo');
+      })
+      .finally(() => setLoading(false));
+  };
+
+  const handleCancelEdit = () => {
+    setEditingTodoId(null);
+  };
+
+  const handleClearCompleted = () => {
```
TarasHoliuk, line 174: It should work for real (delete todos on server)

src/components/header.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useEffect, useRef } from 'react';
+
+interface HeaderProps {
+  newTodo: string;
+  loading: boolean;
+  onAddTodo: (event: React.FormEvent) => void;
```
TarasHoliuk, line 6: Same as for update todo - this function should accept new todo title as an argument (state for it you can have here - in Header component)

src/components/todoList.tsx
```diff
@@ -0,0 +1,84 @@
...
+  editingTodoTitle,
+  loading,
+  onToggleTodo,
+  onDeleteTodo,
+  onEditTodo,
+  onUpdateTodo,
+  onEditingTodoTitleChange,
+  onCancelEdit,
+}) => {
+  const filteredTodos = todos.filter(todo => {
```
TarasHoliuk, line 34: Wrap it in useMemo

src/components/todoList.tsx
```diff
@@ -0,0 +1,84 @@
...
+          loading={loading}
+          onToggleTodo={onToggleTodo}
+          onDeleteTodo={onDeleteTodo}
+          onEditTodo={onEditTodo}
+          onUpdateTodo={onUpdateTodo}
+          onEditingTodoTitleChange={onEditingTodoTitleChange}
+          onCancelEdit={onCancelEdit}
+        />
+      ))}
+      {tempTodo && (
```
TarasHoliuk, line 67: Temp todo should always have loader

TarasHoliuk (changes requested):
Carefully read all requirements of the task

> I just copied the previous task & change 2 lines to solve this task, but this method is works

Some features works with such approach but not all, and not in the proper way
You should try to pass all tests

lerastarynets (changes requested):
Well done, now pls fix [failing tests](https://github.com/mate-academy/react_todo-app-add-and-delete/pull/1374/checks#step:5:37)

src/App.tsx
```diff
@@ -1,26 +1,259 @@
...
-    </section>
+    <div className="todoapp">
+      {!USER_ID ? (
+        <UserWarning />
+      ) : (
+        <>
+          <h1 className="todoapp__title">todos</h1>
+
+          <div className="todoapp__content">
+            {todos.length === 0 ? (
```
volodymyr-soltys97, line 213:
```suggestion
            {!todos.length ? (
```

src/components/errorNotification.tsx
volodymyr-soltys97: Not fixed

src/components/footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </a>
+
+        <a
+          href="#/completed"
+          className={`filter__link ${filter === 'completed' ? 'selected' : ''}`}
+          data-cy="FilterLinkCompleted"
+          onClick={() => onSetFilter(Filter.Completed)}
+        >
+          Completed
+        </a>
```
volodymyr-soltys97, line 55: Use `Object.values()` and render these options with `map()` method

src/components/header.tsx
```diff
@@ -0,0 +1,54 @@
...
+      onAddTodo(newTodo);
+      setNewTodo('');
+    }
+  };
+
+  return (
+    <header className="todoapp__header">
+      <button
+        type="button"
+        className={`todoapp__toggle-all ${allCompleted ? 'active' : ''}`}
```
volodymyr-soltys97, line 35: Use the classnames library rules for add classes with condition, fix it everywhere

volodymyr-soltys97 (changes requested):
Hi, you need to fix errors with tests, if you need help feel free in the fe_chat 
image

Component files must be named in uppercase letters
image

src/App.tsx
TarasHoliuk: Not fixed:
image

src/App.tsx
TarasHoliuk: Not fixed:
image

src/App.tsx
TarasHoliuk: Not fixed

src/App.tsx
TarasHoliuk: Not fixed:
image

src/components/TodoList_temp.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onToggleTodo,
+  onDeleteTodo,
+  onEditTodo,
+  onUpdateTodo,
+  onEditingTodoTitleChange,
+  onCancelEdit,
+}) => {
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      if (filter === Filter.All) {
```
TarasHoliuk, line 42: use switch block here

src/components/TodoItem_temp.tsx
```diff
@@ -0,0 +1,89 @@
...
+  onEditTodo,
+  onUpdateTodo,
+  onEditingTodoTitleChange,
+  onCancelEdit,
+}) => {
+  return (
+    <div
+      key={todo.id}
+      data-cy="Todo"
+      className={`todo ${todo.completed ? 'completed' : ''}`}
```
TarasHoliuk, line 33: use classnames library for conditional classes

TarasHoliuk (changes requested):
You haven't fixed any of the previous comments
Please, before re-requesting review make sure you've fixed all the previous comments

If you have any questions, need help - ask in the chat or join to a Q&A session. Do not ignore comments

Also, revert the last commit changes (_temp from files name)

P.S. If you want to do some test pushes/deploy you can do it, but no need to re-request review if you haven't fixed the previous comments

VitaliyBondarenko1982 (approved):
Well done.

PR #1381, opened 2024-08-08
solution 2 context

yaros-dev:
[DEMO LINK](https://yaros-dev.github.io/react_todo-app-add-and-delete/)


src/components/TodoListItem/TodoListItem.tsx
```diff
@@ -0,0 +1,26 @@
+import { FC } from 'react';
+import { TodoItem } from '../TodoItem/TodoItem';
+import { useTodoContext } from '../../utils/hooks/useTodoContext';
+import { TransitionGroup, CSSTransition } from 'react-transition-group';
+import { TempTodo } from '../TodoTemp/TodoTemp';
+
+export const TodoListItem: FC = () => {
```
TarasHoliuk, line 7:
```suggestion
export const TodoList: FC = () => {
```
ListItem sounds like one todo

src/services/FilterService.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const getCompletedTodos = (todos: Todo[]) => {
+  return filterTodos(todos, true);
+};
+
+export const isAllTodosComplete = (todos: Todo[]) => {
+  return todos.every(todo => todo.completed === true);
+};
+
+export const getOneCompletedTodo = (todos: Todo[]) => {
```
TarasHoliuk, line 20:
```suggestion
export const checkHasCompletedTodo = (todos: Todo[]) => {
```
functions with `get` should return something - if `getOneCompletedTodo` it should return todo

src/services/FilterService.ts
```diff
@@ -0,0 +1,33 @@
+import { Filters } from '../types/Filters/Filters';
```
TarasHoliuk, line 1: this file name shouldn't be in PascalCase but in camelCase. Pascal case for files with React components

Also, I'd rename it - helpers/utils instead of service (optionally)

src/services/TodoService.ts
```diff
@@ -0,0 +1,158 @@
+import { useEffect } from 'react';
+import { addTodos, delateTodos, patchTodos } from '../api/todos';
+import { useTodoContext } from '../utils/hooks/useTodoContext';
+import { ErrorMessages } from '../types/ErrorMessages/ErrorMessages';
+import { Todo } from '../types/Todo/Todo';
+
+export const TodoService = () => {
```
TarasHoliuk, line 7: you've made something similar to a component to be able to call hooks inside. But it's not a component and such an approach is wrong. Instead, make custom hook from it

```suggestion
export const useTodoActions = () => {
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+    } finally {
+      setLoading(null);
+      setTodoTemp(null);
+      setIsSubmitting(false);
+    }
+  };
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, [isSubmitting, inputRef]);
```
TarasHoliuk, line 61: you do it a lot of times

But I guess you can create a ref object here instead of Context and use it only here

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+      completed: false,
+      title: normalizeTitle,
+    };
+
+    try {
+      setIsSubmitting(true);
+      setLoading([todo.id]);
+      await createTodo(todo);
+      reset();
+    } catch (error) {
```
TarasHoliuk, line 51:
```suggestion
```
I guess you can just skip catch block

src/services/TodoService.ts
```diff
@@ -0,0 +1,158 @@
...
+  };
+
+  const createTodo = async ({ userId, completed, title }: Todo) => {
+    try {
+      const newTodo = await addTodos({ userId, completed, title });
+
+      setTodos(curenTodos => [...curenTodos, newTodo]);
+    } catch (error) {
+      showError(ErrorMessages.Add);
+      throw error;
```
TarasHoliuk, line 48: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

Fix everywhere

src/context/todoContext.tsx
```diff
@@ -0,0 +1,109 @@
+import React, {
```
TarasHoliuk, line 1: This file exports the React component so should be named in PascalCase - TodoContext.tsx

src/context/todoContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+  children: ReactNode;
+}
+
+export const TodoProvider: React.FC<ProviderProps> = ({
+  children,
+}): JSX.Element => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState(Filters.All);
+  const [errorMessage, setErrorMessage] = useState<ErrorMessages | string>('');
+  const [loading, setLoading] = useState<number[] | null>(null);
```
TarasHoliuk, line 50:
```suggestion
  const [loadingTodoIds, setLoadingTodoIds] = useState<number[] | null>(null);
```
to describe value better

src/context/todoContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const [editTodoId, setEditTodoId] = useState<number | null>(null);
+  const [focusInput, setFocusInput] = useState<boolean | null>(false);
```
TarasHoliuk, line 52: redundant state - it should depend on other values (todos/tempTodo, etc.). But ref shouldn't exist in this file at all

TarasHoliuk (changes requested):
You did really great job!

I left comments about naming and a few comments about general approach to some things to refactor your code a bit 🙂

src/App.tsx
```diff
@@ -1,26 +1,28 @@
...
-      </p>
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+      <div className="todoapp__content">
+        <Header />
+        <TodoList />
+        {todos.length > 0 && <TodoFilter />}
```
volodymyr-soltys97, line 23:
```suggestion
        {!!todos.length && <TodoFilter />}
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,91 @@
...
+import classNames from 'classnames';
+import { useFocus } from '../../utils/hooks/useFocus';
+
+export const Header: FC = ({}) => {
+  const { showError, setLoadingTodoIds, todos, setTodoTemp, setLockedFocus } =
+    useTodoContext();
+  const { createTodo, toggleAllCompleted } = useTodoActions();
+  const { inputRef } = useFocus();
+  const [title, setTitle] = useState('');
+  const [isSubmitting, setIsSubmitting] = useState(false); // - тут стан для фокусу
```
volodymyr-soltys97, line 18:
```suggestion
  const [isSubmitting, setIsSubmitting] = useState(false);
```

volodymyr-soltys97 (changes requested):
Almost done!
Let's make your code better

etojeDenys (approved):

PR #1386, opened 2024-08-17
add TodoList full functionality

OlehPhw:
[DEMO LINK](https://OlehPhw.github.io/react_todo-app-add-and-delete/)

1 test is failing because I use async function, but functionality works as expected

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -1,0 +1,32 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 1201;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+export const addTodo = ({ userId, title, completed }: Omit<Todo, 'id'>) => {
+  return client.post<Todo>(`/todos?userId=${USER_ID}`, {
+    userId,
+    title,
+    completed,
+  });
+};
+
+export const deleteTodo = (userId: number) => {
+  return client.delete(`/todos/${userId}`);
+};
+
+export const updateTodoTitle = (todoId: number, newTitle: string) => {
+  return client.patch(`/todos/${todoId}`, {
+    title: newTitle,
+  });
+};
+
+export const updateTodoStatus = (todoId: number, newStatus: boolean) => {
+  return client.patch(`/todos/${todoId}`, {
+    completed: newStatus,
+  });
+};
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,0 +1,139 @@
+import React, { useRef, useState, useEffect } from 'react';
+import { Todo } from '../../types/Todo';
+import { addTodo, updateTodoStatus, USER_ID } from '../../api/todos';
+import cn from 'classnames';
+
+type Props = {
+  setErrorMessage: React.Dispatch<React.SetStateAction<string>>;
+  setTempTodo: React.Dispatch<React.SetStateAction<Todo | null>>;
+  setTodos: React.Dispatch<React.SetStateAction<Todo[]>>;
+  todos: Todo[];
+  errorMessage: string;
+  setLoadingTodosId: React.Dispatch<React.SetStateAction<number[]>>;
+  loadingTodosId: number[];
+  addIdToLoad: (arg: number) => void;
+};
+
+export const Header: React.FC<Props> = ({
+  setErrorMessage,
+  setTempTodo,
+  setTodos,
+  todos,
+  errorMessage,
+  setLoadingTodosId,
+  loadingTodosId,
+  addIdToLoad,
+}) => {
+  const [query, setQuery] = useState('');
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const handleOnSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    setErrorMessage('');
+
+    if (!query.trim()) {
+      return setErrorMessage('Title should not be empty');
+    }
+
+    const forFetchTodo: Omit<Todo, 'id'> = {
+      userId: USER_ID,
+      title: query.toString().trim(),
+      completed: false,
+    };
+
+    setTempTodo({
+      ...forFetchTodo,
+      id: 0,
+    });
+
+    addIdToLoad(0);
+
+    try {
+      const addedTodo = await addTodo(forFetchTodo);
+
+      setTodos(prevTodos => [...prevTodos, addedTodo]);
+      setTempTodo(null);
+      setQuery('');
+      inputRef.current?.focus();
+    } catch (error) {
+      setErrorMessage('Unable to add a todo');
+      setTempTodo(null);
+      throw error;
+    } finally {
+      setLoadingTodosId([]);
+      inputRef.current?.focus();
+    }
+  };
+
+  const handleOnClick = async () => {
+    let activeTodosId: number[] = [];
+
+    try {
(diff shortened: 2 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+  useEffect(() => {
+    const handleKeyUp = (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
+        event.preventDefault();
+        cancelEditing();
+      }
+    };
+
+    document.addEventListener('keyup', handleKeyUp);
```
TarasHoliuk, line 52: you can add this listener to the input in common for React way, like
`onKeyUp={handleKeyUp}`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      try {
+        const fetchedTodos = await getTodos();
+
+        setTodos(fetchedTodos);
+      } catch (error) {
+        setErrorMessage('Unable to load todos');
+        throw error;
```
TarasHoliuk, line 67: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,247 @@
...
+            <form onSubmit={handleOnSubmit}>
+              <input
+                data-cy="TodoTitleField"
+                type="text"
+                className="todo__title-field"
+                placeholder="Empty todo will be deleted"
+                ref={inputRef}
+                value={query}
+                onChange={handleOnChange}
+                onBlur={handleOnBlur}
```
TarasHoliuk, line 228:
```suggestion
                onBlur={saveChanges}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,247 @@
...
+      setLoadingTodosId([]);
+    }
+  };
+
+  const saveChanges = async () => {
+    if (!query && editingTodo !== null) {
+      addIdToLoad(editingTodo.id);
+      try {
+        await deleteTodo(editingTodo.id);
+        setTodos(todos.filter(todo => editingTodo.id !== todo.id));
```
TarasHoliuk, line 92:
```suggestion
        setTodos((todos) => todos.filter(todo => editingTodo.id !== todo.id));
```
When you use the current state to set a new one it's always better to pass the callback function instead of taking the current state from the scope: [doc](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state)

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,247 @@
...
+            className={cn('modal', 'overlay', {
+              'is-active': loadingTodosId.includes(todo.id),
+            })}
+          >
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
+      ))}
+      {tempTodo !== null && <TodoItem tempTodo={tempTodo} />}
```
TarasHoliuk, line 244:
```suggestion
      {tempTodo && <TodoItem tempTodo={tempTodo} />}
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,139 @@
...
+  setErrorMessage,
+  setTempTodo,
+  setTodos,
+  todos,
+  errorMessage,
+  setLoadingTodosId,
+  loadingTodosId,
+  addIdToLoad,
+}) => {
+  const [query, setQuery] = useState('');
```
TarasHoliuk, line 27: it's not query (of some filter, etc.) it's title, or just input

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,139 @@
...
+    try {
+      const addedTodo = await addTodo(forFetchTodo);
+
+      setTodos(prevTodos => [...prevTodos, addedTodo]);
+      setTempTodo(null);
+      setQuery('');
+      inputRef.current?.focus();
+    } catch (error) {
+      setErrorMessage('Unable to add a todo');
+      setTempTodo(null);
```
TarasHoliuk, line 65: in finally

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,139 @@
...
+    }
+  };
+
+  const handleOnClick = async () => {
+    let activeTodosId: number[] = [];
+
+    try {
+      if (todos.some(todo => !todo.completed)) {
+        activeTodosId = todos
+          .filter(todo => !todo.completed)
```
TarasHoliuk, line 79: do it before and in `if` just check the length instead of doing `todos.some`

TarasHoliuk (changes requested):
You did a great job!

I added some comments to improve your code a bit 🙂 
About tests: not only 1 test failed - all other tests were skipped after 1 failed, so try to fix tests (async function cannot be the reason for it)

image

OlehPhw:
onKeyUp={handleKeyUp}

Didn't change to this, because If I do like this I have the same event happens both onBlur and onKeyUp. So I fixed this problem with document.addEventListener. Also read that it is better to have 1 eventListener on doc than many on every Todo

src/api/todos.ts
```diff
@@ -0,0 +1,32 @@
...
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 1201;
+
+export const getTodos = () => {
+  return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
+};
+
+export const addTodo = ({ userId, title, completed }: Omit<Todo, 'id'>) => {
+  return client.post<Todo>(`/todos`, {
```
etojeDenys, line 11:
```suggestion
  return client.post<Todo>('/todos', {
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,106 @@
...
+      const resolvedIds = res.reduce(
+        (acc, item) => {
+          if (item.status === 'fulfilled' && item.value.status === 'resolved') {
+            return { ...acc, [item.value.id]: item.value.id };
+          }
+
+          return acc;
+        },
+        {} as Record<number, number>,
+      );
```
etojeDenys, line 56:
```suggestion
      const resolvedIds = res.reduce<Record<number, number>>(
        (acc, item) => {
          if (item.status === 'fulfilled' && item.value.status === 'resolved') {
            return { ...acc, [item.value.id]: item.value.id };
          }

          return acc;
        },
        {},
      );
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,106 @@
...
+      {todos.length !== 0 && (
+        <footer className="todoapp__footer" data-cy="Footer">
+          <span className="todo-count" data-cy="TodosCounter">
+            {todos.filter(todo => !todo.completed).length} items left
+          </span>
+
+          <nav className="filter" data-cy="Filter">
+            {Object.values(FilterMethods).map(method => {
+              return (
+                <a
```
etojeDenys, line 79:
```suggestion
            {Object.values(FilterMethods).map(method => (
                <a
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+type Props = {
+  setTodos: React.Dispatch<React.SetStateAction<Todo[]>>;
+  setErrorMessage: React.Dispatch<React.SetStateAction<string>>;
+  filteredTodos: Todo[];
+  todos: Todo[];
+  tempTodo: Todo | null;
+  setLoadingTodosId: React.Dispatch<React.SetStateAction<number[]>>;
+  loadingTodosId: number[];
+  addIdToLoad: (arg: number) => void;
```
etojeDenys, line 20: addIdToLoad and setLoadingTodosId are almost the same functions, remove one function

etojeDenys (approved):

PR #1398, opened 2024-08-26
Develop

RimmaBuzuluk:
[DEMO LINK](https://RimmaBuzuluk.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,6 +10,7 @@ import * as tadoService from './api/todos';
 import { Footer } from './component/Footer';
 import { Error } from './component/Error';
 import { ListComponent } from './component/ListComponent';
+import { ErrorMessages, errorMessages } from './types/err';
 
 function filterTodos(todos: Todo[], filter: FilterStatusType) {
   switch (filter) {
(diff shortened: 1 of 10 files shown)


YaroslavKolesnyk (approved):

src/App.tsx
```diff
@@ -1,26 +1,182 @@
...
+
+  const handleError = (message: ErrorMessages) => {
+    setErrorMessage(message);
+
+    setTimeout(() => {
+      setErrorMessage(null);
+    }, 3000);
+  };
+
+  useEffect(() => {
```
ArturErmolenko, line 48: where is loading?

src/App.tsx
```diff
@@ -1,26 +1,182 @@
...
+  const handleQueryChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const handleAddTodo = (event: KeyboardEvent<HTMLInputElement>): void => {
+    if (event.key === 'Enter') {
+      event.preventDefault();
+
+      if (query.trim() === '') {
+        handleError('Title should not be empty');
```
ArturErmolenko, line 88: let's create enum with error messages and use its values instead of magic string

src/types/err.ts
```diff
@@ -0,0 +1,9 @@
+export const errorMessages = {
+  load: 'Unable to load todos',
+  title: 'Title should not be empty',
+  add: 'Unable to add a todo',
+  delete: 'Unable to delete a todo',
+  update: 'Unable to update a todo',
+} as const;
```
ArturErmolenko, line 7: ok you already has const, let's add title error as well

ArturErmolenko (approved):

PR #1403, opened 2024-08-27
add task solution

mykhailo-shvets:
[DEMO LINK](https://mykhailo-shvets.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,74 @@
+import React from 'react';
+import { Todo } from '../types/Todo';
+import { Status } from '../types/Status';
+import cn from 'classnames';
+
+type Props = {
+  todos: Todo[];
+  status: Status;
+  setStatus: (status: Status) => void;
+  deleteCompletedTodo: () => void;
+};
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  status,
+  setStatus,
+  deleteCompletedTodo: deleteCompletedTodo,
+}) => {
+  const todosLeft = todos.filter(todo => !todo.completed).length;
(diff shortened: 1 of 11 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+        </a>
+
+        <a
+          href="#/completed"
+          className={cn('filter__link', {
+            selected: status === Status.Completed,
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setStatus(Status.Completed)}
+        >
```
volodymyr-soltys97, line 58: Use `Object.values(your enum)` and render these options with `map()` method

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <span data-cy="TodoTitle" className="todo__title">
+        {title}
+      </span>
+      <button
+        type="button"
+        className="todo__remove"
+        data-cy="TodoDelete"
+        onClick={() => {
+          deleteTodo(id);
+          setLoadingTodo(todo);
```
volodymyr-soltys97, line 35: Move this logic from jsx to the helper variable and use it here

volodymyr-soltys97 (changes requested):
Good job! 👍 
Let's make your code better

src/components/TodoItem.tsx
TarasHoliuk: Not fixed

image

please, fix it in the next stage task:
```suggestion
        onClick={handleDelete}
```
create handle delete function above

TarasHoliuk (approved):

PR #1420, opened 2024-09-07
add task solution

kravets111:
[DEMO LINK](https://kravets111.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,197 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { USER_ID } from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorMessage } from './components/ErrorMessage';
+import { Todo } from './types/Todo';
+import * as todoService from './api/todos';
+import { Options } from './types/Options';
+import { Errors } from './types/Errors';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [selectedOption, setSelectedOption] = useState<Options>(Options.All);
+  const [focus, setFocus] = useState(false);
+
+  const filterTodos = (posts: Todo[], option: Options) => {
+    switch (option) {
+      case Options.Active:
+        return posts.filter(post => !post.completed);
+
+      case Options.Completed:
+        return posts.filter(post => post.completed);
+
+      default:
+        return posts;
+    }
+  };
+
+  const todoFilterSelectedOptions = filterTodos(todos, selectedOption);
+  const todoFilterOptionActive = filterTodos(todos, Options.Active);
+
+  const clearError = useCallback(() => {
+    setErrorMessage(null);
+  }, []);
+
+  const showError = useCallback(
+    (error: Errors) => {
+      setErrorMessage(error);
+
+      setTimeout(() => {
+        clearError();
+      }, 3000);
+    },
+    [clearError],
+  );
+
+  useEffect(() => {
+    setFocus(true);
+
+    todoService
+      .getTodos()
+      .then(setTodos)
+      .catch(() => showError(Errors.LoadTodos));
+  }, [showError]);
+
+  const addTodo = useCallback(
+    async ({ userId, title, completed }: Omit<Todo, 'id'>) => {
+      const newTempTodo: Todo = {
+        id: 0,
+        userId,
+        title,
+        completed,
+      };
+
+      setTempTodo(newTempTodo);
+
+      try {
+        const newTodo = await todoService.createTodos({
+          userId,
+          title,
+          completed,
+        });
+
+        setTodos(currentTodo => [...currentTodo, newTodo]);
+      } catch {
+        showError(Errors.AddTodo);
+        throw new Error();
+      } finally {
+        setTempTodo(null);
+        setFocus(true);
+      }
+    },
+    [showError],
+  );
+
+  const deleteTodo = useCallback(
+    async (todoId: number) => {
+      try {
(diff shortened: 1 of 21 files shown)


src/App.tsx
```diff
@@ -1,26 +1,197 @@
...
+
+  const deleteTodo = useCallback(
+    async (todoId: number) => {
+      try {
+        await todoService.deleteTodos(todoId);
+        setTodos(currentTodo => currentTodo.filter(todo => todo.id !== todoId));
+        setFocus(true);
+      } catch {
+        showError(Errors.DeleteTodo);
+        throw new Error();
```
TarasHoliuk, line 100: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

src/App.tsx
```diff
@@ -1,26 +1,197 @@
...
+import { Errors } from './types/Errors';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [selectedOption, setSelectedOption] = useState<Options>(Options.All);
+  const [focus, setFocus] = useState(false);
+
+  const filterTodos = (posts: Todo[], option: Options) => {
```
TarasHoliuk, line 22: you can move this function out of the component or into another file and import it. Because now it recreates on every render

TarasHoliuk, line 22:
```suggestion
  const filterTodos = (todos: Todo[], option: Options) => {
```

src/App.tsx
```diff
@@ -1,26 +1,197 @@
...
+
+      case Options.Completed:
+        return posts.filter(post => post.completed);
+
+      default:
+        return posts;
+    }
+  };
+
+  const todoFilterSelectedOptions = filterTodos(todos, selectedOption);
```
TarasHoliuk, line 35:
```suggestion
  const filteredTodos = filterTodos(todos, selectedOption);
```
or even visibleTodos

because from `todoFilterSelectedOptions` variable name is difficult to predict what exactly is stored there - filter value/some options?

Similar for the variable below - just activeTodos

P.S. use useMemo hook

src/App.tsx
```diff
@@ -1,26 +1,197 @@
...
+        id: 0,
+        userId,
+        title,
+        completed,
+      };
+
+      setTempTodo(newTempTodo);
+
+      try {
+        const newTodo = await todoService.createTodos({
```
TarasHoliuk, line 74:
```suggestion
        const newTodo = await todoService.createTodo({
```
it works with one todo

fix similar cases

src/App.tsx
```diff
@@ -1,26 +1,197 @@
...
+
+  const toggleCompleted = useCallback((todoId: number) => {
+    setTodos(currentTodo =>
+      currentTodo.map(todo =>
+        todo.id === todoId ? { ...todo, completed: !todo.completed } : todo,
+      ),
+    );
+  }, []);
+
+  // function updateTodo(updatedTodo: Todo) {
```
TarasHoliuk, line 137: remove comments

It's nice you already have the solution for the next stage, but add it in the next task then or uncomment here)

src/App.tsx
```diff
@@ -1,26 +1,197 @@
...
+          todos={todos}
+          showError={showError}
+          onAddTodo={addTodo}
+          onToogleAll={onToogleAll}
+          focusInput={focus}
+          setFocusInput={setFocus}
+        />
+
+        <TodoList
+          filteredTodo={todoFilterSelectedOptions}
```
TarasHoliuk, line 177:
```suggestion
          filteredTodos={todoFilterSelectedOptions}
```
it's an array

src/App.tsx
```diff
@@ -1,26 +1,197 @@
...
+          onToggleComplete={toggleCompleted}
+        />
+
+        {!!todos.length && (
+          <Footer
+            todos={todos}
+            selected={selectedOption}
+            setSelected={setSelectedOption}
+            onClearCompleted={clearCompleted}
+            activeTodo={todoFilterOptionActive}
```
TarasHoliuk, line 189:
```suggestion
            activeTodos={todoFilterOptionActive}
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,114 @@
...
+      inputRef.current.focus();
+      setFocusInput(false);
+    }
+  }, [isLoading, focusInput, setFocusInput]);
+
+  useEffect(() => {
+    setButtonVisible(todos.length > 0);
+  }, [todos]);
+
+  const toggleCheckbox = useMemo(() => {
```
TarasHoliuk, line 78:
```suggestion
  const areAllTodosCompleted = useMemo(() => {
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,77 @@
...
+          onClick={() => handleDelete(todo.id)}
+          disabled={isDeleting}
+        >
+          ×
+        </button>
+      )}
+
+      <Loader isTemp={isDeleting} />
+
+      {isTemp && <Loader isTemp={isTemp} />}
```
TarasHoliuk, line 74: Now you use separate Loader to handle each case
Instead, you can do this:

```suggestion
      <Loader isLoading={isTemp || isDeleting} />
```

Or you can make Loader always active and render it like this:

```
{(isTemp || isDeleting) && <Loader />}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,39 @@
...
+          onDelete={onDeleteTodo}
+          onToggleComplete={onToggleComplete}
+        />
+      ))}
+
+      {tempTodo && (
+        <TodoItem
+          todo={tempTodo}
+          onDelete={() => Promise.resolve()}
+          isTemp={true}
```
TarasHoliuk, line 33:
```suggestion
          isTemp
```
for boolean props

src/App.tsx
```diff
@@ -1,26 +1,197 @@
...
+  const onToogleAll = (completed: boolean) => {
+    const updatedTodos = todos.map(todo => ({
+      ...todo,
+      completed,
+    }));
+
+    setTodos(updatedTodos);
+  };
+
+  const clearCompleted = useCallback(() => {
```
TarasHoliuk, line 115: you should show loader for each todo you delete

TarasHoliuk (changes requested):
GJ 👍 

One thing should be improved
Also, added some comments to improve your code (mostly about naming)

PR #1430, opened 2024-09-20
solution for part 2

syavaYki:
[DEMO LINK](https://syavaYki.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,159 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useCallback, useEffect, useState } from 'react';
 
-const USER_ID = 0;
+import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+
+import { USER_ID } from './api/todos';
+import { todos as todosClient } from './api/todos';
+import { ErrorNotification } from './Components/ErrorNotification';
+import { TodoHead } from './Components/TodoHead';
+import { TodoList } from './Components/TodoList';
+import { TodoFilter } from './Components/TodoFilter';
+import { Errors } from './types/Errors';
+import { Filters } from './types/Filters';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [filter, setFilter] = useState<Filters>(Filters.all);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,26 +1,159 @@
...
+
+      case Filters.active:
+        setFilteredTodos(todos.filter(todo => !todo.completed));
+        break;
+
+      case Filters.all:
+      default:
+        setFilteredTodos(todos);
+        break;
+    }
```
etojeDenys: move this logic to some function and return the filtered array from that function

src/App.tsx
```diff
@@ -1,26 +1,159 @@
...
+import { TodoList } from './Components/TodoList';
+import { TodoFilter } from './Components/TodoFilter';
+import { Errors } from './types/Errors';
+import { Filters } from './types/Filters';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [filter, setFilter] = useState<Filters>(Filters.all);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
etojeDenys: and after that you can just remove this state and make it the simple var
```suggestion
  const filteredTodos = getFilteredTodos();
```

src/Components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,39 @@
...
+            todo={todo}
+            loadingId={loadingId}
+            onDelete={onDelete}
+          />
+        );
+      })}
+
+      {tempTodoTitle && (
+        <TodoItem
+          key={0}
```
etojeDenys: key is redundant here

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,26 +1,158 @@
...
+        })
+        .catch(() => {
+          handleErrors(Errors.add);
+        })
+        .finally(() => {
+          setInputLoadingFlag(false);
+          setTempTodo(null);
+        });
+    },
+    [todos],
```
SanyaBratashchuk:
```suggestion
    [],
```

src/App.tsx
```diff
@@ -1,26 +1,158 @@
...
+        })
+        .catch(() => {
+          handleErrors(Errors.delete);
+        })
+        .finally(() => {
+          setLoadingTodoId(() => null);
+          setInputResetFlag(Math.random());
+        });
+    },
+    [todos],
```
SanyaBratashchuk:
```suggestion
    [],
```

SanyaBratashchuk (changes requested):
fix linter errors

SanyaBratashchuk (approved):

PR #1440, opened 2024-09-30
Develop

modeltoIT:
[DEMO LINK](https://modeltoIT.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -47,7 +47,6 @@ export const App: React.FC = () => {
       removeErrorMessage();
     }, 3000);
   }, []);
-
   // <br />
   // Unable to update a todo
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,18 +6,15 @@ import { Header } from './components/Header';
 import { TodoList } from './components/TodoList';
 import { Footer } from './components/Footer';
 import { ErrorMessage } from './components/ErrorMessage';
+import { TodoFilter } from './types/TodoFilter';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [fetchedTodos, setFetchedTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
   const [errorMessage, setErrorMessage] = useState('');
-  const [filterBy, setFilterBy] = useState<'all' | 'active' | 'completed'>(
-    'all',
-  );
+  const [filterBy, setFilterBy] = useState<TodoFilter>('all');
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
-  const [completedTodoIds, setCompletedTodoIds] = useState<number[] | null>(
-    null,
-  );
+  const [idsToDelete, setIdsToDelete] = useState<number[]>([]);
   const [isHiddenError, setIsHiddenError] = useState(true);
 
   const timeoutRef = useRef<number | null>(null);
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,26 +1,125 @@
...
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorMessage } from './components/ErrorMessage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [fetchedTodos, setFetchedTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterBy, setFilterBy] = useState<'all' | 'active' | 'completed'>(
```
TarasHoliuk: use TodoFilter type here

src/App.tsx
```diff
@@ -1,26 +1,125 @@
...
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+    }
+
+    timeoutRef.current = window.setTimeout(() => {
+      removeErrorMessage();
+    }, 3000);
+  }, []);
+  // <br />
+  // Unable to update a todo
```
TarasHoliuk: delete comments

src/App.tsx
```diff
@@ -1,26 +1,125 @@
...
+  useEffect(() => {
+    let filteredTodos = todos;
+
+    if (filterBy !== 'all') {
+      filteredTodos = todos.filter(todo =>
+        filterBy === 'active' ? !todo.completed : todo.completed,
+      );
+    }
+
+    setFetchedTodos([...filteredTodos]);
```
TarasHoliuk: why fetchedTodos if it's filtered todos?

But you don't need this state at all. It depends on other state values, so re-render will happen anyway - create a common variable `const visibleTodos = useMemo(...` and use useMemo hook to recalculate it only when it's necessary

src/components/TodoList.tsx
```diff
@@ -0,0 +1,100 @@
...
+  onError: (error: Error) => void;
+  setForClear: (todoIds: number[] | null) => void;
+  handleDelete: (val: number) => void;
+}
+
+export const TodoList: React.FC<Props> = ({
+  renderedList,
+  tempTodo,
+  onError,
+  forClear,
```
TarasHoliuk: it's not clear what exactly it is - rename to smth like `todoIdsToDelete`

src/App.tsx
```diff
@@ -1,26 +1,125 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [fetchedTodos, setFetchedTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterBy, setFilterBy] = useState<'all' | 'active' | 'completed'>(
+    'all',
+  );
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [completedTodoIds, setCompletedTodoIds] = useState<number[] | null>(
```
TarasHoliuk: There might be completed todos, and it's not these todos - i.e. this name is misleading
use `todoIdsToDelete`

use empty array as initial value - that's common way to do it for arrays in state

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import * as Action from '../api/todos';
```
TarasHoliuk:
```suggestion
import * as action from '../api/todos';
```
camelCase for common variables (PascalCase for classes, components, types)

src/App.tsx
```diff
@@ -1,26 +1,125 @@
...
+          errorMessage={errorMessage}
+        />
+
+        <TodoList
+          renderedList={fetchedTodos}
+          tempTodo={tempTodo}
+          todos={todos}
+          forClear={completedTodoIds}
+          setForClear={setCompletedTodoIds}
+          onDelete={setTodos}
```
TarasHoliuk: it looks strange
define delete function here and pass it (that might allow you not to pass `completedTodoIds` variable and handle clear completed here)

or at least change prop name

TarasHoliuk (changes requested):

VitaliyBondarenko1982 (approved):
GJ

PR #1444, opened 2024-10-03
Develop

YarikKedrych:
[DEMO LINK](https://YarikKedrych.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -42,6 +42,7 @@ export const App: React.FC = () => {
       .getTodos()
       .then(setTodos)
       .catch(() => showError(Load));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   const setTodoLoading = (id: number, loading: boolean) => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,161 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
-
+import React, { useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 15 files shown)


totum-revolutum (approved):
It looks complicated, but the main thing is that it works! You are doing great!

src/App.tsx
```diff
@@ -1,26 +1,168 @@
...
+      setErrorMassage(None);
+    }, 3000);
+  };
+
+  useEffect(() => {
+    todoService
+      .getTodos()
+      .then(setTodos)
+      .catch(() => showError(Load));
+  }, []);
```
dothearts, line 46: You can disabled eslint in this case

src/components/Footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+interface Props {
+  setFilter: Dispatch<SetStateAction<FilterType>>;
+  filter: string;
+  todos: Todo[];
+  onDelete: (userId: number) => Promise<void>;
+}
+
+export const Footer: FC<Props> = ({ setFilter, filter, todos, onDelete }) => {
+  const handleActiveTodos = todos.reduce((acc, val) => {
```
dothearts: It's not handler. It count of active todo. 
Also, you can use filter in this case, that returns you all completed, then use .length for get count.

src/components/Header.tsx
```diff
@@ -0,0 +1,83 @@
...
+      showError(ErrorMessages.Add);
+      inputRef.current?.focus();
+    }
+  };
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [isLoading]);
```
dothearts, line 56: You can disabled eslint in this case

DanilWeda (changes requested):
Good job. All great! 
Check some small minor issues
Many thanks!)

DanilWeda (approved):
Good job. 
Approved. 
Many thanks)

PR #1448, opened 2024-10-05
solution

pasha1932:
[DEMO LINK](https://pasha1932.github.io/react_todo-app-add-and-delete/)

I did all that was described at this task, all functionality I added in the last part of todo-task, and i have a question Did I do everything right with the handleDeleteCompleted function or should I have used allSettled?

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,14 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { FormEvent, useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { addTodo, deleteTodo, getTodos, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
-import classNames from 'classnames';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
 
-enum Status {
+export enum Status {
   all = 'all',
   active = 'active',
   completed = 'completed',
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,425 @@
...
+      )
+      .catch(error => {
+        setTodos(todos);
+        setErrorMessage('Unable to delete a todo');
+        throw error;
+      })
+      .finally(() => setIsLoading(0));
+  };
+
+  const handleDeleteCompleted = () => {
```
anastasiiavorobiova, line 135: It's better to use Promise.allSettled here and proceed results

anastasiiavorobiova (changes requested):
Great work! Try to pass all tests

src/App.tsx
pasha1932: > It's better to use Promise.allSettled here and proceed results

through the loop, push all promises into some array and then write Promise.allSettled(...array)?

src/App.tsx
```diff
@@ -1,26 +1,425 @@
...
+    for (let i = 0; i < completedTodos.length; i++) {
```
volodymyr-soltys97: Use `forEach` method instead of for loop

volodymyr-soltys97 (changes requested):
Hi, you can read more here about this method and how to use it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

It is a bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoList, TodoItem`

pasha1932:
image
why test  1) should send a create request don`t pass?

src/App.tsx
```diff
@@ -1,26 +1,200 @@
...
+        })
+        .finally(() => setIsLoading(0));
+    });
+
+    Promise.allSettled(promises); // З
+    // completedTodos.forEach(el => )
+    //   const el = completedTodos[i];
+
+    //   setIsLoading(el);
+    //   deleteTodo(el)
```
volodymyr-soltys97: Remove all comments

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+          })}
+          data-cy="FilterLinkActive"
+          onClick={() => onStatus(Status.active)}
+        >
+          Active
+        </a>
+
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
```
volodymyr-soltys97: Use `Object.values(your enum)` and render these options with `map()` method

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+        />
+      </label>
+
+      <span data-cy="TodoTitle" className="todo__title">
+        {todo.title}
```
volodymyr-soltys97: Use destructuring for `todo`

volodymyr-soltys97 (changes requested):
GJ!
Let's improve your code

lerastarynets (approved):
You're making excellent progress!

PR #1450, opened 2024-10-08
Develop

IrbisKronos:
[DEMO LINK](https://IrbisKronos.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ Take your code implemented for [Loading todos](https://github.com/mate-academy/r
 and implement the ability to add and remove todos.
 
 > Here is [the working example](https://mate-academy.github.io/react_todo-app-with-api/)
+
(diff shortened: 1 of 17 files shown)


anastasiiavorobiova (changes requested):
Make sure to deploy your solution

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix this failed test. when we add a new todo, we have loaders displayed in all todos, this is wrong
image
image

etojeDenys (changes requested):
1. show loader only on todo that user tries to delete
2. see in working example how the temp logic should work, or feel free to ask for some help in the chat how to implement it

you can always ping me in the thread and I will try to help you

PR #1461, opened 2024-10-15
Solution

KrivorotkoYurii:
[DEMO LINK](https://KrivorotkoYurii.github.io/react_todo-app-add-and-delete/)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  isBeingSaved: boolean;
+  tempTodo: Todo;
+}
+
+export const TodoItem: React.FC<Props> = ({ isBeingSaved, tempTodo }) => {
```
Ann-Maksay: I'm not sure it's the best idea to create a separate component like this and not use it when rendering the list. These elements have too much in common and I would rather make them one component and use optional props.

src/App.tsx
```diff
@@ -1,26 +1,88 @@
...
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState(Filter.ALL);
+  const [inputChange, setInputChange] = useState('');
+  const [isBeingSaved, setIsBeingSaved] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [focusOnChange, setFocusOnChange] = useState(true);
+  const [isBeingDeleted, setIsBeingDeleted] = useState<number>(0);
```
Ann-Maksay: Does it work as expected when deleting several items at the same time?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,78 @@
...
+  setTodos,
+  setFocusOnChange,
+  setErrorNotification,
+}) => {
+  const activeTodosCount = todos.filter(todo => !todo.completed).length;
+
+  const linksValues = Object.values(Filter);
+
+  const handleDeleteCompletedTodos = () => {
+    const completedTodos: Todo[] = todos.filter(todo => todo.completed);
```
Ann-Maksay, line 26: Do we really need a whole object here?

src/handlers/handleSubmit.ts
```diff
@@ -0,0 +1,51 @@
...
+  setIsBeingSaved: React.Dispatch<React.SetStateAction<boolean>>,
+  setTodos: React.Dispatch<React.SetStateAction<Todo[]>>,
+  onInputChange: React.Dispatch<React.SetStateAction<string>>,
+  setFocusOnChange: React.Dispatch<React.SetStateAction<boolean>>,
+) => {
+  if (!inputChange) {
+    handleError(Errors.EMPTY, setErrorNotification);
+
+    return;
+  } else {
```
Ann-Maksay: Do we need an `else` block when we already exit the function when an error occurs?

Ann-Maksay (approved):
Great job 🔥

src/App.tsx
BudnikOleksii: > Does it work as expected when deleting several items at the same time?

Good point, it is better to create a state with `loadingTodoIds` so you can store them in array and depending on it show the loader for corresponding todos instead of isBeingSaved/isBeingDeleted

src/components/Footer/Footer.tsx
BudnikOleksii: > Do we really need a whole object here?

Good point, it looks like IDs are enough

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    handleSubmit(
+      inputChange,
+      setErrorNotification,
+      setTempTodo,
+      setIsBeingSaved,
+      setTodos,
+      onInputChange,
+      setFocusOnChange,
+    );
```
BudnikOleksii: Why?
You should send only form values, the form shouldn't keep other logic

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,102 @@
...
+      setFocusOnChange,
+    );
+  };
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {filteredTodos.map(todo => (
+        <div
+          key={todo.id}
+          data-cy="Todo"
```
BudnikOleksii: You have a todo component, use it

src/handlers/handleSubmit.ts
BudnikOleksii: > Do we need an `else` block when we already exit the function when an error occurs?

Good point

BudnikOleksii (changes requested):
Good work 👍 
Check my and a previous reviewer suggestions and improve your solution

BudnikOleksii (approved):
Looks good 🔥

PR #1463, opened 2024-10-16
solution

YuliiaShatkovska:
[DEMO LINK](https://YuliiaShatkovska.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,7 +22,7 @@ export const App: FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [isAddingTodo, setIsAddingTodo] = useState(false);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
-  const [todoToDelete, setTodoToDelete] = useState<number[]>([]);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
 
   const [errorMessage, setErrorMessage] = useState(Errors.DEFAULT);
   const [filterStatus, setFilterStatus] = useState(FilterStatus.ALL);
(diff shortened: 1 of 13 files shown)


src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
```
PikalovAndrey, line 2: Also, the constant must be in a separate file.

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  if (data) {
+    // We add body and Content-Type only for the requests with data
+    options.body = JSON.stringify(data);
+    options.headers = {
+      'Content-Type': 'application/json; charset=UTF-8',
+    };
+  }
+
+  // DON'T change the delay it is required for tests
```
PikalovAndrey: I think you can delete these comments, you won't need them in the future

src/api/todos.ts
```diff
@@ -0,0 +1,16 @@
+import { Todo } from '../types/Todo';
+import { client } from '../utils/fetchClient';
+
+export const USER_ID = 1595;
```
PikalovAndrey, line 4: This constant should be in a separate file.

PikalovAndrey (approved):
Overall very good code and structure, just check a few little things)

src/App.tsx
```diff
@@ -1,26 +1,133 @@
...
+
+import { Todo } from './types/Todo';
+import { Errors } from './types/Errors';
+import { FilterStatus } from './types/FilterStatus';
+
+export const App: FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isAddingTodo, setIsAddingTodo] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todoToDelete, setTodoToDelete] = useState<number[]>([]);
```
BudnikOleksii:
```suggestion
  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
```
It stores IDs, not todos as I see, and it's probably better to have a more general name as you can also change status of todo as well, not only delete

src/utils/fetchClient.ts
BudnikOleksii: > I think you can delete these comments, you won't need them in the future

Good point

BudnikOleksii (approved):
Looks good 👍 
Approved, just a few suggestions for the future in comments

PR #1465, opened 2024-10-16
Develop

hma-3:
[DEMO LINK](https://hma-3.github.io/react_todo-app-add-and-delete/)


EustasTheMonk (approved):
Very nice job! I`m impressed.

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,128 @@
...
+import { ErrorMessages, StatusFilter, Todo } from '../types';
+import { addTodo, deleteTodo, getTodos, USER_ID } from '../api/todos';
+import { useTodoForm } from './useTodoForm';
+
+export const useTodo = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [activeTodosAmount, setActiveTodosAmount] = useState(0);
+  const [completedTodoIds, setCompletedTodoIds] = useState<number[]>([]);
```
dinaminahi: I think activeTodosAmount and completedTodoIds are derived from the todos array, then they don't need to be stored in state. Instead, you can calculate them dynamically using useMemo, which will only recompute when todos change. This approach prevents unnecessary state updates and keeps the logic cleaner

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,128 @@
...
+  const {
+    newTodoTitle,
+    setNewTodoTitle,
+    isLoadingSubmit,
+    setIsLoadingSubmit,
+    isFocusedInput,
+    setIsFocusedInput,
+  } = useTodoForm();
+
+  const todosAmount = useMemo(() => todos.length, [todos]);
```
dinaminahi: array length is a simple calculation, useMemo might not always be necessary for such lightweight operations

dinaminahi (approved):
Good job 👍 Pay attention to these comments in the next task

PR #1466, opened 2024-10-16
solution of the second part

PikalovAndrey:
[DEMO LINK](https://PikalovAndrey.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+      userId: USER_ID,
+      completed: false,
+    };
+
+    setTempTodo(temporaryTodo);
+    setIsSubmitting(true);
+    setShouldFocus(false);
+    setLoadingTodosCount(currentCount => [...currentCount, 0]);
+
+    addTodo({ title: inputValue.trim(), userId: USER_ID, completed: false })
```
hma-3: you can use `temporaryTodo` for this

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+  const [shouldFocus, setShouldFocus] = useState(true);
+  const [loadingTodosCount, setLoadingTodosCount] = useState<number[]>([]);
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [todos]);
```
hma-3: you have `shouldFocus`, so why don't you use it here?

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+  };
+
+  const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (inputValue.trim() === '') {
+      setErrorMessage(ErrorMessages.EMPTY_TITLE);
+
+      return;
+    }
```
hma-3: i think it's better to use `length`
```ts
if (!inputValue.trim().length)
```

and you use it not once, so consider creating a variable

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+    addTodo({ title: inputValue.trim(), userId: USER_ID, completed: false })
+      .then((createdTodo: Todo) => {
+        setTodos((currentTodos: Todo[]) => [...currentTodos, createdTodo]);
+        setInputValue('');
+        setErrorMessage(ErrorMessages.NO_ERRORS);
+        setShouldFocus(true);
+
+        if (inputRef.current) {
+          inputRef.current.focus();
+        }
```
hma-3: why?🥹
just add right dependencies and conditions in `useEffect`

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+          inputRef.current.disabled = false;
+          inputRef.current.focus();
+        }
+
+        setIsSubmitting(false);
+        setTempTodo(null);
+        setLoadingTodosCount(currentCount =>
+          currentCount.filter(todoId => todoId !== 0),
+        );
+      });
```
hma-3: move `setShouldFocus(true)` from `then()` and `catch()` here

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+        if (inputRef.current) {
+          inputRef.current.disabled = false;
+          inputRef.current.focus();
+        }
+
+        setIsSubmitting(false);
+        setTempTodo(null);
+        setLoadingTodosCount(currentCount =>
+          currentCount.filter(todoId => todoId !== 0),
+        );
```
hma-3, line 75: i guess it will be more readable if you swap `setLoadingTodosCount` and `setTempTodo`
then you'll be able to filter `ids` depending on `tempTodo.id`, not just zero

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+      });
+  };
+
+  const handleTodoDelete = useCallback(
+    (todoId: number) => {
+      setIsSubmitting(true);
```
hma-3: first of all, why do you need `setIsSubmitting` here?
and you are using the same one for the form, so it's a bit confusing and maybe can lead to errors in some cases

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+          setErrorMessage(ErrorMessages.DELETING_ERROR);
+        })
+        .finally(() => {
+          setIsSubmitting(false);
+          setLoadingTodosCount(current =>
+            current.filter(deletingTodoId => todoId !== deletingTodoId),
+          );
+        });
+    },
+    [todos],
```
hma-3: your function depends on `todoId`, not `todos`

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+  );
+  const handleCompletedTodosDeleted = useCallback(() => {
+    setIsSubmitting(true);
```
hma-3: > first of all, why do you need `setIsSubmitting` here?
and you are using the same one for the form, so it's a bit confusing and maybe can lead to errors in some cases

same

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+          .catch(() => {
+            setErrorMessage(ErrorMessages.DELETING_ERROR);
+          })
+          .finally(() => {
+            setLoadingTodosCount(current =>
+              current.filter(deletingTodoId => deletingTodoId !== id),
+            );
+          }),
+      ),
+    ).finally(() => setIsSubmitting(false));
```
hma-3: you can just use `handleTodoDelete` and `forEach` here

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+  }, [todos]);
+
+  const handleTodosToggle = useCallback(() => {
+    const hasIncompleteTodos = todos.some(todo => !todo.completed);
+
+    setTodos(currentTodos =>
+      currentTodos.map(todo => ({
+        ...todo,
+        completed: hasIncompleteTodos,
```
hma-3, line 134: that's cool🔥

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+                : todo,
+            ),
+          );
+        })
+        .catch(() => {
+          setErrorMessage(ErrorMessages.UPDATING_ERROR);
+        })
+        .finally(() => setIsSubmitting(false));
+    },
+    [todos],
```
hma-3: not `todos`, but `todoId` and `currentCompletedStatus`

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+          setTodos(loadedTodos);
+        })
+        .catch(error => {
+          setErrorMessage(ErrorMessages.LOADING_ERROR);
+          throw new Error(error);
+        });
+    };
+
+    loadTodos();
+  }, []);
```
hma-3: why do you need another function (`loadTodos`) here?

src/api/todos.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const deleteTodos = (todoId: number) => {
+  return client.delete(`/todos/${todoId}`);
+};
+
+export const addTodo = ({
+  title,
+  userId,
+  completed,
+}: {
```
hma-3: `newTodo: Omit<Todo, 'id'>`?

src/components/Footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: filter === FilterOptions.COMPLETED,
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setFilter(FilterOptions.COMPLETED)}
+        >
+          Completed
+        </a>
```
hma-3: would be good if  map through `FilterOptions`

src/components/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+      <form onSubmit={onSubmit}>
+        <input
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          autoFocus={shouldFocus && tempTodo == null}
+          value={value}
+          ref={inputRef}
```
hma-3: why do you use `autoFocus` if you have `ref`?🥹

hma-3 (approved):
Good job!
It turned out that I didn't have enough time to review your implementation, but I definitely recommend you to review your code again. I'm sure you can do even better

src/App.tsx
igorkruz: Remove `todos `from the dependency array in the `useCallback `hooks where it's not directly used.

src/App.tsx
igorkruz: Good point!

src/App.tsx
igorkruz: fix it

src/App.tsx
igorkruz: You can add more dependencies in `useEffect `and avoid rewriting this

src/App.tsx
igorkruz: `isSubmitting` its a loader that is displayed when you are manipulating with todos, maybe it's better to choose a more descriptive name for it

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+              currentTodos.filter(todo => todo.id !== id),
+            );
+          })
+          .catch(() => {
+            setErrorMessage(ErrorMessages.DELETING_ERROR);
+          })
+          .finally(() => {
+            setLoadingTodosCount(current =>
+              current.filter(deletingTodoId => deletingTodoId !== id),
+            );
```
igorkruz, line 122: why are u changing the `loadingTodosCount` in `finally`, it will be counted not depending on the request status, maybe its better to count `loadingTodosCount` if the request ends with success.

src/App.tsx
igorkruz: Remove unnecessary dependency: `todos'

src/App.tsx
```diff
@@ -1,26 +1,254 @@
...
+    const loadTodos = () => {
+      getTodos()
+        .then(loadedTodos => {
+          setTodos(loadedTodos);
+        })
+        .catch(error => {
+          setErrorMessage(ErrorMessages.LOADING_ERROR);
+          throw new Error(error);
+        });
+    };
```
igorkruz: It's better to move the logic inside the useEffect hook to a separate function, as it improves code readability, reusability, and makes testing easier. It also helps keep useEffect focused on side effects.

src/components/Footer.tsx
igorkruz: Good point 👍

src/components/TodoList.tsx
```diff
@@ -0,0 +1,62 @@
...
+    <section className="todoapp__main" data-cy="TodoList">
+      <TransitionGroup>
+        {todosAfterFiltering.map(todo => (
+          <CSSTransition
+            key={todo.id}
+            classNames="item"
+            timeout={300}
+            addEndListener={(node, done) => {
+              node.addEventListener('transitionend', done, false);
+            }}
```
igorkruz: create a handle function for this logic

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

src/App.tsx
PikalovAndrey: Deleting loadingTodosCount in the finally block is done to clear the loading indication for each todo, regardless of whether the delete operation was successful or failed.

src/App.tsx
```diff
@@ -1,26 +1,228 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodosCount, setLoadingTodosCount] = useState<number[]>([]);
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [todos, inputRef]);
```
igorkruz: Move the `useEffect `to the end of code
Placing `useEffect `at the end improves readability by letting developers first grasp the main logic before seeing side effects. This promotes clearer structure and consistency in the codebase.

src/components/Errors.tsx
```diff
@@ -0,0 +1,30 @@
+import classNames from 'classnames';
+import { ErrorMessages } from '../enums/ErrorMessages';
+
+interface ErrorsProps {
+  errorMessage: ErrorMessages;
+  setErrorMessage: React.Dispatch<React.SetStateAction<ErrorMessages>>;
```
igorkruz:
```suggestion
  setErrorMessage: Dispatch<SetStateAction<ErrorMessages>>;
```

src/components/Errors.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export const Errors: React.FC<ErrorsProps> = ({
```
igorkruz:
```suggestion
export const Errors: FC<ErrorsProps> = ({
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,57 @@
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import { FilterOptions } from '../enums/FilterOptions';
+
+interface FooterProps {
+  todos: Todo[];
+  completedTodos: Todo[];
+  filter: FilterOptions;
+  setFilter: React.Dispatch<React.SetStateAction<FilterOptions>>;
```
igorkruz:
```suggestion
  setFilter: Dispatch<SetStateAction<FilterOptions>>;
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  onCompletedTodosDeleted: () => void;
+}
+
+export const Footer: React.FC<FooterProps> = ({
```
igorkruz:
```suggestion
export const Footer: FC<FooterProps> = ({
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,57 @@
...
+                selected: filter === filterOption,
+              })}
+              data-cy={`FilterLink${filterOption.charAt(0).toUpperCase() + filterOption.slice(1)}`}
+              onClick={() => setFilter(filterOption)}
+            >
+              {filterOption.charAt(0).toUpperCase() + filterOption.slice(1)}
+            </a>
+          ))}
+        </nav>
+      </nav>
```
igorkruz: Avoid Nested <nav> Elements
You have a nested <nav> tag, which isn't semantically correct. You can remove one of them to improve semantics.

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

igorkruz (approved):
Well done 🔥

PR #1470, opened 2024-10-16
Solution

v1RnT:
[DEMO LINK](https://v1rnt.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,5 @@
 import React, { useEffect, useMemo, useState } from 'react';
-import { getTodos, USER_ID } from './api/todos';
-import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
 
 import { Todo } from './types/Todo';
 import { Filter } from './types/Filter';
@@ -43,15 +42,13 @@ export const App: React.FC = () => {
         setErrorMessage,
       );
     }
+  }, [deletionIds]);
 
+  useEffect(() => {
     getTodos()
       .then(setTodosFromServer)
       .catch(() => handleError(setErrorMessage, Errors.LoadingTodos));
-  }, [deletionIds]);
-
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  }, []);
 
   return (
     <div className="todoapp">
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,21 +22,6 @@ export const App: React.FC = () => {
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [deletionIds, setDeletionIds] = useState<number[]>([]);
 
-  if (deletionIds.length !== 0) {
-    handleDeletion(
-      deletionIds,
-      setTodosFromServer,
-      setDeletionIds,
-      setErrorMessage,
-    );
-  }
-
-  useEffect(() => {
-    getTodos()
-      .then(setTodosFromServer)
-      .catch(() => handleError(setErrorMessage, Errors.LoadingTodos));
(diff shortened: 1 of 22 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+      })
+      .catch(() => {
+        setTempTodo(null);
+        handleError(setError, Errors.AddTodo);
+
+        if (inputRef.current?.value) {
+          inputRef.current.value = inputValue;
+        }
+      });
+  };
```
Hradivnyk: It would be better to create some function for it. For example, addNewTodo

Hradivnyk (approved):

src/App.tsx
```diff
@@ -1,26 +1,95 @@
...
+      setDeletionIds,
+      setErrorMessage,
+    );
+  }
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodosFromServer)
+      .catch(() => handleError(setErrorMessage, Errors.LoadingTodos));
+  }, []);
```
igorkruz: Move the `useEffect `to the end of code
Placing `useEffect `at the end improves readability by letting developers first grasp the main logic before seeing side effects. This promotes clearer structure and consistency in the codebase.

igorkruz: It's better to move the logic inside the `useEffect `hook to a separate function, as it improves code readability, reusability, and makes testing easier. It also helps keep `useEffect `focused on side effects.

src/App.tsx
```diff
@@ -1,26 +1,95 @@
...
+  const [deletionIds, setDeletionIds] = useState<number[]>([]);
+
+  if (deletionIds.length !== 0) {
+    handleDeletion(
+      deletionIds,
+      setTodosFromServer,
+      setDeletionIds,
+      setErrorMessage,
+    );
+  }
```
igorkruz: The `deletionIds `check occurs at the top level of the component, which may lead to potential issues. It would be more appropriate to place it inside a `useEffect `hook to ensure it runs when `deletionIds `change, rather than on every render.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+  tempTodo,
+  todos,
+  setError,
+  setTodos,
+}) => {
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, [todos, tempTodo]);
```
igorkruz: Move the `useEffect `to the end of code
Placing `useEffect `at the end improves readability by letting developers first grasp the main logic before seeing side effects. This promotes clearer structure and consistency in the codebase.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    setTempTodo(temporaryTodo);
+
+    postTodo(temporaryTodo)
+      .then(res => {
+        setTodos(current => [...current, res]);
+
+        if (inputRef.current?.value) {
+          inputRef.current.value = '';
+        }
```
igorkruz: After posting the `todo`, you clear the input by setting` inputRef.current.value` to an empty string. This is effective but consider using `setState `for controlled components if you're maintaining form state with React. 
It provides better control and may prevent issues when the component re-renders.

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

KosolapovAnatolii (approved):
Good job! 🔥 🔥 🚒

PR #1474, opened 2024-10-17
Add task solution

WatarJoy:
https://WatarJoy.github.io/react_todo-app-add-and-delete/


src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
+  const [title, setTitle] = useState<string>('');
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      try {
+        const data = await getTodos();
+
+        setTodos(data);
+      } catch (err) {
+        setError('Unable to load todos');
```
dokrod: You can separate all types of errors into a separate type

src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
+    } finally {
+      setIsLoading(false);
+      setTimeout(() => {
+        const input =
+          document.querySelector<HTMLInputElement>('.todoapp__new-todo');
+
+        input?.focus();
+      }, 100);
+    }
+  };
```
dokrod: I think it is better to put such large functions in a separate file

src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
+        setError(null);
+      }, 3000);
+    } finally {
+      setIsLoading(false);
+      setTimeout(() => {
+        const input =
+          document.querySelector<HTMLInputElement>('.todoapp__new-todo');
+
+        input?.focus();
+      }, 100);
```
dokrod: You can return focus with `useRef` and .`current?.focus()`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const [isDeletingItem, setIsDeletingItem] = useState<boolean>(false);
+
+  const handleDelete = async () => {
+    setIsDeletingItem(true);
+    await onDelete(todo.id);
+    setIsDeletingItem(false);
+  };
+
+  return (
+    <div data-cy="Todo" className={`todo ${todo.completed ? 'completed' : ''}`}>
```
dokrod: Use `className` correctly `<div data-cy="Todo" className={classNames('todo', { completed: todo.completed })}>`

src/types/Props.ts
```diff
@@ -0,0 +1,38 @@
...
+  isLoading: boolean;
+}
+
+export interface TodoListProps {
+  todos: Todo[];
+  onDelete: (todoId: number) => Promise<void>;
+  isLoading: boolean;
+  setIsLoading: React.Dispatch<React.SetStateAction<boolean>>;
+  isDeleting: boolean;
+}
```
dokrod: I don't think it is necessary to put the `Props` in a separate file

src/types/enums.ts
```diff
@@ -0,0 +1,9 @@
+export enum ErrorMessages {
+  LOAD_TODOS = 'Unable to load todos',
+}
```
dokrod: you can write all types of Error here and use it like this `ErrorMessages.LOAD_TODOS`

dokrod (approved):
Nice work, but you can optimize your code and make it more readable

src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filter, setFilter] = useState<FilterStates>(FilterStates.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [title, setTitle] = useState<string>('');
```
VolodymyrKirichenko: If u specify initialValue as a simple type (boolean, number, string, etc.) - TypeScript itself understands what type should be in the state
```suggestion
  const [isLoading, setIsLoading] = useState(false);
  const [title, setTitle] = useState('');
```

src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
+import { ErrorNotification } from './components/ErrorNotification';
+import { FilterStates } from './types/enums';
+import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
```
VolodymyrKirichenko: u can create `enum` for all ur error. Do it for all error
```suggestion
  const [error, setError] = useState<ErrorType>(ErrorType.DEFAULT);
```
```
export enum ErrorType = {
  DEFAULT: '',
  ERROR_TITLE: 'Error text........'
  ............
}
```

src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
+  const [filter, setFilter] = useState<FilterStates>(FilterStates.ALL);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [title, setTitle] = useState<string>('');
+
+  useEffect(() => {
```
VolodymyrKirichenko: ur `useEffect` hooks should always be at the bottom, before rendering. This is done for readability and convenience. Move it

src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
+        }, 3000);
+
+        return () => clearTimeout(timer);
+      }
+    };
+
+    fetchTodos();
+  }, []);
+
+  const handleAddTodo = async (): Promise<void> => {
```
VolodymyrKirichenko: useCallback? For all such cases

src/App.tsx
```diff
@@ -1,26 +1,185 @@
...
+      setTimeout(() => {
+        const input =
+          document.querySelector<HTMLInputElement>('.todoapp__new-todo');
+
+        input?.focus();
+      }, 100);
+    }
+  };
+
+  const filteredTodos = todos.filter(todo => {
```
VolodymyrKirichenko: use `useMemo` or move this function to separate component

src/types/Props.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export interface ErrorNotificationProps {
+  error: string | null;
+  onHideError: () => void;
+}
+
+export interface FooterProps {
+  todos: Todo[];
+  filter: FilterStates;
+  setFilter: React.Dispatch<React.SetStateAction<FilterStates>>;
```
VolodymyrKirichenko: use this syntax for all such cases. More readable
```suggestion
  setFilter: (filterStates: FilterStates) => void;
```

src/types/Props.ts
VolodymyrKirichenko: yep, u don't need it. 
it is better to keep props in the component you need, it is more convenient

src/components/Footer.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodos} items left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        {Object.values(FilterStates).map(state => (
+          <a
+            key={state}
+            href={`#/${state}`}
+            className={`filter__link ${filter === state ? 'selected' : ''}`}
```
VolodymyrKirichenko: use [classnames](https://www.npmjs.com/package/classnames) library for it. In all such cases

VolodymyrKirichenko (changes requested):
Overall - good job, but let's make some changes 🥹

src/App.tsx
igorkruz: Not fixed!

src/App.tsx
```diff
@@ -1,26 +1,178 @@
...
+  }, [todos, filter]);
+
+  const hasTodos = todos.length > 0;
+  const allCompleted = todos.length > 0 && todos.every(todo => todo.completed);
+
+  useEffect(() => {
+    if (error) {
+      setTimeout(() => {
+        setError(null);
+      }, 3000);
```
igorkruz: It's a good practice to clear timeouts in the cleanup function of `useEffect `to prevent memory leaks. This ensures that any pending timeouts are cleared if the component unmounts or if the dependencies change.

src/components/Footer.tsx
```diff
@@ -0,0 +1,54 @@
...
+  filter,
+  setFilter,
+  onClearCompleted,
+}) => {
+  const activeTodos = todos.filter(todo => !todo.completed).length;
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodos} items left`}
```
igorkruz:
```suggestion
        {activeTodos} items left
```

src/App.tsx
```diff
@@ -1,26 +1,178 @@
...
+import { ErrorNotification } from './components/ErrorNotification';
+import { ErrorMessages, FilterStates } from './types/enums';
+import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
```
igorkruz: You have `ErrorMessages enum` why you don't use it?

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

igorkruz (approved):
Good Job 👍

PR #1479, opened 2024-10-18
Develop

coldmelnyk:
[DEMO LINK](https://coldmelnyk.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,11 +25,42 @@ export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [newTodoTitle, setNewTodoTitle] = useState<string>('');
   const [filterType, setFilterType] = useState<FilterTypes>(FilterTypes.All);
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [isLoading, setIsLoading] = useState(false);
   const [errorMessage, setErrorMessage] = useState<ErrorMessage>(
     ErrorMessage.none,
   );
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [arrayOfTodoId, setArrayOfTodoId] = useState<number[]>([]);
+
+  const clearAllCompletedTodos = () => {
+    setIsLoading(true);
+    const ids: number[] = [];
+
+    for (const todo of todos) {
+      if (todo.completed) {
+        ids.push(todo.id);
+      }
+    }
+
+    if (ids.length > 0) {
+      setArrayOfTodoId(ids);
+
+      for (const id of ids) {
+        deleteTodo(id)
+          .then(() =>
+            setTodos(currentTodos =>
+              currentTodos.filter(todo => todo.id !== id),
+            ),
+          )
+          .catch(() => {
+            setErrorMessage(ErrorMessage.delete);
+          });
+      }
+
+      setArrayOfTodoId([]);
+      setIsLoading(false);
+    }
+  };
 
   const uploadingTodos = useMemo(() => {
     setErrorMessage(ErrorMessage.none);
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+
+      for (const id of ids) {
+        deleteTodo(id)
+          .then(() => {
+            successIds.push(id);
+          })
+          .catch(() => {
+            setErrorMessage(ErrorMessage.delete);
+          });
+      }
```
igorkruz: Instead of handling deletions one by one, you can use `Promise.all` to wait for all delete operations to complete. This can improve performance and simplify error handling.

src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+  const clearAllCompletedTodos = () => {
+    setIsLoading(true);
+    const ids: number[] = [];
+    const successIds: number[] = [];
+
+    for (const todo of todos) {
+      if (todo.completed) {
+        ids.push(todo.id);
+      }
+    }
```
igorkruz:
```suggestion
    const ids = todos.filter(todo => todo.completed).map(todo => todo.id);
```

src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+          });
+      }
+
+      setTimeout(() => {
+        setTodos(currentTodos =>
+          currentTodos.filter(todo => !successIds.includes(todo.id)),
+        );
+        setArrayOfTodoId([]);
+        setIsLoading(false);
+      }, 300);
```
igorkruz: Instead of using `setTimeout `to update the state after a delay, wait for the delete operations to finish before updating the state.

src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+        .finally(() => {
+          setTempTodo(null);
+          setIsLoading(false);
+        });
+    }
+
+    return setErrorMessage(ErrorMessage.emptyTitle);
+  };
+
+  const isTodosEmpty = todos.length === 0;
```
igorkruz, line 126:
```suggestion
  const isTodosEmpty = !todos.length;
```

src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+        setTodos(currentTodos => {
+          const updatingTodos = [...currentTodos];
+          const indexOfTodo = todos.findIndex(
+            todoFromArray => todoFromArray.id === patchedTodo.id,
+          );
+
+          updatingTodos.splice(indexOfTodo, 1, todoWithNewStatus);
+
+          return updatingTodos;
+        }),
```
igorkruz, line 157:
```suggestion
        setTodos(currentTodos =>
          currentTodos.map(todo =>
             todo.id === patchedTodo.id ? todoWithNewStatus : todo
          )
        );
```

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const ErrorNotification: React.FC<Props> = ({
+  errorMessage,
+  handleError,
+}) => {
+  useEffect(() => {
+    if (errorMessage !== ErrorMessage.none) {
+      setTimeout(() => {
+        handleError(ErrorMessage.none);
+      }, 3000);
```
igorkruz: It's a good practice to clear timeouts in the cleanup function of useEffect to prevent memory leaks. This ensures that any pending timeouts are cleared if the component unmounts or if the dependencies change.

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

src/App.tsx
KosolapovAnatolii: Don't forget fix it

src/App.tsx
KosolapovAnatolii: Same here

KosolapovAnatolii (approved):
Good solution!
But it looks like you skiped few recomendations from previous review. Don't forget fit it and go forward to the next step 🚀

coldmelnyk:
oops, strange because I've changed it

PR #1491, opened 2024-10-24
task solution

olshum8:
[DEMO LINK](https://olshum8.github.io/react_todo-app-add-and-delete/)


etojeDenys (changes requested):
1. when user tries to add the todo you need to show the temp todo and display loader only on that temp todo
2. and when the user tries to delete the todo you need to show the loader on that todo

feel free to ask for some help in the chat how to implement it

anastasiiavorobiova (approved):

PR #1508, opened 2024-11-14
solution

svitlanak31:
- [DEMO LINK](https://svitlanak31.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+        setError(null);
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+
+    return undefined;
+  }, [error]);
+
+  function deleteTodo(todoId: number) {
```
volodymyr-soltys97: It is important to follow the same way of creating functions everywhere, fix it everywhere
```suggestion
  const deleteTodo = (todoId: number) => {
```

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+      .catch(() => {
+        setError('Unable to delete a todo');
+      });
+  }
+
+  const filteredTodos = (() => {
+    switch (statusFilter) {
+      case 'Active':
+        return todos.filter(todo => !todo.completed);
+      case 'Completed':
```
volodymyr-soltys97: Create a enum for 'All', 'Active', 'Completed' and use it everywhere

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+              disabled={isInputDisabled}
+            />
+          </form>
+        </header>
+
+        <section className="todoapp__main" data-cy="TodoList">
+          {filteredTodos.map(todo => (
+            <div
+              data-cy="Todo"
+              className={`todo ${todo.completed ? 'completed' : ''}`}
```
volodymyr-soltys97: Use the classnames library for add classes with condition, fix it everywhere

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+                  id="title"
+                  data-cy="TodoStatus"
+                  type="checkbox"
+                  className="todo__status"
+                  checked={todo.completed}
+                />
+              </label>
+
+              <span data-cy="TodoTitle" className="todo__title">
+                {todo.title}
```
volodymyr-soltys97: Use destructuring for `todo.id`

src/App.tsx
```diff
@@ -1,26 +1,296 @@
...
+              </a>
+
+              <a
+                href="#/completed"
+                className={`filter__link ${statusFilter === 'Completed' ? 'selected' : ''}`}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setStatusFilter('Completed')}
+              >
+                Completed
+              </a>
```
volodymyr-soltys97: Use `Object.value(your created enum)` and render these options with `map()` method

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is a bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

anastasiiavorobiova (approved):
Great work!

PR #1523, opened 2024-12-11
add task solution

DimaDamage91:
[DEMO LINK](https://DimaDamage91.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,151 @@
...
+      case 'update':
+        setError('Unable to update a todo');
+        break;
+      default:
+        setError('An unknown error occured');
+    }
+    setIsErrorVisible(true);
+  };
+
+  function deleteTodo(todoId: number) {
```
volodymyr-soltys97: It is important to follow the same way of creating functions everywhere, fix it everywhere
```suggestion
  const deleteTodo = (todoId: number) => {
```

src/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+          Active
+        </a>
+        <a
+          href="#/completed"
+          className={filter === 'completed' ? 'selected' : ''}
+          onClick={() => setFilter('completed')}
+          data-cy="FilterLinkCompleted"
+        >
+          Completed
+        </a>
```
volodymyr-soltys97: Create a enum for 'all', 'active', 'completed' and use it everywhere. Also, use `Object.values(your created enum)` and render these options with `map()` method

src/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+          href="#/"
+          className={filter === 'all' ? 'selected' : ''}
+          onClick={() => setFilter('all')}
+          data-cy="FilterLinkAll"
+        >
+          All
+        </a>
+        <a
+          href="#/active"
+          className={filter === 'active' ? 'selected' : ''}
```
volodymyr-soltys97: Use the classnames library for add classes with condition, fix it everywhere

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code
You need to fix this failed test, if you need some help feel free in the fe_chat 
image

Move these components to the `components` folder 
image

volodymyr-soltys97 (approved):
Great! 🔥

PR #1534, opened 2024-12-17
upd

SeemsGood78:
[DEMO LINK] (https://SeemsGood78.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,8 +38,7 @@ export const App: React.FC = () => {
     }
   }, [isAdding]);
 
-
-  function filterTodos(todos:Todo[]) {
+  function filterTodos() {
     switch (filter) {
       case Filter.Active:
         return todos.filter(todo => !todo.completed);
@@ -129,7 +128,7 @@ export const App: React.FC = () => {
       });
   };
 
-  const filteredTodos = filterTodos(todos);
+  const filteredTodos = filterTodos();
 
   useEffect(() => {
     if (errorMessage) {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,26 +1,299 @@
...
+    if (!isAdding && inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [isAdding]);
+
+  const filterTodos = () => {
+    if (filter === 'active') {
+      return todos.filter(todo => !todo.completed);
+    } else if (filter === 'completed') {
+      return todos.filter(todo => todo.completed);
```
volodymyr-soltys97: Create a enum for 'all', 'active', 'completed' and use it everywhere. Also, will be better if you use switch case here

src/App.tsx
```diff
@@ -1,26 +1,299 @@
...
+              {activeCount.length} items left
+            </span>
+
+            {/* Active link should have the 'selected' class */}
+            <nav className="filter" data-cy="Filter">
+              <a
+                href="#/"
+                className={
+                  filter === 'all' ? 'filter__link selected' : 'filter__link'
+                }
```
volodymyr-soltys97: Use the classnames library for add classes with conditionm fix it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is bad practice to write all the code in one component, you need to separate the logic into different components. For example `Header, Footer, TodoItem, TodoList`

volodymyr-soltys97 (changes requested):
Hi, the comments from the previous review still not fixed, you need to fix all comments, if you need some help feel free in the fe_chat

SeemsGood78:
There is no such requirement in the specification, and I don't see the point in spending a couple of hours splitting the code into separate components if it's not a mandatory part of the task.

vadiimvooo (approved):
Good job, but please, in final part separate your app on smaller components as it was mentioned above. This project can be used in your portfolio and it should look as good as it possible.

PR #1539, opened 2024-12-17
Task solution

AnderZerfall:
[DEMO LINK](https://anderzerfall.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,131 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
+// General libs
 import React from 'react';
+import { useEffect, useState } from 'react';
+import classNames from 'classnames';
+// components
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoCreate } from './components/TodoCreate/TodoCreate';
+import { TodoFooter } from './components/TodoFooter/TodoFooter';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+// api
+import * as todoServices from './api/todos';
+// Types
+import { Todo } from './types/Todo';
+import { StatusFilter } from './types/StatusFilter';
+import { filterTodos } from './utils/filterTodos';
+import { ErrorType } from './types/ErrorType';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todoList, setTodoList] = useState<Todo[]>([]);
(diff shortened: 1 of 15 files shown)


AlyonaV22 (approved):
Good Job!

src/App.tsx
```diff
@@ -1,26 +1,104 @@
...
+        setTodoList(todos);
+      } catch {
+        sendError(ErrorType.LOAD_TODOS);
+      }
+    })();
+  }, []);
+
+  const filteredTodos = filterTodos(statusFilter, todoList);
+  const activeTodosCount = todoList.filter(todo => !todo.completed).length;
+  const hasCompletedTodos = todoList.length > activeTodosCount;
```
VolodymyrKirichenko, line 51: useMemo?

src/App.tsx
```diff
@@ -1,26 +1,104 @@
...
+
+  const clearCompleted = async () => {
```
VolodymyrKirichenko: useCallback?

```suggestion
  const handleClearCompleted = async () => {
```

src/App.tsx
```diff
@@ -1,26 +1,104 @@
...
+      <div className="todoapp__content">
+        <TodoCreate
+          setTodoList={setTodoList}
+          sendError={sendError}
+          userId={todoServices.USER_ID}
+          setTempTodo={setTempTodo}
+          todoList={todoList}
+        />
+
+        {todoList.length > 0 && (
```
VolodymyrKirichenko: will be more readable i think

```suggestion
        {!!todoList.length && (
```

src/App.tsx
```diff
@@ -1,26 +1,104 @@
...
+    (async () => {
+      try {
+        const todos = await todoServices.getTodos();
+
+        setTodoList(todos);
+      } catch {
+        sendError(ErrorType.LOAD_TODOS);
+      }
+    })();
+  }, []);
```
VolodymyrKirichenko: better or no?what do u think? 🤔 

```suggestion
  useEffect(() => {
    const getTodos = async () => {
      try {
        const todos = await todoServices.getTodos();

        setTodoList(todos);
      } catch {
        sendError(ErrorType.LOAD_TODOS);
      }
    }
    
    getTodos();
  }, []);
```

VolodymyrKirichenko (approved):
Well done 🔥

src/App.tsx
AnderZerfall: The funny thing is that I used to have this variant in a previous task, but I was told to use IIFE instead. I prefer your suggestion because of its clarity, so I'll go with it in the next task.

PR #1546, opened 2024-12-18
testtest

kozlovoleksii:
https://kozlovoleksii.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,17 +4,23 @@ import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { deleteTodo, getTodos, postTodo, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
-import classNames from 'classnames';
 import { sendErrorMessage } from './utils/sendError';
+import { useRef } from 'react';
+import { Footer } from './components/footer';
+import { ErrorMessage } from './components/errorsUnderFooter';
+import { MainSection } from './components/section';
+import { Header } from './components/header';
 
 export const App: React.FC = () => {
   const [todoList, setTodoList] = useState<Todo[]>([]);
   const [filteredTodoList, setFilteredTodoList] = useState<Todo[]>([]);
-  const [errorMessage, setErrorMessage] = useState('');
   const [selectedFilter, setSelectedFilter] = useState('all');
   const [title, setTitle] = useState('');
   const [loader, setLoader] = useState(false);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const [deletingTodos, setDeletingTodos] = useState<number[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
 
   useEffect(() => {
     getTodos(USER_ID)
@@ -26,6 +32,12 @@ export const App: React.FC = () => {
         sendErrorMessage('Unable to load todos', setErrorMessage);
       });
   }, []);
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [tempTodo, deletingTodos]);
 
   const filterTodos = useCallback(
     (status: string): void => {
@@ -77,7 +89,7 @@ export const App: React.FC = () => {
     }
 
     const newTempTodo: Todo = {
-      id: 0, // Тимчасовий id
+      id: 0,
       userId: USER_ID,
       title: title.trim(),
       completed: false,
@@ -88,7 +100,7 @@ export const App: React.FC = () => {
 
     postTodo({
       userId: USER_ID,
-      title: title,
+      title: title.trim(),
       completed: false,
     })
       .then(data => {
@@ -96,10 +108,9 @@ export const App: React.FC = () => {
         setTempTodo(null);
         setTitle('');
       })
-      .catch(error => {
+      .catch(() => {
         setTempTodo(null);
         sendErrorMessage('Unable to add a todo', setErrorMessage);
-        throw new Error(error);
       })
       .finally(() => {
         setLoader(false);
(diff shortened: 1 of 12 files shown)


volodymyr-soltys97 (changes requested):
Hi, there is no demo link, also, you need to add demo to the PR description 
image

kozlovoleksii:
Володимир...допоможіть розібратись, чого не проходить тест:
Це блок тестів: Page with mixed todos
```
1) should not have loaders
  12 passing (18s)
  1 failing

  1)
       Page with mixed todos
         should not have loaders:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-cy="TodoLoader"]`, but never found it. Queried from:

              > <div.todo.completed>
      at Object.assertNotLoading (webpack://react_todo-app-add-and-delete/./cypress/integration/page.spec.js:80:68)
      at Context.eval (webpack://react_todo-app-add-and-delete/./cypress/integration/page.spec.js:219:12)
```

volodymyr-soltys97 (changes requested):
Good job! 
Let's improve your code
It is bad practice to write all the code in one component, you need to separate the logic into different components. Also, if you need help with tests pls feel free in the fe_chat

kozlovoleksii:
> хороша робота! Давайте покращимо ваш код. Погана практика писати весь код в одному компоненті, вам потрібно розділити логіку на різні компоненти. Крім того, якщо вам потрібна допомога з тестами, будь ласка, не соромтеся завітати до fe_chat

thanks for the advice :)
I applied... I've been waiting for an answer for a day now... well, I'll wait some more))

src/components/section.tsx
```diff
@@ -0,0 +1,100 @@
...
+          >
+            ×
+          </button>
+
+          {deletingTodos.includes(todo.id) && (
+            <div data-cy="TodoLoader" className="modal overlay is-active">
+              <div className="modal-background has-background-white-ter" />
+              <div className="loader" />
+            </div>
+          )}
```
vadiimvooo: This Loader component should be always displayed. Just use class is-active to hide or show when it's needed

src/components/section.tsx
vadiimvooo, line 1: You should work a bit more on separation of your App to smaller components. I would recommend you to check class names of elements and create component per each logical block. Example: TodoItem, TodoList etc.

src/utils/sendError.ts
```diff
@@ -0,0 +1,9 @@
+export function sendErrorMessage(
+  errorText: string,
+  setErrorMessage: React.Dispatch<React.SetStateAction<string>>,
+) {
+  setErrorMessage(errorText);
+  setTimeout(() => {
+    setErrorMessage('');
+  }, 3000);
+}
```
vadiimvooo: This logic can be moved to Error component as logic in useEffect hook. You don't need to move it in utils

src/App.tsx
```diff
@@ -1,26 +1,198 @@
...
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(data => {
+        setTodoList(data);
+        setFilteredTodoList(data);
+      })
+      .catch(() => {
+        sendErrorMessage('Unable to load todos', setErrorMessage);
```
vadiimvooo: It's enouth here to use errorMessage state as you can pass it to error component as well

vadiimvooo (changes requested):
Good job, but you have to pass all tests. Please feel free to ask for help in fe_chat during refactor.

src/components/section.tsx
kozlovoleksii: thanks for the advice) I wrote twice already...but no one has answered me since friday xD

src/App.tsx
```diff
@@ -1,26 +1,199 @@
...
+import { sendErrorMessage } from './components/errorsUnderFooter';
+import { useRef } from 'react';
+import { Footer } from './components/footer';
+import { ErrorMessage } from './components/errorsUnderFooter';
+import { MainSection } from './components/section';
+import { Header } from './components/header';
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [filteredTodoList, setFilteredTodoList] = useState<Todo[]>([]);
```
volodymyr-soltys97: Don't need to create additional state for filtered todos, use only one state
```suggestion
  const [todoList, setTodoList] = useState<Todo[]>([]);
```

src/App.tsx
```diff
@@ -1,26 +1,199 @@
...
+      setSelectedFilter(status);
+
+      switch (status) {
+        case 'active':
+          setFilteredTodoList(todoList.filter(todo => !todo.completed));
+          break;
+        case 'completed':
+          setFilteredTodoList(todoList.filter(todo => todo.completed));
+          break;
+        case 'all':
```
volodymyr-soltys97: Create a enum for 'all', 'active', 'completed' and use it everywhere

src/App.tsx
```diff
@@ -1,26 +1,199 @@
...
+        prevTodoList.filter(todo => todo.id !== todoId),
+      );
+    } catch {
+      sendErrorMessage('Unable to delete a todo', setErrorMessage);
+    } finally {
+      setDeletingTodos(prev => prev.filter(id => id !== todoId));
+    }
+  }
+
+  async function handleDeleteCompletedTodo() {
```
volodymyr-soltys97: It is important to follow the same way of creating functions everywhere
```suggestion
  const handleDeleteCompletedTodo = async () => {
```

src/App.tsx
```diff
@@ -1,26 +1,199 @@
...
+    const completedTodoIds = todoList
+      .filter(todo => todo.completed)
+      .map(todo => todo.id);
+
+    setDeletingTodos(prev => [...prev, ...completedTodoIds]);
+
+    let hasError = false;
+
+    for (const id of completedTodoIds) {
```
volodymyr-soltys97, line 132: Use `forEach` method instead of for loop here

src/components/footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+export const Footer: React.FC<Props> = ({
+  todoList,
+  selectedFilter,
+  filterTodos,
+  handleDeleteCompletedTodo,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {todoList.filter(todo => !todo.completed).length} items left
```
volodymyr-soltys97: Move this logic from jsx to the helper variable and use it here

src/components/footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: selectedFilter === 'completed',
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => filterTodos('completed')}
+        >
+          Completed
+        </a>
```
volodymyr-soltys97: Use `Object.values(your created enum)` and render these options with `map()` method

src/components/todoList.tsx
```diff
@@ -0,0 +1,63 @@
...
+              className="todo__status"
+              checked={todo.completed}
+              id={`todo-status-${todo.id}`}
+              onChange={() => handleToggleCompletion(todo.id)}
+              disabled={deletingTodos.includes(todo.id)}
+            />
+          </label>
+
+          <span data-cy="TodoTitle" className="todo__title">
+            {todo.title}
```
volodymyr-soltys97: Use destructuring for `todo`

volodymyr-soltys97 (changes requested):
Good job! 
Let's improve your code

anastasiiavorobiova (approved):

PR #1559, opened 2024-12-20
solution

yuliiazubenko:
[DEMO LINK](https://yuliiazubenko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,134 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useEffect, useMemo, useState, useRef } from 'react';
+import { addTodo, deleteTodo, getTodos, USER_ID } from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
+import { TodoFooter } from './components/TodoFooter/TodoFooter';
+import { ErrorType } from './types/ErrorType';
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
+import { TodoHeader } from './components/TodoHeader/TodoHeader';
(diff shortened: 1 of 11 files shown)


maxsabo (approved):
good job!

src/App.tsx
```diff
@@ -1,26 +1,132 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loadingTodoIds, setLoadingTodoIds] = useState<number[]>([]);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const filteredTodos = useMemo(
+    () =>
+      todos.filter(todo => {
+        if (filterStatus === FilterStatus.All) {
+          return true;
+        }
```
BudnikOleksii: You don't need to filter at all when `filterStatus === FilterStatus.All`

src/App.tsx
```diff
@@ -1,26 +1,132 @@
...
+        title: todoTitle,
+        completed: false,
+        userId: USER_ID,
+      });
+      const newTodo = await addTodo({ title: todoTitle, completed: false });
+
+      setTodos(prev => [...prev, newTodo]);
+    } catch (error) {
+      setErrorMessage(ErrorType.UnableToAdd);
+      throw error;
```
BudnikOleksii, line 52: You don't need to throw an error if you don't plan to handle it in the other block. Fix in all places

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+type Props = {
+  error: ErrorType;
+  setError: Dispatch<SetStateAction<ErrorType>>;
+};
+
+export const ErrorNotification: React.FC<Props> = props => {
+  const { error, setError } = props;
+
+  const [isHidden, setIsHidden] = useState(true);
```
BudnikOleksii: You don't need this state. Just show or hide depending on the error

BudnikOleksii (approved):
Good work 👍
Approved, but left a few suggestions for the next task

src/App.tsx
yuliiazubenko: I don't get it :(
What should I do here? My solutions are not working

PR #1561, opened 2024-12-22
Develop

maxmodrr:
[DEMO LINK](https://maxmodrr.github.io/react_todo-app-add-and-delete/)

I failed 2 tests but I really don't know what's wrong. I asked for help from one of them but didn't get a response that would help. But my app works as it should
(at least seems like) 

Tests I failed:
should keep an error message for 3s after the last fail
should send a create request


etojeDenys (approved):
👍

PR #1570, opened 2024-12-24
Solution added

CamilleHaus:
[DEMO LINK](https://CamilleHaus.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/cypress/integration/page.spec.js
+++ b/cypress/integration/page.spec.js
@@ -979,7 +979,7 @@ describe('', () => {
     });
   });
 
-  describe.skip('Todo Toggling', () => {
+  describe('Todo Toggling', () => {
     beforeEach(() => {
       page.mockLoad().as('loadRequest');
       page.visit();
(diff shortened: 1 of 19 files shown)


pedro-ruas (changes requested):
Camille, there are some minor missing dependencies in the useEffect hooks, but I have pointed it out in the last task, so fixing only once should be ok.
Besides that, your code is failing the tests, take a look on that, seems that you missed a `loading` state for a todo when toggling it between completed/not-completed.
The toggle should be sent to the API and only after the API returns an OK it should be changed. In the meantime it should only be loading

CamilleHaus:
@pedro-ruas Entendi, dei uma olhada aqui e ainda estou um pouco confusa. Será que seria possível você me apontar um pouco melhor onde eu deveria mexer?

pedro-ruas (approved):
Now it looks perfect!

PR #1574, opened 2024-12-30
solution

RUBIC0N5545:
https://RUBIC0N5545.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -40,8 +40,6 @@ export const App: React.FC = () => {
     setTodosToShow(
       todosDb.filter(todo => {
         switch (selectedFilter) {
-          case FilterTypes.ALL:
-            return true;
           case FilterTypes.ACTIVE:
             return !todo.completed;
           case FilterTypes.COMPLETED:
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,7 +4,7 @@ import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { Todo } from './types/Todo';
 import { client } from './utils/fetchClient';
-import { FILTER_TYPES, FilterType } from './types/FilterType';
+import { FilterTypes } from './types/FilterType';
 import {
   createTodo,
   deleteTodo,
@@ -15,6 +15,7 @@ import {
 import Header from './Components/Header';
 import TodoList from './Components/TodoList';
 import Footer from './Components/Footer';
+import cn from 'classnames';
 
 interface OnDblClickParams {
   index: number;
@@ -30,8 +31,8 @@ export const App: React.FC = () => {
   const [inputLoading, setInputLoading] = useState<boolean>(false);
   const [editIndex, setEditIndex] = useState<number | null>(null);
   const [editValue, setEditValue] = useState<string>('');
-  const [selectedFilter, setSelectedFilter] = useState<FilterType>(
-    FILTER_TYPES.ALL as FilterType,
+  const [selectedFilter, setSelectedFilter] = useState<FilterTypes>(
+    FilterTypes.ALL,
   );
   const [newTodoTitle, setNewTodoTitle] = useState<string>('');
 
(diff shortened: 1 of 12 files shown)


RUBIC0N5545:
> https://RUBIC0N5545.github.io/react_todo-app-add-and-delete/

A2B63E27-A2F1-40D3-A4F9-25B491F1C0B1

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is bad practice to write all the code in one component, it is hard to read now. You need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

src/types/FilterType.ts
```diff
@@ -0,0 +1,6 @@
+export type FilterType = 'all' | 'active' | 'completed';
+export const FILTER_TYPES = {
+  ALL: 'all',
+  ACTIVE: 'active',
+  COMPLETED: 'completed',
+};
```
volodymyr-soltys97: Use enum here instead of object 
image

src/Components/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+  selectedFilter,
+  setSelectedFilter,
+  clearCompleted,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {todosDb.reduce((accumulator, value) => {
+          return accumulator + (value.completed ? 0 : 1);
+        }, 0)}{' '}
```
volodymyr-soltys97: Move this logic from jsx to the helper variable and use it here

src/Components/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </a>
+
+        <a
+          href="#/completed"
+          className={`filter__link ${selectedFilter === FILTER_TYPES.COMPLETED ? 'selected' : ''}`}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setSelectedFilter('completed')}
+        >
+          Completed
+        </a>
```
volodymyr-soltys97: Use `Object.values(your created enum)` and render these options with `map()` method

src/Components/Header.tsx
```diff
@@ -0,0 +1,43 @@
...
+  newTodoTitle,
+  setNewTodoTitle,
+  inputLoading,
+  addTodo,
+}) => {
+  return (
+    <header className="todoapp__header">
+      <button
+        type="button"
+        className={`todoapp__toggle-all ${todosDb.every(todo => todo.completed) ? 'active' : ''}`}
```
volodymyr-soltys97: Use classnames library for add classes with condition, fix it everywhere

src/Components/TodoItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+          />
+        </form>
+      ) : (
+        <>
+          <span
+            data-cy="TodoTitle"
+            className="todo__title"
+            onDoubleClick={() => onDblClick({ index, todo })}
+          >
+            {todo.title}
```
volodymyr-soltys97: Use destructuring for `todo`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/App.tsx
```diff
@@ -1,26 +1,289 @@
...
+    FilterTypes.ALL,
+  );
+  const [newTodoTitle, setNewTodoTitle] = useState<string>('');
+
+  useEffect(() => {
+    setTodosToShow(
+      todosDb.filter(todo => {
+        switch (selectedFilter) {
+          case FilterTypes.ALL:
+            return true;
```
volodymyr-soltys97: Redundant case, because you return it in default case
```suggestion
```

src/Components/Header.tsx
volodymyr-soltys97: Not fixed 
image

src/Components/Footer.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {countItemsLeft} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        {Object.values(FilterTypes).map(filter => (
+          <a
+            key={filter}
+            href={`#/${filter}`}
+            className={`filter__link ${selectedFilter === filter ? 'selected' : ''}`}
```
volodymyr-soltys97: Not fixed from the previous review 
image

volodymyr-soltys97 (changes requested):
Almost done!

volodymyr-soltys97 (approved):
Good job! 🔥

PR #1578, opened 2025-01-03
Develop

bhoriunov:
[DEMO LINK](https://bhoriunov.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -83,17 +83,19 @@ export const App: React.FC = () => {
 
     try {
       const newTodoData = await client.post<Todo>('/todos', {
-        title: newTodo,
-        userId: USER_ID,
+        title: tempTodoItem.title,
+        userId: tempTodoItem.userId,
         completed: tempTodoItem.completed,
       });
 
+      setTempTodo(null);
       setTodos(prev => [...prev, newTodoData]);
+      setNewTodo('');
     } catch {
       setError('Unable to add todo');
-      setTempTodo(null);
       inputRef.current?.focus();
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,26 +1,312 @@
...
+
+  const filteredTodos = todos.filter(todo => {
+    if (filter === 'active') {
+      return !todo.completed;
+    }
+
+    if (filter === 'completed') {
+      return todo.completed;
+    }
+
```
volodymyr-soltys97: Consider will be better if you use switch case here. Also, create a enum for 'all', 'active', 'completed' and use it everywhere

src/App.tsx
```diff
@@ -1,26 +1,312 @@
...
+                data-cy="FilterLinkAll"
+                onClick={() => setFilter('all')}
+              >
+                All
+              </a>
+              <a
+                href="#/active"
+                className={`filter__link ${
+                  filter === 'active' ? 'selected' : ''
+                }`}
```
volodymyr-soltys97: Use the classnames library for add classes with condition, fix it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. You need to focus input after delete todo for fix this failed test 
image
image

2. It is bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

VitaliyBondarenko1982 (approved):
Looks good

PR #1590, opened 2025-01-12
add task solution part2 - Add and Delete

polinabs:
[DEMO LINK](https://polinabs.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,157 @@
...
+  const filteredTodos = useMemo((): Todo[] => {
+    if (filterStatus === 'all') {
+      return todos;
+    }
+
+    if (filterStatus === 'active') {
+      return todos.filter(todo => !todo.completed);
+    }
+
+    return todos.filter(todo => todo.completed);
```
volodymyr-soltys97: Consider will be better if you use switch case here, also, create a enum for 'all', 'active', 'completed' and use it everywhere

src/api/todos.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+// Add more methods here
+/*
+  Your userId is
+  2214
+
+  Please use it for all your requests to the Students API. For example:
+  https://mate.academy/students-api/todos?userId=2214 
+*/
```
volodymyr-soltys97: Remove comments

src/copmonents/Footer/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </a>
+
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: filterStatus === 'completed',
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setFilterStatus('completed')}
+        >
```
volodymyr-soltys97: Use `Object.values(your created enum)` and render these options with `map()` method

src/App.tsx
```diff
@@ -1,26 +1,157 @@
...
+      setErrorMessage('Unable to add a todo');
+      setTempTodo(null);
+      throw error;
+    } finally {
+      setTempTodo(null);
+      setInputDisabled(false);
+    }
+  }
+
+  async function deleteTodo(todoId: number) {
```
volodymyr-soltys97: It is important to follow the same way of creating functions in all components, fix this everywhere
```suggestion
  const deleteTodo = async (todoId: number) => {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

etojeDenys (approved):

PR #1596, opened 2025-01-20
add task solution

kovaden414:
[DEMO LINK](https://kovaden414.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,339 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import * as todoServise from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import classNames from 'classnames';
+import { TodoInfo } from './components/TodoInfo';
 
-const USER_ID = 0;
+export enum TodoType {
+  all = 'All',
+  active = 'Active',
+  completed = 'Completed',
+}
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  enum Error {
+    loadError = 'Unable to load todos',
+    titleError = 'Title should not be empty',
+    addError = 'Unable to add a todo',
+    deleteError = 'Unable to delete a todo',
+    updateError = 'Unable to update a todo',
+  }
+
+  const [title, setTitle] = useState('');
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,339 @@
...
+  enum Error {
+    loadError = 'Unable to load todos',
+    titleError = 'Title should not be empty',
+    addError = 'Unable to add a todo',
+    deleteError = 'Unable to delete a todo',
+    updateError = 'Unable to update a todo',
+  }
+
+  const [title, setTitle] = useState('');
+  // const [changedTitle, setChangedTitle] = useState('');
```
volodymyr-soltys97: Remove all comments
```suggestion
```

src/App.tsx
```diff
@@ -1,26 +1,339 @@
...
+import { Footer } from './components/Footer';
+import classNames from 'classnames';
+import { TodoInfo } from './components/TodoInfo';
 
-const USER_ID = 0;
+export enum TodoType {
+  all = 'All',
+  active = 'Active',
+  completed = 'Completed',
+}
```
volodymyr-soltys97: Move enum to the separate file

src/App.tsx
```diff
@@ -1,26 +1,339 @@
...
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  enum Error {
+    loadError = 'Unable to load todos',
+    titleError = 'Title should not be empty',
+    addError = 'Unable to add a todo',
+    deleteError = 'Unable to delete a todo',
+    updateError = 'Unable to update a todo',
+  }
```
volodymyr-soltys97: The same as above

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix the problem with the requests because now they are being sent continuously.
image

src/App.tsx
volodymyr-soltys97: Not fixed, move this enum from component to the separate file

src/components/TodoInfo.tsx
```diff
@@ -0,0 +1,65 @@
...
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+          onChange={() => completeTodo(todo.id)}
+        />
+      </label>
+
+      <>
+        <span data-cy="TodoTitle" className="todo__title">
+          {todo.title}
```
volodymyr-soltys97: Use destructuring for `todo`

volodymyr-soltys97 (changes requested):
Hi, the demo link still not fixed from the previous review, you need to make a deploy again and check after the demo link 
image

volodymyr-soltys97 (approved):
Great work 🔥

PR #1631, opened 2025-02-15
solved part #2

NickDevX:
[DEMO LINK](https://NickDevX.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,445 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { addTodo, deleteTodo, getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import clsx from 'clsx';
 
 export const App: React.FC = () => {
+  const input = useRef<HTMLInputElement>(null);
+  const hideErrorTimer = useRef<number | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deletingTodoId, setDeletingTodoId] = useState<number[] | null>(null);
+  const [filter, setFilter] = useState('all');
+  const [title, setTitle] = useState('');
+  const [loadingError, setLoadingError] = useState({
+    queryError: false,
+    addError: false,
+    todosError: false,
+    deleteError: false,
+    updateError: false,
+  });
+
+  const anyExistingError =
+    loadingError.queryError ||
+    loadingError.addError ||
+    loadingError.todosError ||
+    loadingError.deleteError ||
+    loadingError.updateError;
+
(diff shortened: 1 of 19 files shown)


src/App.tsx
```diff
@@ -1,26 +1,445 @@
...
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { addTodo, deleteTodo, getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import clsx from 'clsx';
```
volodymyr-soltys97: Need to install this library

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

src/App.tsx
```diff
@@ -1,26 +1,283 @@
...
+        hideErrorTimer.current = null;
+      }, 3000);
+    },
+    [clearError],
+  );
+
+  const filterTodos = useCallback((todosList: Todo[], filterBy: string) => {
+    if (filterBy === 'active') {
+      return todosList.filter(item => !item.completed);
+    } else if (filterBy === 'completed') {
```
volodymyr-soltys97: Create a enum for 'all', 'active', 'completed' and use it everywhere

src/App.tsx
```diff
@@ -1,26 +1,283 @@
...
+      return todosList.filter(item => item.completed);
+    }
```
volodymyr-soltys97: Consider will be better if you use switch case here instead of if conditions

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+  todos,
+  filter,
+  checkTodoCompleted,
+  removeCompletedTodos,
+  handleClick,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {todos.filter(item => !item.completed).length} items left
```
volodymyr-soltys97: Move this logic from jsx to the helper variable and use it here

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+          Active
+        </a>
+        <a
+          href="#/completed"
+          className={clsx('filter__link', {
+            selected: filter === 'completed',
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={event => handleClick(event, 'completed')}
+        >
```
volodymyr-soltys97: Use `Object.values(your created enum)` and render these options with `map()` method

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+  complateTodo,
+  deletingTodoId,
+  isTemp = false,
+}) => {
+  return (
+    <div
+      data-cy="Todo"
+      className={clsx('todo', { completed: todoItem.completed })}
+    >
+      <label className="todo__status-label" htmlFor={todoItem.id + 't'}>
```
volodymyr-soltys97: It looks strange
```suggestion
      <label className="todo__status-label" htmlFor={todoItem.id}>
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):

PR #1640, opened 2025-02-20
task solution

W3zzie:
[DEMO LINK](https://W3zzie.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -56,9 +56,9 @@ export const App: React.FC = () => {
   }
 
   function onDelete(todoIds: number[]) {
-    setLoadingTodoIds(todoIds);
+    setLoadingTodoIds(prevIds => [...prevIds, ...todoIds]);
 
-    todoIds.map(todoId =>
+    todoIds.forEach(todoId =>
       deleteTodos(todoId)
         .then(() =>
           setTodos(currentTodos =>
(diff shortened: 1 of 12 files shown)


natalia-klonowska (changes requested):
When deleting multiple todos (if the network is slow) loader disappears from the previously deleted todo and appears on the newly clicked one even though previous request is still pending

src/App.tsx
```diff
@@ -1,26 +1,110 @@
...
+  const filteredTodos = todos.filter(todo => {
+    switch (filter) {
+      case TodoStatus.active:
+        return !todo.completed;
+      case TodoStatus.completed:
+        return todo.completed;
+      default:
+        return true;
+    }
+  });
```
natalia-klonowska: you can improve performance by using useMemo

natalia-klonowska (changes requested):

src/App.tsx
```diff
@@ -1,26 +1,112 @@
...
+        if (titleRef.current) {
+          titleRef.current.value = '';
+        }
+      })
+      .catch(() => setErrorMessage(ErrorStatus.ADD_TODO))
+      .finally(() => {
+        setIsLoading(false);
+        setTempTodo(null);
+      });
+  }
```
zibi-dev, line 58: When you are passing functions around as props its good to use `useCallback`

Zibi95 (approved):
Good job. Left one hint for future.

PR #1669, opened 2025-03-14
Develop

nex1994:
[DEMO LINK](https://nex1994.github.io/react_todo-app-add-and-delete/)


natalia-klonowska (changes requested):
- when deleting multiple todos loader disappears for all items as soon as the first one is removed even though others are still in progress
- Currently the same error can't be triggered twice which may cause issues since failed actions don't display error again
- error notification is too big:
image

danon321 (approved):

PR #1670, opened 2025-03-14
Develop

Artem-Brui:
[DEMO LINK](https://Artem-Brui.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/ContextProvider/ContextProvider.tsx
+++ b/src/ContextProvider/ContextProvider.tsx
@@ -58,24 +58,20 @@ const MainContextProvider: FC<{ children: ReactNode }> = ({ children }) => {
     }, 3000);
   }
 
-  return (
-    <MainContext.Provider
-      value={{
-        filter,
-        setFilter,
-        loadingIds,
-        setLoadingIds,
-        todos,
-        setTodos,
-        tempTodo,
-        setTempTodo,
-        error,
-        setError,
-      }}
-    >
-      {children}
-    </MainContext.Provider>
-  );
+  const values = {
+    filter,
+    setFilter,
+    loadingIds,
+    setLoadingIds,
+    todos,
+    setTodos,
+    tempTodo,
+    setTempTodo,
+    error,
+    setError,
+  };
+
+  return <MainContext.Provider value={values}>{children}</MainContext.Provider>;
 };
 
 export default MainContextProvider;
--- a/src/ContextProvider/ContextProvider.tsx
+++ b/src/ContextProvider/ContextProvider.tsx
@@ -1,0 +1,81 @@
+import { createContext, FC, ReactNode, useState } from 'react';
+import { Filter } from '../components/Footer/types';
+import { Todo } from '../types/Todo';
+import { ErrorType } from '../types/Error';
+
+type MainContextType = {
+  filter: Filter;
+  setFilter: React.Dispatch<React.SetStateAction<Filter>>;
+  loadingIds: number[];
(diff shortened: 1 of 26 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,108 @@
...
+const Footer: React.FC = () => {
+  const context = useContext(MainContext);
+  const { todos, setTodos, filter, setFilter, setError, setLoadingIds } =
+    context;
+
+  const [isClearDisabled, setIsClearDisabled] = useState(false);
+
+  useEffect(() => {
+    setIsClearDisabled(!todos.some(todo => todo.completed));
+  }, [todos]);
```
etojeDenys: it's just can be the var
```
const isClearDisabled = !todos.some(todo => todo.completed)
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const [todoState, setTodoState] = useState({
+    isEdited: false,
+    editedValue: title,
+    completed: completed,
+  });
+
+  useEffect(() => {
+    setIsLoading(loadingIds.some(x => x === id));
+  }, [loadingIds, setIsLoading, id]);
```
etojeDenys: same here, it's just can be the var

etojeDenys (changes requested):

src/components/TodoItem/TodoItem.tsx
Artem-Brui: Thanks, solved and use "useRef" here instead of a variable, because I'm change it inside useCallback

src/components/Footer/Footer.tsx
Artem-Brui: Thanks, solved and use "const" here

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,104 @@
...
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: filter === 'FilterLinkCompleted',
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={handleFilterClick}
+        >
+          Completed
+        </a>
```
volodymyr-soltys97: Сreate a enum for 'All', 'Active', 'Completed' and use here `Object.values(your created enum)` and render these options with `map()` method

src/ContextProvider/ContextProvider.tsx
```diff
@@ -0,0 +1,81 @@
...
+        setFilter,
+        loadingIds,
+        setLoadingIds,
+        todos,
+        setTodos,
+        tempTodo,
+        setTempTodo,
+        error,
+        setError,
+      }}
```
volodymyr-soltys97: Move these values to the object and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/Footer/Footer.tsx
Artem-Brui: Thanks, solved.

src/ContextProvider/ContextProvider.tsx
Artem-Brui: Thanks, solved.

vadiimvooo (approved):
Well done

PR #1679, opened 2025-03-22
add and delete

VolodimirKurhan:
(https://VolodimirKurhan.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/types/ErrorType.ts
+++ b/src/types/ErrorType.ts
@@ -1,0 +1,5 @@
+export enum FilterType {
+  All = 'All',
+  active = 'Active',
+  completed = 'Completed',
+}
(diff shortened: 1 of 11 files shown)


src/types/ErrorType.ts
```diff
@@ -0,0 +1,5 @@
+export enum FilterType {
+  All = 'All',
+  active = 'Active',
+  completed = 'Completed',
+}
```
volodymyr-soltys97: Don't need to duplicate enum with the same values 
image
```suggestion
```

volodymyr-soltys97 (changes requested):
Almost done! 
Let's make your code better

vadiimvooo (approved):
Well done

PR #1695, opened 2025-03-28
Develop

fromjkeee16:
[DEMO LINK](https://fromjkeee16.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,178 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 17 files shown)


maponomarenko (approved):
поки не можу затестити Clear completed, бо в списку таких немає, але виглядає наче працює. гуд джоб 👍

src/App.tsx
```diff
@@ -1,26 +1,183 @@
...
+        }
+
+        default: {
+          return true;
+        }
+      }
+    });
   }
 
+  return copy;
```
alexander-ignatow, line 36: Why do you copy multiple times?
Why do you copy at all, when there is no filtering set?

alexander-ignatow (approved):
Good job, but take. look at the following:

1. don't `memo` everything. it makes sense to memoise `TodoItem` because you have multiple instances of it. But memoising other components gains very little performance. On the other hand *memo is not free* - it adds some overheads to the runtime and complicates the app logic.
2. After cleaning up the memo, get rid of the unneded `useCallback` calls

Please, fix it in the next iteration

PR #1697, opened 2025-03-28
solution

kovalmaria:
[DEMO LINK](https://kovalmaria.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import * as todoService from './api/httpClient';
 import { ErrorMessages, FilterStatus, Todo } from './types/Todo';
 import { Header } from './component/Header';
@@ -21,22 +21,24 @@ export const App: React.FC = () => {
   const [query, setQuery] = useState('');
   const [isLoadingTodo, setIsLoadingTodo] = useState<number[]>([]);
 
-  function loadTodos() {
+  const handleErrorMessage = useCallback((message: ErrorMessages) => {
+    setErrorMessage(message);
+    const timeoutId = setTimeout(() => {
+      setErrorMessage(ErrorMessages.DEFAULT);
+    }, 3000);
+
+    return () => clearTimeout(timeoutId);
+  }, []);
+
+  useEffect(() => {
     setIsLoading(true);
 
     todoService
       .getTodos()
       .then(setTodos)
-      .catch(() => {
-        setErrorMessage(ErrorMessages.LOAD_TODOS);
-        setTimeout(() => {
-          setErrorMessage(ErrorMessages.DEFAULT);
-        }, 3000);
-      })
+      .catch(() => handleErrorMessage(ErrorMessages.LOAD_TODOS))
       .finally(() => setIsLoading(false));
-  }
-
-  useEffect(loadTodos, []);
+  }, [handleErrorMessage]);
 
   const filteredTodos = todos.filter(todo => {
     switch (filterStatus) {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,215 @@
...
+    setIsLoading(true);
+
+    todoService
+      .getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage(ErrorMessages.LOAD_TODOS);
+        setTimeout(() => {
+          setErrorMessage(ErrorMessages.DEFAULT);
+        }, 3000);
```
fromjkeee16: u have a lot of such blocks in your code, better set this logic inside of ErrorNotification to avoid code duplication + there should be clearTimeout when component unmounts, because it will lead to a situation when there will be an execution of callback if there's no component mounted

src/App.tsx
```diff
@@ -1,26 +1,215 @@
...
+  );
+  const [isLoading, setIsLoading] = useState(false);
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>(
+    FilterStatus.ALL,
+  );
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [isLoadingTodo, setIsLoadingTodo] = useState<number[]>([]);
+
+  function loadTodos() {
```
fromjkeee16: imho better move creation of this function in corresponding useEffect which calls it to avoid re-creation of a function on each re-render

src/App.tsx
```diff
@@ -1,26 +1,215 @@
...
+  };
+
+  const handleChangeCompletedAllTodos = () => {
+    const areAllCompleted = todos.every(todo => todo.completed);
+
+    setTodos(
+      todos.map(todo => ({
+        ...todo,
+        completed: !areAllCompleted,
+      })),
```
fromjkeee16: you should send requests to the server to change affected todos because this handler is frontend-only (no changes after reloading the page)

src/component/Header.tsx
```diff
@@ -0,0 +1,73 @@
...
+    event.preventDefault();
+    handleCreateTodo();
+  };
+
+  return (
+    <header className="todoapp__header">
+      <button
+        type="button"
+        className={classNames('todoapp__toggle-all', {
+          active: todos.every(todo => todo.completed),
```
fromjkeee16: it is better to move this logic outside of return render function

fromjkeee16 (changes requested):
please remove code duplication
everything else looks +- ok
btw try to use React.memo and useCallback() for optimization

src/App.tsx
```diff
@@ -1,26 +1,228 @@
...
+  }, [handleErrorMessage]);
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filterStatus) {
+      case FilterStatus.ACTIVE:
+        return !todo.completed;
+      case FilterStatus.COMPLETED:
+        return todo.completed;
+      default:
+        return true;
```
ArturErmolenko, line 50: where is ALL case?

src/App.tsx
```diff
@@ -1,26 +1,228 @@
...
+  const [query, setQuery] = useState('');
+  const [isLoadingTodo, setIsLoadingTodo] = useState<number[]>([]);
+
+  const handleErrorMessage = useCallback((message: ErrorMessages) => {
+    setErrorMessage(message);
+    const timeoutId = setTimeout(() => {
+      setErrorMessage(ErrorMessages.DEFAULT);
+    }, 3000);
+
+    return () => clearTimeout(timeoutId);
```
ArturErmolenko, line 30: let's hande timer i useEffect in Error component

ArturErmolenko (approved):
LGTM 👍

PR #1699, opened 2025-03-29
solution

dmytroKolesnikov52:
[DEMO LINK](https://Mazafakerqq.github.io/react_todo-app-add-and-delete/)
Мені було складно зрозуміти що саме потребується в кожному із завдань, тому я зробив все завдання.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,464 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { dataTodos, USER_ID } from './api/todos';
+import { Filter, Todo, TodoInput } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [todoTitle, setTodoTitle] = useState<string>('');
+  const [filter, setFilter] = useState(Filter.All);
+  const [loadingAllTodos, setLoadingAllTodos] = useState<boolean>(false);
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,464 @@
...
-const USER_ID = 0;
+import { dataTodos, USER_ID } from './api/todos';
+import { Filter, Todo, TodoInput } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [todoTitle, setTodoTitle] = useState<string>('');
+  const [filter, setFilter] = useState(Filter.All);
+  const [loadingAllTodos, setLoadingAllTodos] = useState<boolean>(false);
```
reb0rned, line 17: You can use one loader for all loading states, for example in your state below.

src/App.tsx
```diff
@@ -1,26 +1,464 @@
...
+  const [loadingTodoId, setLoadingTodoId] = useState<number[] | null>(null);
+  const [editingTitle, setEditingTitle] = useState<string>('');
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    dataTodos
+      .getTodos()
```
reb0rned, line 26: it is better to put the logic in a separate function instead of writing it inside  useEffect

src/App.tsx
```diff
@@ -1,26 +1,464 @@
...
+          setErrorMessage('Unable to delete some todos');
+        }
+      })
+      .catch(() => {
+        setErrorMessage('Unable to delete a todo');
+      })
+      .finally(() => {
+        setLoadingTodoId(null);
+        setTimeout(() => {
+          if (inputRef.current) {
```
reb0rned: The block of code for setting the focus can be extracted into a separate function for better readability and reusability

reb0rned (changes requested):
Good work, but the main problem is that everything is written in one file, this is bad practice for scaling a project, you need to divide everything into components. Also small improvements below

ArturErmolenko (approved):
Я тут ставлю апрув , го фіксити коменти в 3 частині

PR #1714, opened 2025-04-02
Solution 2 part

Bankir4ik87:
[DEMO LINK](https://Bankir4ik87.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,104 @@
+import React, { useCallback, useContext } from 'react';
+import { Filter } from './types';
+import classNames from 'classnames';
+import { MainContext } from '../../ContextProvider/ContextProvider';
+import { deleteTodo } from '../../api/todos';
+import callError from '../../utils/callError';
+
+const Footer: React.FC = () => {
+  const context = useContext(MainContext);
+  const { todos, setTodos, filter, setFilter, setError, setLoadingIds } =
+    context;
+
+  const isClearDisabled = !todos.some(todo => todo.completed);
+
+  const handleFilterClick = useCallback(
+    (event: React.MouseEvent): void => {
+      const element = event.target as HTMLElement;
+
+      setFilter(element.dataset.cy as Filter);
+    },
+    [setFilter],
+  );
+
+  const handleClearCompletedClick = useCallback(() => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    setLoadingIds(completedTodos.map(todo => todo.id));
+
+    Promise.allSettled(completedTodos.map(todo => deleteTodo(todo.id)))
+      .then(res => {
+        setLoadingIds([0]);
+
+        const deletedTodos = completedTodos.filter((_t, index) => {
+          return res[index].status === 'fulfilled';
+        });
+
+        if (deletedTodos.length < completedTodos.length) {
+          callError(setError, 'delete');
+        }
+
+        setTodos(todos.filter(todo => !deletedTodos.includes(todo)));
+      })
+      .catch(() => callError(setError, 'delete'));
+  }, [todos, setLoadingIds, setError, setTodos]);
+
+  const activeNumbers = todos.filter(todo => !todo.completed).length;
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeNumbers} items left`}
+      </span>
+
+      {/* Active link should have the 'selected' class */}
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={classNames('filter__link', {
+            selected: filter === 'FilterLinkAll',
+          })}
+          data-cy="FilterLinkAll"
+          onClick={handleFilterClick}
+        >
+          All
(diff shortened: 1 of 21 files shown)


src/components/Footer/types.ts
```diff
@@ -0,0 +1,4 @@
+export type Filter =
```
TarasHoliuk: nit: I'd say enum is more suitable here 🙂

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,96 @@
...
+} from 'react';
+import classNames from 'classnames';
+import callError from '../../utils/callError';
+import { addTodo, USER_ID } from '../../api/todos';
+import { MainContext } from '../../ContextProvider/ContextProvider';
+
+const Header: React.FC = ({}) => {
+  const context = useContext(MainContext);
+  const { todos, setTodos, setError, loadingIds, setLoadingIds, setTempTodo } =
+    context;
```
TarasHoliuk:
```suggestion
  const { todos, setTodos, setError, loadingIds, setLoadingIds, setTempTodo } =
    useContext(MainContext);
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const isAllActive = todos.every(todo => todo.completed);
+
+  useEffect(() => {
+    if (focusedInput.current) {
+      const input = focusedInput.current as HTMLElement;
+
+      input.focus();
+    }
+  }, [isDisabled, todos]);
```
TarasHoliuk: Do you need isDisabled here?

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const Header: React.FC = ({}) => {
+  const context = useContext(MainContext);
+  const { todos, setTodos, setError, loadingIds, setLoadingIds, setTempTodo } =
+    context;
+
+  const [todoInputValue, setTodoInputValue] = useState('');
+  const [isDisabled, setIsDisabled] = useState(false);
+
+  const focusedInput = useRef(null);
```
TarasHoliuk:
```suggestion
  const inputRef = useRef(null);
```
It's ref with input as a value. It's not always focused

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+    completed: completed,
+  });
+
+  const isLoading = useRef(loadingIds.some(x => x === id));
+
+  if (isLoading.current) {
+    setTimeout(() => (isLoading.current = false), 3000);
+  }
+
+  useEffect(() => {
```
TarasHoliuk: you don't need this

Instead, you can add `onBlur` listener to input. And on blur just do some action (but this part in the next task) and change `isEdited` to false

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+  todo: Todo;
+};
+
+const TodoItem: React.FC<TodoProps> = ({ todo }) => {
+  const context = useContext(MainContext);
+  const { todos, setTodos, setError, loadingIds } = context;
+
+  const { id, title, completed } = todo;
+
+  const [todoState, setTodoState] = useState({
```
TarasHoliuk: usually, we have a separate useState for each value if there is no certain reason to join them into one object

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+        onClick={handleDeleteClick}
+      >
+        ×
+      </button>
+
+      {/* overlay will cover the todo while it is being deleted or updated */}
+      <div
+        data-cy="TodoLoader"
+        className={classNames('modal overlay', {
+          'is-active': isLoading.current,
```
TarasHoliuk: Why ref and not state? to store isLoading value?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const { todos, setTodos, filter, setFilter, setError, setLoadingIds } =
+    context;
+
+  const isClearDisabled = !todos.some(todo => todo.completed);
+
+  const handleFilterClick = useCallback(
+    (event: React.MouseEvent): void => {
+      const element = event.target as HTMLElement;
+
+      setFilter(element.dataset.cy as Filter);
```
TarasHoliuk: It's a bad practice to use data.cy value for it

TarasHoliuk (changes requested):
GJ! Let's improve your code a bit

src/components/Header/Header.tsx
Bankir4ik87: No, I most likely won't need isDisabled in the dependency array since it doesn't directly affect the behavior of the useEffect logic.

src/components/TodoItem/TodoItem.tsx
Bankir4ik87: The fields are logically related:
- isEdited indicates whether the task is being edited.
- editedValue stores the current value of the edited field.
- completed indicates whether the task has completed.

vadiimvooo (approved):
Well done

PR #1717, opened 2025-04-03
Solution

Olehgoon093:
[DEMO LINK](https://Olehgoon093.github.io/react_todo-app-add-and-delete/)


src/src2/src/components/toDoRow.tsx
```diff
@@ -0,0 +1,165 @@
...
+      }
+    } else {
+      if (onDelete) {
+        await onDelete();
+      }
+    }
+
+    setLoading(false);
+  };
+
```
ansofidev, line 82: `handleDeleteTodo` changes loading, otherwise deleteTodos does not return to `await.` You need to wrap `try/catch` and wrap `await.`

ansofidev:
First of all, it is not clear why you need the second scr folder, you just need to copy the files from the first part to this.

src/src2/src/App.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+      return;
+    }
+
+    if (tempTodo !== null) {
+      return;
+    }
+
+    setTempTodo({
+      id: 0,
```
ansofidev, line 76: When tempTodo is created, userId: 11 is installed, and USER_ID is added.

src/src2/src/App.tsx
```diff
@@ -0,0 +1,204 @@
...
+      completed: false,
+      title,
+      userId: 11,
+    });
+
+    createTodo(title)
```
ansofidev, line 82: userId is not transmitted, it may be in the body of the POST request

src/src2/src/App.tsx
```diff
@@ -0,0 +1,204 @@
...
+      .then(todo => {
+        setTodos(currentTodos => [...currentTodos, todo]);
+      })
+      .catch(error => {
+        showErorr('Unable to create a todo');
+        throw error;
+      })
```
ansofidev, line 89: The text in the field is not updated, but must, according to the assignment, keep the text in case of error.

src/src2/src/App.tsx
```diff
@@ -0,0 +1,204 @@
...
+      .finally(() => {
+        setTempTodo(null);
```
ansofidev, line 91: need to click` inputRef?.current?.focus();`

src/src2/src/App.tsx
```diff
@@ -0,0 +1,204 @@
...
+            </a>
+          </nav>
+
+          {/* this button should be disabled if there are no completed todos */}
+          <button
+            type="button"
+            className="todoapp__clear-completed"
+            data-cy="ClearCompletedButton"
+          >
+            Clear completed
```
ansofidev, line 179: There is no logic to delete all the ones executed there but not completed - the deletion should work as several separate deletions started simultaneously.

ansofidev (changes requested):
Please remove the commented code, it still doesn't work DEMO link :(

alexander-ignatow (approved):
It is not perfect, but let's focus on the final iteration

PR #1727, opened 2025-04-10
Solution

IvanSolopenko:
[DEMO LINK](https://IvanSolopenko.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,355 @@
...
+
+              if (!todoTitle.trim()) {
+                setTitleErrorMessage('Title should not be empty');
+                titleField.current?.focus();
+
+                return;
+              }
+
+              addTodo({ title: todoTitle.trim(), completed: false });
+              setTitleErrorMessage('');
```
volodymyr-soltys97: Move this logic to the separate `Header` component

src/App.tsx
```diff
@@ -1,26 +1,355 @@
...
+              event.preventDefault();
+
+              if (!todoTitle.trim()) {
+                setTitleErrorMessage('Title should not be empty');
+                titleField.current?.focus();
+
+                return;
+              }
+
+              addTodo({ title: todoTitle.trim(), completed: false });
```
volodymyr-soltys97: Move this logic from jsx to the helper function and use it here

src/App.tsx
```diff
@@ -1,26 +1,355 @@
...
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => {
+            setTitleErrorMessage('');
+            setAddErrorMessage('');
+            setDeleteErrorMessage('');
+            setUpdateErrorMessage('');
+            setLoadingErrorMessage('');
+          }}
```
volodymyr-soltys97: The same as above

src/components/Footer.tsx
```diff
@@ -0,0 +1,78 @@
...
+    href: '#/active',
+    dataCy: 'FilterLinkActive',
+  },
+  {
+    label: 'Completed',
+    value: FilterType.completed,
+    href: '#/completed',
+    dataCy: 'FilterLinkCompleted',
+  },
+];
```
volodymyr-soltys97: Don't need to create additional array, use `Object.values(FilterType)` instead

src/components/TodoList.tsx
```diff
@@ -0,0 +1,110 @@
...
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
+      )}
+
+      {/* This todo is in loadind state */}
+      {false && (
+        <div data-cy="Todo" className="todo">
+          <label className="todo__status-label">
```
volodymyr-soltys97: Looks redundant, remove it

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/Footer.tsx
```diff
@@ -0,0 +1,75 @@
...
+  [FilterType.all]: '#/',
+  [FilterType.active]: '#/active',
+  [FilterType.completed]: '#/completed',
+};
+
+const dataCyMap: Record<FilterType, string> = {
+  [FilterType.all]: 'FilterLinkAll',
+  [FilterType.active]: 'FilterLinkActive',
+  [FilterType.completed]: 'FilterLinkCompleted',
+};
```
volodymyr-soltys97: Don't need to create 3 array, use only `Object.value(FilterType)` for creating one array and you can use slice method inside. If you need some help feel free in the fe_chat

volodymyr-soltys97 (changes requested):
Almost done!
Let's make your code better

anastasiiavorobiova (approved):
Awesome!

PR #1737, opened 2025-04-28
error fail

pryge:
[DEMO LINK](https://pryge.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,131 +1,111 @@
-import React, { useEffect, useMemo, useState } from 'react';
-import cn from 'classnames';
-
-import { Todo } from './types/Todo';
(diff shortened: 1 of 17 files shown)


src/App.tsx
```diff
@@ -1,26 +1,111 @@
...
+    }
+
+    const timer = setTimeout(() => {
+      setCurrentError('');
+    }, 3000);
+
+    return () => clearTimeout(timer);
+  }, [currentError]);
+
+  const activeTodos: number = [...todos].filter(
```
etojeDenys, line 55:
```suggestion
  const activeTodos: number = todos.filter(
```
the filter method doesn't mutate the array

etojeDenys (approved):

PR #1738, opened 2025-04-28
https://andrew1256.github.io/react_todo-app-add-and-delete/

Andrew1256:


src/components/Footer.tsx
```diff
@@ -0,0 +1,65 @@
...
+  selected,
+  setSelected,
+  clearAll,
+  completedTodos,
+}) => {
+  const filters = [
+    { title: 'All', type: Selected.All, href: '#/' },
+    { title: 'Active', type: Selected.Active, href: '#/active' },
+    { title: 'Completed', type: Selected.Completed, href: '#/completed' },
+  ];
```
VitaliyBondarenko1982, line 26: it can be constant outside of component. in component such array will be created each render.

src/components/Header.tsx
```diff
@@ -0,0 +1,52 @@
...
+        className="todoapp__toggle-all"
+        data-cy="ToggleAllButton"
+        onClick={updateAll}
+      ></button>
+
+      <form
+        onSubmit={e => {
+          e.preventDefault();
+          handleAdd();
+        }}
```
VitaliyBondarenko1982, line 37:
```suggestion
        onSubmit={handleAdd}
```

you can make handleAdd as function which except event.

```
  const handleAdd = (e: FormEvent) => {
     e.preventDefault();
     // do something
  }
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+          checked={todo.completed}
+          readOnly
+        />
+      </label>
+
+      {updatingId === todo.id ? (
+        <form
+          onSubmit={e => {
+            e.preventDefault();
+            handleSave(todo.id);
```
VitaliyBondarenko1982, line 53: same as in prev comment with handleAdd.
In general handleSave can live in TodoItem component and not need todo.id as argument.

VitaliyBondarenko1982 (approved):
Looks good.
Please check comment  and you can fix them in next task.

PR #1745, opened 2025-05-07
task solution

BoGoBooM:
[DEMO LINK](https://bogoboom.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,197 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { FormEvent, useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,26 +1,197 @@
...
+
+  const showErrorContainer = (errorMessage: string) => {
+    setError(errorMessage);
+    setTimeout(() => {
+      setError('');
+    }, 3000);
+  };
+
+  useEffect(() => {
+    async function fetchData() {
```
volodymyr-soltys97, line 54: It is important to follow the same way of creating functions everywhere
```suggestion
    const fetchData = async () => {
```

volodymyr-soltys97 (changes requested):
Hi, need to fix data attributes 
image

anastasiiavorobiova (approved):
Great work!

PR #1747, opened 2025-05-12
Dev

DrDakka:
Unfortunatelly, I am unable to deploy my solution atm due to some technical issues

### DIFF
--- a/src/components/ErrorMsg/ErrorMsg.tsx
+++ b/src/components/ErrorMsg/ErrorMsg.tsx
@@ -1,4 +1,5 @@
 import { useEffect } from 'react';
+import classNames from 'classnames';
 
 type Props = {
   error: string;
@@ -26,10 +27,12 @@ export const ErrorMsg: React.FC<Props> = ({ error, setError }) => {
   return (
     <div
       data-cy="ErrorNotification"
-      className={
-        `notification is-danger is-light has-text-weight-normal` +
-        (error ? '' : ' hidden')
-      }
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal',
+        {
+          hidden: error === '',
+        },
+      )}
     >
       <button
         data-cy="HideErrorButton"
--- a/src/components/ErrorMsg/ErrorMsg.tsx
+++ b/src/components/ErrorMsg/ErrorMsg.tsx
@@ -1,0 +1,43 @@
+import { useEffect } from 'react';
+
+type Props = {
+  error: string;
+  setError: React.Dispatch<React.SetStateAction<string>>;
+};
+
+export const ErrorMsg: React.FC<Props> = ({ error, setError }) => {
+  useEffect(() => {
+    if (error === '') {
+      return;
+    }
+
+    const timer = setTimeout(() => {
+      setError('');
+    }, 3000);
+
+    return () => clearTimeout(timer);
+  }, [error]);
+
+  const closeErrorMsg = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    setError('');
+  };
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={
+        `notification is-danger is-light has-text-weight-normal` +
+        (error ? '' : ' hidden')
+      }
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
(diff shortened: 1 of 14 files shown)


DrDakka:
Deploy complete
https://drdakka.github.io/react_todo-app-add-and-delete/

src/components/ErrorMsg/ErrorMsg.tsx
```diff
@@ -0,0 +1,43 @@
...
+    if (error === '') {
+      return;
+    }
+
+    const timer = setTimeout(() => {
+      setError('');
+    }, 3000);
+
+    return () => clearTimeout(timer);
+  }, [error]);
```
volodymyr-soltys97, line 20: Fix linter errors 
image

src/components/ErrorMsg/ErrorMsg.tsx
```diff
@@ -0,0 +1,43 @@
...
+    event.preventDefault();
+    setError('');
+  };
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={
+        `notification is-danger is-light has-text-weight-normal` +
+        (error ? '' : ' hidden')
```
volodymyr-soltys97: Use the classnames library for add classes with condition, fix it everywhere

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+    event.preventDefault();
+    await addTodo();
+  };
+
+  return (
+    <header className="todoapp__header">
+      {todos.length > 0 && (
+        <button
+          type="button"
+          className={`todoapp__toggle-all${todos.every(todo => todo.completed === true) ? ' active' : ''}`}
```
volodymyr-soltys97: Also, move this logic to the helper variable ans use classnames
```suggestion
          className={`todoapp__toggle-all${todos.every(todo => todo.completed) ? ' active' : ''}`}
```

src/components/TodoItem/TempTodo.tsx
```diff
@@ -0,0 +1,38 @@
...
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={false}
+        />
+      </label>
+
+      <span data-cy="TodoTitle" className="todo__title">
+        {todo.title}
```
volodymyr-soltys97: Use destructuring for `todo`

src/hooks/general.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  const visibleTodos = useMemo(() => {
+    switch (filter) {
+      case 'all':
+        return todosFromServer;
+
+      case 'active':
+        return todosFromServer.filter((todo: Todo) => !todo.completed);
+
+      case 'completed':
```
volodymyr-soltys97: Create a enum for 'all', 'active', 'completed' and use it everywhere

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+        </a>
+
+        <a
+          href="#/completed"
+          onClick={() => setFilter('completed')}
+          className={`filter__link${filter === 'completed' ? ' selected' : ''}`}
+          data-cy="FilterLinkCompleted"
+        >
+          Completed
+        </a>
```
volodymyr-soltys97: Use `Object.values(your created enum)` and render these options with `map()` method

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/ErrorMsg/ErrorMsg.tsx
DrDakka: It will cause setting timeout twice. Error here is a state and it is being pushed to the element by props right from the hooks file. The dependencies are correct, no need to use a setError as dependency here. Also, according to https://legacy.reactjs.org/docs/hooks-reference.html#usestate   state "React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list."

src/components/Footer/Footer.tsx
DrDakka: I admit I've missed it completely :D

volodymyr-soltys97 (approved):
Great!

PR #1753, opened 2025-05-19
react_todo-app-add-and-delete

MRKucherenko:
[DEMO LINK](https://MRKucherenko.github.io/react_todo-app-add-and-delete/)


src/types/Todo.ts
```diff
@@ -0,0 +1,6 @@
+export interface Todo {
+  id?: number;
```
TarasHoliuk, line 2: Should everything be optional here? - I don't think so. You can create a correct interface and use Partial, Omit, Pick utility types where you need

TarasHoliuk (approved):
Well done - approved ✅ 

Added one small comment FYI (apply it in the next task)

PR #1756, opened 2025-05-20
commit 1.0

KvashaIhor:
[DEMO LINK](https://kvashaihor.github.io/react_todo-app-add-and-delete/)


VitaliyBondarenko1982 (approved):
Looks good.

During deleting multiply todos loader should be visible on each todo which in process on server now.
Fix it in next task.

PR #1765, opened 2025-05-29
Solution

Farrelzum:
[DEMO LINK](https://farrelzum.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,380 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { deleteTodo, getTodos, postTodo, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [filter, setFilter] = useState<Filter>('All');
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [inputValue, setInputValue] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deleteTodoById, setDeleteTodoById] = useState<number[]>([]);
+  const [isAddingTodo, setIsAddingTodo] = useState(false);
+  const [shouldFocusInput, setShouldFocusInput] = useState(false);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    let newTodos = [...todos];
+
+    switch (filter) {
+      case 'All':
+        break;
+      case 'Active':
+        newTodos = todos.filter(todo => !todo.completed);
+        break;
+      case 'Completed':
+        newTodos = todos.filter(todo => todo.completed);
+        break;
+    }
+
+    setFilteredTodos(newTodos);
+  }, [filter, todos]);
+
+  useEffect(() => {
+    setErrorMessage(null);
+
+    getTodos()
+      .then(data => {
+        setTodos(data);
+      })
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+
+        setTimeout(() => {
+          setErrorMessage(null);
+        }, 3000);
+      });
+  }, []);
+
+  useEffect(() => {
+    if (shouldFocusInput && inputRef.current) {
+      inputRef.current.focus();
+      setShouldFocusInput(false); // Zresetuj flagę, aby nie fokusować ponownie przy każdym renderowaniu
+    }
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,26 +1,380 @@
...
-
-const USER_ID = 0;
+import { deleteTodo, getTodos, postTodo, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [filter, setFilter] = useState<Filter>('All');
```
natalia-klonowska, line 17: It's better to avoid hardcoding filter values like 'All', 'Active', and 'Completed' directly in useState, setFilter and switch cases. Instead consider defining them as union type or enum. This makes the code easier to maintain and less error-prone especially when these values are used in multiple places or need to be updated.

src/App.tsx
```diff
@@ -1,26 +1,380 @@
...
+
+  useEffect(() => {
+    setErrorMessage(null);
+
+    getTodos()
+      .then(data => {
+        setTodos(data);
+      })
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```
natalia-klonowska: Instead of passing arbitrary values into setErrorMessage define fixed set of possible error types to ensure consistency and maintainability

natalia-klonowska (changes requested):
Great work! You should split your App into smaller components to improve code readability and makes it easier to maintain. It also helps isolate concerns allowing each part of the UI to manage its own logic and state more cleanly.

src/types/Filter.ts
```diff
@@ -0,0 +1,10 @@
+export const FilterValues = {
+  ALL: 'All' as const,
+  ACTIVE: 'Active' as const,
+  COMPLETED: 'Completed' as const,
+};
+
+export type Filter =
+  | typeof FilterValues.ALL
+  | typeof FilterValues.ACTIVE
+  | typeof FilterValues.COMPLETED;
```
zibi-dev, line 10:
```suggestion
export type FilterValue = typeof FilterValues[keyof typeof FilterValues];
```

src/types/Filter.ts
```diff
@@ -0,0 +1,10 @@
+export const FilterValues = {
+  ALL: 'All' as const,
+  ACTIVE: 'Active' as const,
+  COMPLETED: 'Completed' as const,
+};
```
zibi-dev:
```suggestion
export const FilterValues = {
  ALL: 'All',
  ACTIVE: 'Active',
  COMPLETED: 'Completed',
} as const;
```

src/components/ToDoList.tsx
```diff
@@ -0,0 +1,109 @@
...
+                data-cy="TodoDelete"
+                onClick={() => handleDeleteTodo(todo.id)}
+                disabled={deleteTodoById.includes(todo.id)}
+              >
+                ×
+              </button>
+
+              <div
+                data-cy="TodoLoader"
+                className={`modal overlay${deleteTodoById.includes(todo.id) ? ' is-active' : ''}`}
```
zibi-dev: You have library installed that helps you handle conditional classes

zibi-dev: its called `"classnames": "^2.5.1",`

src/components/ToDoList.tsx
```diff
@@ -0,0 +1,109 @@
...
+          </button>
+
+          {/* 'is-active' class puts this modal on top of the todo */}
+          <div data-cy="TodoLoader" className="modal overlay is-active">
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
+      )}
+    </>
```
zibi-dev: that markup for todo looks the same. So its good case to create new component

src/App.tsx
```diff
@@ -1,26 +1,237 @@
...
+import { Footer } from './components/Footer';
+import { ErrorType } from './types/ErrorType';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<ErrorType | null>(null);
+  const [filter, setFilter] = useState<Filter>('All');
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [inputValue, setInputValue] = useState('');
```
zibi-dev: Be aware that each keystroke is rendering whole App right now

zibi-dev: It would be good to move that into lower component where its used.

src/App.tsx
```diff
@@ -1,26 +1,237 @@
...
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [deleteTodoById, setDeleteTodoById] = useState<number[]>([]);
+  const [isAddingTodo, setIsAddingTodo] = useState(false);
+  const [shouldFocusInput, setShouldFocusInput] = useState(false);
```
zibi-dev: That state seems not needed. You can just create funtion `focusInput` and call that in place where it should focus.

Zibi95 (changes requested):

Zibi95 (approved):

PR #1767, opened 2025-05-30
solution

maxokol:
https://maxokol.github.io/react_todo-app-add-and-delete/


maxokol:
@mateacademy-ai-mentor the newTitle is trimmed in TodoList.tsx so addTodo function receives already trimmed value

PR #1776, opened 2025-06-10
solution

dennispaixao:
[DEMO LINK](https://dennispaixao.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/utils/fetchClient.ts
+++ b/src/utils/fetchClient.ts
@@ -1,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const BASE_URL = 'https://mate.academy/students-api';
+
+// returns a promise resolved after a given delay
+function wait(delay: number) {
+  return new Promise(resolve => {
+    setTimeout(resolve, delay);
+  });
+}
+
+// To have autocompletion and avoid mistypes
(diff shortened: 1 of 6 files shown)


src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
DanielCamposDevX: remove lint-disabler
u can type every function here

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+}
+
+// To have autocompletion and avoid mistypes
+type RequestMethod = 'GET' | 'POST' | 'PATCH' | 'DELETE';
+
+function request<T>(
+  url: string,
+  method: RequestMethod = 'GET',
+  data: any = null, // we can send any data to the server
```
DanielCamposDevX: You need to send only the right data ( you can send any data, but you need to have the correct one on the type)

DanielCamposDevX (changes requested):

pedro-ruas (approved):

DanielCamposDevX (approved):

PR #1779, opened 2025-06-12
add and delete

zbvika:
[DEMO LINK](https://zbvika.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -21,8 +21,6 @@ export const Error: React.FC<Props> = ({ errorMessage, setErrorMessage }) => {
         onClick={() => setErrorMessage(null)}
       />
       {errorMessage}
-
-      {/* Unable to update a todo */}
     </div>
   );
 };
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -1,0 +1,28 @@
+import cn from 'classnames';
+import React from 'react';
+
+interface Props {
+  errorMessage: string | null;
+  setErrorMessage: (error: string | null) => void;
+}
+
+export const Error: React.FC<Props> = ({ errorMessage, setErrorMessage }) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={cn('notification is-danger is-light has-text-weight-normal', {
+        hidden: !errorMessage,
+      })}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => setErrorMessage(null)}
+      />
+      {errorMessage}
+
+      {/* Unable to update a todo */}
+    </div>
+  );
+};
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,10 +1,11 @@
 import React from 'react';
-import classNames from 'classnames';
+import cn from 'classnames';
 import { FilterType } from '../../types/FilterType';
+import { FilterTypeValues } from '../../types/FilterType';
 
 interface Props {
-  filter: FilterType;
-  setFilter: (filter: FilterType) => void;
+  filter: FilterTypeValues;
+  setFilter: (filter: FilterTypeValues) => void;
   countOfActiveTodos: number;
   deleteCompletedTodos: () => void;
   hasCompletedTodos: boolean;
@@ -24,19 +25,17 @@ export const Footer: React.FC<Props> = ({
       </span>
 
       <nav className="filter" data-cy="Filter">
-        {Object.keys(FilterType).map(type => (
+        {Object.entries(FilterType).map(([text, value]) => (
           <a
-            key={type}
+            key={value}
             href="#/"
-            className={classNames('filter__link', {
-              selected: filter === FilterType[type as keyof typeof FilterType],
+            className={cn('filter__link', {
+              selected: filter === value,
             })}
-            data-cy={`FilterLink${FilterType[type as keyof typeof FilterType]}`}
-            onClick={() =>
-              setFilter(FilterType[type as keyof typeof FilterType])
-            }
+            data-cy={`FilterLink${text}`}
+            onClick={() => setFilter(value)}
           >
-            {FilterType[type as keyof typeof FilterType]}
+            {text}
           </a>
         ))}
       </nav>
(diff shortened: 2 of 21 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,49 @@
...
+        data-cy="ToggleAllButton"
+      />
+
+      <form
+        onSubmit={e => {
+          e.preventDefault();
+          addTodo(title)
+            .then(() => setTitle(''))
+            .catch(() => {});
+        }}
```
ystyslo: Maybe it'll be better to do handler for submit?

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,49 @@
+import { useEffect, useRef, useState } from 'react';
+
+interface Props {
+  addTodo: (title: string) => Promise<void>;
+  isInputDisabled: boolean;
+  errorMessage: string | null;
```
ystyslo: also you aren't using this prop

src/components/Header/Header.tsx
zbvika: If I understood you correctly, I didn’t create a handler to avoid adding more code.
The idea was that since we’re only accepting a single parameter and performing just one check, we can create the object inline.
The code works the same way as it would if we had created a handler.

ystyslo (approved):
Nice

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,28 @@
...
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => setErrorMessage(null)}
+      />
+      {errorMessage}
+
+      {/* Unable to update a todo */}
```
BudnikOleksii: Remove comments

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,55 @@
...
+            href="#/"
+            className={classNames('filter__link', {
+              selected: filter === FilterType[type as keyof typeof FilterType],
+            })}
+            data-cy={`FilterLink${FilterType[type as keyof typeof FilterType]}`}
+            onClick={() =>
+              setFilter(FilterType[type as keyof typeof FilterType])
+            }
+          >
+            {FilterType[type as keyof typeof FilterType]}
```
BudnikOleksii: Avoid `as` and you repeat `keyof typeof FilterType` 3 times.
You can create a [type guard ](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) to validate type
image

src/components/Header/Header.tsx
BudnikOleksii: It works the same, but with handlers it's much easier to read as you separate logic from presentation.
This function is performing two actions(prevent default browser behaviour and set a request).
Also the catch block is empty, you should handle errors in this block and notify user when an error occurs

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import classNames from 'classnames';
+
+interface Props {
+  isTempTodo: Todo | null;
```
BudnikOleksii:
```suggestion
  todo: Todo;
```
You shouldn't render `TodoItem` at all if you don't have one. And it's not a flag, so the name is confusing

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,66 @@
...
+  isTempTodo,
+  isTodoDeleted,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos?.map(todo => {
+        return (
+          <div
+            key={todo.id}
+            data-cy="Todo"
```
BudnikOleksii: You have a `TodoItem` component

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,127 @@
+import { useEffect, useState } from 'react';
+import * as todoService from '../api/todos';
+import { Todo } from '../types/Todo';
+
+export function useTodo() {
+  const [data, setData] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [isTempTodo, setIsTempTodo] = useState<Todo | null>(null);
```
BudnikOleksii:
```suggestion
  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
```

src/hooks/useTodo.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    setIsTodoDeleted(null);
+
+    return Promise.allSettled(
+      completedIds.map(id => todoService.deleteTodos(id).then(() => id)),
+    )
+      .then(results => {
+        const successIds = results
+          .filter(r => r.status === 'fulfilled')
+          .map(r => (r as PromiseFulfilledResult<number>).value);
```
BudnikOleksii: Avoid one letter names

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoItem } from '../TodoItem';
+import classNames from 'classnames';
+
+interface Props {
+  todos: Todo[] | null;
```
BudnikOleksii, line 6:
```suggestion
  todos: Todo[];
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import cn from 'classnames';
+
+interface Props {
+  todo: Todo;
+  deleteTodo: (todoId: number) => void;
+  isDeleting: boolean;
+  isTemporary?: boolean;
```
BudnikOleksii, line 10:
```suggestion
  isProcessing: boolean;
```
No need to add such flag as `isTemporary` it should be just regular todo.
And `isDeleting` is a specific name, in the next task you will another functionality when you need to show loader so you can name as `isLoading` or `isProcessing`

src/components/TodoList/TodoList.tsx
BudnikOleksii: You don't need to render `TodoList` if you don't have todos

BudnikOleksii (approved):
Good work 👍 
Left a few suggestions for the next task

PR #1788, opened 2025-06-12
solution w too many crutches and failing tests

yelyzavetaz:
[DEMO LINK](https://yelyzavetaz.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,6 +8,9 @@ import { TodoList } from './components/TodoList/TodoList';
 import { ErrorComponent } from './components/ErrorComponent/ErrorComponent';
 import { Footer } from './components/Footer/Footer';
 import { FilterStatusType } from './types/FilterStatusType';
+import cn from 'classnames';
+import { FormComponent } from './components/FormComponent/FormComponent';
+import { ErrorMessage } from './types/ErrorStatusType';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
@@ -17,6 +20,11 @@ export const App: React.FC = () => {
   const [isCompletedTodosExist, setIsCompletedTodosExist] = useState(true);
   const [numberOfNotCompletedTodos, setNumberOfNotCompletedTodos] = useState(0);
   const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [needToRefresh, setNeedToRefresh] = useState(true);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isTempTodoLoading, setIsTempTodoLoading] = useState(false);
+  const [todoIdsToDelete, setTodoIdsToDelete] = useState<number[]>([]);
+  const [isFocusTitleInput, setIsFocusTitleInput] = useState(false);
 
   useEffect(() => {
     if (showError) {
@@ -45,27 +53,33 @@ export const App: React.FC = () => {
   }, [todos]);
 
   useEffect(() => {
-    if (filterStatus === FilterStatusType.All) {
-      setVisibleTodos(todos);
-    } else if (filterStatus === FilterStatusType.Active) {
-      setVisibleTodos(todos.filter(todo => !todo.completed));
-    } else {
-      setVisibleTodos(todos.filter(todo => todo.completed));
+    switch (filterStatus) {
+      case FilterStatusType.Active:
+        setVisibleTodos(todos.filter(todo => !todo.completed));
+        break;
+      case FilterStatusType.Completed:
+        setVisibleTodos(todos.filter(todo => todo.completed));
+        break;
+      default:
+        setVisibleTodos(todos);
     }
   }, [filterStatus, todos]);
 
   useEffect(() => {
-    getTodos()
-      .then(todosFromServer => {
-        setTodos(todosFromServer);
-        setShowError(false);
-        setErrorMessage('');
-      })
-      .catch(() => {
-        setErrorMessage('Unable to load todos');
-        setShowError(true);
-      });
-  }, []);
+    if (needToRefresh) {
+      getTodos()
+        .then(todosFromServer => {
+          setTodos(todosFromServer);
+          setShowError(false);
+          setErrorMessage('');
+        })
+        .catch(() => {
+          setErrorMessage(ErrorMessage.LoadTodos);
+          setShowError(true);
+        });
+      setNeedToRefresh(false);
+    }
+  }, [needToRefresh]);
 
   if (!USER_ID) {
     return <UserWarning />;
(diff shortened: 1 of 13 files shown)


TetianaHodis1308 (approved):
You are well done for your efforts! 
I guess the professional opinion will be given by the mentor

src/App.tsx
```diff
@@ -1,26 +1,143 @@
...
+import { ErrorComponent } from './components/ErrorComponent/ErrorComponent';
+import { Footer } from './components/Footer/Footer';
+import { FilterStatusType } from './types/FilterStatusType';
+import cn from 'classnames';
+import { FormComponent } from './components/FormComponent/FormComponent';
+import { ErrorMessage } from './types/ErrorStatusType';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [showError, setShowError] = useState(false);
```
ArturErmolenko:
```suggestion
```
why we need this state?
let's handle error with just one state KISS

src/App.tsx
```diff
@@ -1,26 +1,143 @@
...
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterStatus, setFilterStatus] = useState(FilterStatusType.All);
+  const [isCompletedTodosExist, setIsCompletedTodosExist] = useState(true);
+  const [numberOfNotCompletedTodos, setNumberOfNotCompletedTodos] = useState(0);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [needToRefresh, setNeedToRefresh] = useState(true);
```
ArturErmolenko:
```suggestion
```
All this should be const, keep in mind that every state change make your App rerender

src/App.tsx
```diff
@@ -1,26 +1,143 @@
...
+  const [todoIdsToDelete, setTodoIdsToDelete] = useState<number[]>([]);
+  const [isFocusTitleInput, setIsFocusTitleInput] = useState(false);
+
+  useEffect(() => {
+    if (showError) {
+      setTimeout(() => {
+        setShowError(false);
+      }, 3000);
+    }
+  }, [showError]);
```
ArturErmolenko: thats should be in Error component

src/App.tsx
```diff
@@ -1,26 +1,143 @@
...
+    let counter = 0;
+
+    for (const todo of todos) {
+      if (!todo.completed) {
+        counter++;
+      }
+    }
+
+    setNumberOfNotCompletedTodos(counter);
+  }, [todos]);
```
ArturErmolenko: to many useEffects --> it's rerenders --> it's bad for us
use const instead of state and useEffects

src/App.tsx
```diff
@@ -1,26 +1,143 @@
...
+      case FilterStatusType.Active:
+        setVisibleTodos(todos.filter(todo => !todo.completed));
+        break;
+      case FilterStatusType.Completed:
+        setVisibleTodos(todos.filter(todo => todo.completed));
+        break;
+      default:
+        setVisibleTodos(todos);
+    }
+  }, [filterStatus, todos]);
```
ArturErmolenko: same here 
const visibleTodos = getPreparedTodod(todos,filter) 
Remember your prev work we already do things like that

src/App.tsx
```diff
@@ -1,26 +1,143 @@
...
+
+  useEffect(() => {
+    if (needToRefresh) {
```
ArturErmolenko: what does it mean?
you should work with state we don't want to make extra calls on server

ArturErmolenko: looks like kostyl ;)

src/components/FormComponent/FormComponent.tsx
```diff
@@ -0,0 +1,108 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (isFocusTitleInput) {
+      requestAnimationFrame(() => {
+        inputRef.current?.focus();
+      });
+      setIsFocusTitleInput(false);
+    }
+  }, [isFocusTitleInput, setIsFocusTitleInput]);
```
ArturErmolenko: can we have just one useEffect

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }, [isTempTodoLoading]);
+
+  useEffect(() => {
+    if (todoIdsToDelete && todoIdsToDelete.includes(todo.id)) {
+      setIsLoading(true);
+    } else {
+      setIsLoading(false);
+    }
+  }, [todoIdsToDelete, todo.id]);
```
ArturErmolenko: so many useEffects we should get rid of them
can we handle loading state in App and pass value here
Also temp todo never loading at all 
i cant understand what happened here

ArturErmolenko (changes requested):
Good job but we need to refactor it
Pay attention on extra useeffects and states - > they are kills your App performance
Try to find better way
use more const

src/App.tsx
```diff
@@ -1,26 +1,82 @@
...
-        <br />
-        <a href="https://github.com/mate-academy/react_todo-app-loading-todos#react-todo-app-load-todos">
-          React Todo App - Load Todos
-        </a>
-      </p>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <header className="todoapp__header">
```
BudnikOleksii, line 39: Create a header component

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,121 @@
...
+        </a>
+
+        <a
+          href="#/active"
+          className={cn('filter__link', {
+            selected: filterStatus === FilterStatusType.Active,
+          })}
+          data-cy="FilterLinkActive"
+          onClick={() => handleChangeFilterStatus(FilterStatusType.Active)}
+        >
```
BudnikOleksii: You can convert `FilterStatusType` into an array an map through it to render links
```
{Object.values(FilterStatusType).map(filterStatus => (
   // link
))}
```

src/components/FormComponent/FormComponent.tsx
```diff
@@ -0,0 +1,84 @@
...
+    setNewTodoTitle(event.target.value);
+  };
+
+  const handleSubmitForm = (event: React.FormEvent) => {
+    event.preventDefault();
+    if (!newTodoTitle.trim()) {
+      setErrorMessage(ErrorMessage.EmptyTitle);
+      requestAnimationFrame(() => {
+        inputRef.current?.focus();
+      });
```
BudnikOleksii, line 33: Why do you need `requestAnimationFrame`?

src/components/FormComponent/FormComponent.tsx
```diff
@@ -0,0 +1,84 @@
...
+    };
+
+    setTempTodo(temp);
+
+    if (inputRef.current) {
+      inputRef.current.disabled = true;
+    }
+
+    client
+      .post<Todo>('/todos', temp)
```
BudnikOleksii: - Create functions to do requests
- `/todos` is a magic value

src/components/FormComponent/FormComponent.tsx
```diff
@@ -0,0 +1,84 @@
...
+      .then(createdTodo => {
+        setTodos([...todos, createdTodo]);
```
BudnikOleksii: Use functional `setState` when you need to update a state depending on the previous one

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+}) => {
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    if (todoIdsToDelete && todoIdsToDelete.includes(todo.id)) {
+      setIsLoading(true);
+    } else {
+      setIsLoading(false);
+    }
+  }, [todoIdsToDelete, todo.id]);
```
BudnikOleksii, line 33:
```suggestion
  const isLoading = todoIdsToDelete && todoIdsToDelete.includes(todo.id);
```
You don't need a state, variable is enough

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const handleOnClickDelete = () => {
+    setIsLoading(true);
+    client
+      .delete(`/todos/${todo.id}`)
```
BudnikOleksii: The same

src/utils/getVisibleTodos.ts
```diff
@@ -0,0 +1,19 @@
+import { FilterStatusType } from '../types/FilterStatusType';
+import { Todo } from '../types/Todo';
+
+export const getVisibleTodos = (todoList: Todo[], filter: FilterStatusType) => {
+  let filteredTodoList = [...todoList];
```
BudnikOleksii: You don't need to create a copy, filter doesn't mutate an array

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

src/App.tsx
yelyzavetaz: Kyrylo said it`s better to create two separate components like FormComponent and ToggleAllComponent, which i did in part 3 of the task. Is it critical to merge them into one or can i keep them separate?

src/components/FormComponent/FormComponent.tsx
yelyzavetaz: Tests failed without `requestAnimationFrame`. Might be a crutch, but it`s the only thing that worked for me.

src/App.tsx
BudnikOleksii: Thats ok. You can create a `header` component as well, that renders `FormComponent`and `ToggleAllComponent`

src/components/TodoItem/TodoItem.tsx
BudnikOleksii: Not fixed, please, do these changes in the next task

BudnikOleksii (approved):
👍

src/components/TodoItem/TodoItem.tsx
yelyzavetaz: Sorry, forgot to mention - I changed the state to variable in the third part.

PR #1804, opened 2025-06-17
task

Ayume-beep:
[DEMO LINK](https://nixita-1.github.io/react_todo-app-add-and-delete/)


deadslowright (approved):
Keep it up!^^

src/App.tsx
```diff
@@ -1,26 +1,286 @@
...
+  const [shouldDeleteCompleted, setShouldDeleteCompleted] = useState(false);
+  const [todoBeingAdded, setTodoBeingAdded] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [clearInput, setClearInput] = useState(false);
+  const focusedTodoRef = useRef<HTMLInputElement>(null);
+  const defaultInputRef = useRef<HTMLInputElement>(null);
+  const [shouldToggleAllCompleted, setShouldToggleAllCompleted] =
+    useState(false);
+
+  function filterTodos(filter: string) {
```
BudnikOleksii: Use one type of functions, in most cases you have arrow functions

src/App.tsx
```diff
@@ -1,26 +1,286 @@
...
+
+  useEffect(() => {
+    if (isFirstRender.current) {
+      isFirstRender.current = false;
+
+      return;
+    }
+
+    toggleTodoCompletedStatus();
+  }, [shouldToggleAllCompleted, toggleTodoCompletedStatus]);
```
BudnikOleksii: Don't toggle in `useEffect` just do it when a user clicks on the corresponding button

src/App.tsx
```diff
@@ -1,26 +1,286 @@
...
+    getAndShowTodos();
+  }, [getAndShowTodos]);
+
+  useEffect(() => {
+    deleteCompletedTodos();
+  }, [shouldDeleteCompleted, deleteCompletedTodos]);
+
+  useEffect(() => {
+    focusedTodoRef.current?.focus();
+  }, [focusedTodoRef, selectedTodo]);
```
BudnikOleksii: The same, don't do these things in useEffect

src/components/Error.tsx
```diff
@@ -0,0 +1,55 @@
...
+      )}
+      {errorMessage === 'Unable to delete a todo' && (
+        <>
+          Unable to delete a todo
+          <br />
+        </>
+      )}
+      {errorMessage === 'Unable to update a todo' && (
+        <>
+          Unable to update a todo
```
BudnikOleksii: Just render the corresponding error message

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,166 @@
...
+  currentTodos,
+}) => {
+  const [inputValue, setInputValue] = useState(todo.title);
+  const [loader, setLoader] = useState(false);
+  const currentToggle = currentTodos.some(toDo => !toDo.completed);
+
+  const handleDelete = async (todoToDel: Todo) => {
+    setLoader(true);
+
+    await new Promise(resolve => setTimeout(resolve, 100));
```
BudnikOleksii, line 40: Why do you need this?

src/components/TodoList.tsx
```diff
@@ -0,0 +1,83 @@
...
+              shouldToggleAllCompleted={shouldToggleAllCompleted}
+              currentTodos={currentTodos}
+            />
+          ))}
+        </div>
+      )}
+
+      {tempTodo && (
+        <div
+          data-cy="Todo"
```
BudnikOleksii: You have TodoItem component

src/utils/fetchClient.ts
```diff
@@ -0,0 +1,41 @@
+const BASE_URL = 'https://mate.academy/students-api';
+
+function wait(delay: number) {
```
BudnikOleksii, line 3: You are add additional waiting here so you don't need to do it in your handlers

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments and questions, check them

src/App.tsx
```diff
@@ -1,26 +1,243 @@
...
+  const focusedTodoRef = useRef<HTMLInputElement>(null);
+  const didLoad = useRef(false);
+
+  const focusDefaultInput = () => {
+    if (defaultInputRef.current) {
+      defaultInputRef.current.focus();
+    }
+  };
+
+  const showError = (msg: string) => {
```
BudnikOleksii:
```suggestion
  const showError = (message: string) => {
```
Don't shorthand, name should be descriptive

src/components/TodoItem.tsx
BudnikOleksii: Not fixed

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,166 @@
...
+    setLoader(false);
+  };
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(e.target.value);
+  };
+
+  const handleBlurOrSubmit = () => {
+    if (inputValue !== todo.title) {
+      if (inputValue === '') {
```
BudnikOleksii:
```suggestion
      if (inputValue) {
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,166 @@
...
+          title: inputValue.trim(),
+          userId: USER_ID,
+          completed: todo.completed,
+        });
+      }
+    }
+
+    if (inputValue === todo.title) {
+      setSelectedTodo(undefined);
+    }
```
BudnikOleksii:
```suggestion
    if (inputValue === todo.title) {
      setSelectedTodo(null);
      
      return;
    }
    
    if (inputValue) {
    // other code
```
You can add early return to avoid nested `if` statements.
Also, it's better to use `null` instead of `undefined` in this case

src/components/TodoList.tsx
BudnikOleksii: Not fixed

BudnikOleksii (changes requested):
Almost done, a few comments still not fixed. And added a few more suggestions

src/components/TodoItem.tsx
Ayume-beep: Without it test wont pass while app works just fine,a little "kostyl" for tests

Nixita-1:
Additional wait is added because of tests. If I remove that delays tests will fail. And also moving tempTodo to todoItem was broken,it always were render with each todo,not only for new one,so I created seperate component for tempTodo. And done some minor adjustments like switching undefined for null and msg for message

BudnikOleksii (approved):

PR #1816, opened 2025-06-19
resolution app-and-delete

14lucas-mendes:
[DEMO LINK](https://14lucas-mendes.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,6 +28,8 @@ export const App: React.FC = () => {
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   // eslint-disable-next-line react-hooks/rules-of-hooks
   const inputRef = useRef<HTMLInputElement>(null);
+  // Estado para ids de todos sendo deletados
+  const [deletingIds, setDeletingIds] = useState<number[]>([]);
 
   // Busca os todos quando o componente carrega
   // eslint-disable-next-line react-hooks/rules-of-hooks
@@ -88,6 +90,7 @@ export const App: React.FC = () => {
 
   // Função para deletar todo
   const deleteTodo = (id: number) => {
+    setDeletingIds(prev => [...prev, id]);
     setLoading(`deletando-${id}`);
 
     client.delete(`/todos/${id}`)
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,26 +1,363 @@
...
+              >
+                <input
+                  id={`todo-checkbox-${todo.id}`}
+                  type="checkbox"
+                  className="todo__status"
+                  checked={todo.completed}
+                  onChange={() => handleToggleCompleted(todo.id)}
+                  disabled={loadingAction === `toggle-${todo.id}`}
+                />
+                <span className="visually-hidden">Toggle {todo.title}</span>
```
raulriato: Why is this span here?

raulriato: I think this is possibly the major problem in the project

raulriato (changes requested):
Your code almost works fine, but that span should not be there.

src/App.tsx
```diff
@@ -1,26 +1,361 @@
...
+                disabled={loadingAction === `delete-${todo.id}`}
+                aria-label={`Delete ${todo.title}`}
+              >
+                ×
+              </button>
+            </div>
+          ))}
+
+          {tempTodo && (
+            <div data-cy="Todo" className="todo">
```
pedro-ruas: I believe the problem is this tempTodo.
If tempTodo is just a loading before the todo is saved. Why are you rendering an input, a label and a span?
Does it need to receive user input? No -> no need for the input
No need for the input -> no need for label
And what is the point of saving the todo in the `tempTodo` stte, if you are rendering a fixed text "Loading todo"? Shouldn't it be the tempTodo itself being rendered?

pedro-ruas (changes requested):

DanielCamposDevX (changes requested):
Works fine but tests have failed on  finding the `[data-cy="ErrorNotification"]`

DanielCamposDevX (changes requested):
Tests are still failing in the same data-cy did you ran tests locally before submitting the PR?

pedro-ruas (changes requested):
Lucas, I see there are failing tests, with errors that seem related to missing `data-qa` properties. Be sure to check that.
Also, I've noticed that there is no loading when add a ToDo, and the checkbox when a todo is completed is also missing styling.
If I'm not mistaken, the styling from this task is already done, you only need to use the same classes as the mockup (the code that was in the task before you start).

pedro-ruas (approved):

DanielCamposDevX (approved):

PR #1840, opened 2025-07-11
Solution part two

Mariiaantoniv:
- [DEMO LINK](https://Mariiaantoniv.github.io/react_todo-app-add-and-delete/)


vitalii-bondarenko-ua (changes requested):
Good job.

- Check styles for temp todo and compare view with working example.

https://github.com/user-attachments/assets/6ac658f3-284e-4ba0-9bdd-e9da78891f17

- during removing todo should be visible a loader on todo.

vitalii-bondarenko-ua (approved):
Looks good.

PR #1842, opened 2025-07-12
add task solution

hannachachacha:
- [DEMO LINK](https://hannachachacha.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -119,7 +119,7 @@ export const App: React.FC = () => {
           onInputChange={onInputChange}
         />
 
-        {todos?.length !== 0 && (
+        {(todos.length > 0 || tempTodo) && (
           <>
             <TodoList
               todos={todos ?? []}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,148 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useState, useEffect, useRef } from 'react';
+import { USER_ID } from './api/todos';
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,26 +1,149 @@
...
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <Header
+          handleEmptyTitle={showError}
+          handleNewTitle={handleNewTitle}
+          isLoading={isLoading}
+          input={input}
+          onInputChange={onInputChange}
```
vitalii-bondarenko-ua: input, onInputChange, handleNewTitle can live in Header component.

src/App.tsx
```diff
@@ -1,26 +1,149 @@
...
+        setInput('');
+      } catch {
+        showError('Unable to add a todo');
+        setTempTodo(null);
+      } finally {
+        setIsLoading(false);
+      }
+    };
+
+    addTodo();
```
vitalii-bondarenko-ua:
```suggestion
      try {
        setIsLoading(true);
        const response = await client.post<Todo>(`/todos`, newTodo);

        setTodos(prevTodos => [...prevTodos, response]);
        setTempTodo(null);
        setInput('');
      } catch {
        showError('Unable to add a todo');
        setTempTodo(null);
      } finally {
        setIsLoading(false);
      }
```

looks like redundant create addTodo function and immediately call it in one place

vitalii-bondarenko-ua (changes requested):
Good job.
Consider some improvements.

src/App.tsx
hannachachacha: @vitalii-bondarenko-ua Hi! Thanks for the review! Moved input and onInputChange to Header component, also simplified addTodo part. But I left handleNewTitle in App, because it also controls other states (Todos, TempTodo, isLoading)

anastasiiavorobiova (approved):
🌞

PR #1861, opened 2025-07-25
Develop

Knysh19:


src/App.tsx
```diff
@@ -1,26 +1,293 @@
...
+                  href="#/completed"
+                  className={`filter__link ${filter === 'completed' ? 'selected' : ''}`}
+                  data-cy="FilterLinkCompleted"
+                  onClick={e => {
+                    e.preventDefault();
+                    setFilter('completed');
+                  }}
+                >
+                  Completed
+                </a>
```
vadiimvooo: Use enum and map to render the list

src/App.tsx
```diff
@@ -1,26 +1,293 @@
...
+        method: 'POST',
+        body: JSON.stringify({
+          userId: USER_ID,
+          title: trimmedTitle,
+          completed: false,
+        }),
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      });
```
vadiimvooo: Create helper functions which will get, delete, update, create todo. Move them to separate file

src/App.tsx
```diff
@@ -1,26 +1,293 @@
...
+import { USER_ID } from './api/todos';
+import { getTodos } from './api/todos';
+import { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Todo';
+import { ErrorNotification } from './components/error/ErrorNotification';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  // const [isLoading, setIsLoading] = useState(false);
```
vadiimvooo: remove comments

vadiimvooo (changes requested):
Good job, but please separate your code on some smaller parts: TodoItem, TodoList, Footer, Header etc. Feel free to ask for help in fe_chat.

TarasHoliuk (approved):
Approved with one comment:

If a component doesn't have some additional files (related only to it child components, style files, etc.) and includes only one .tsx file there is no need to create a folder for it - just file inside of the components folder. Also, files with components should always be in PascalCase, i.e. Header.tsx

PR #1867, opened 2025-07-31
Solution

Whslv:
[DEMO LINK](https://Whslv.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,191 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { deleteTodo, getTodos, postTodo, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [input, setInput] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [tempDeleteTodo, setTempDeleteTodo] = useState<Todo | null>(null);
+  const [tempClearAll, setTempClearAll] = useState<Todo[] | null>(null);
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const onError = (errorString: string) => {
+    setErrorMessage(errorString);
+    setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
+  };
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filter) {
+      case 'active':
+        return !todo.completed;
+      case 'completed':
+        return todo.completed;
+      case 'all':
+      default:
+        return todo;
+    }
+  });
+
+  const handlClearAll = async () => {
+    try {
+      setTempClearAll(completedTodos);
+      const deletionList = await Promise.allSettled(
+        completedTodos.map(todo => deleteTodo(todo.id)),
+      );
+
+      const hasErorr = deletionList.some(item => item.status === 'rejected');
+      const seccessIds = completedTodos
+        .map((todo, i) =>
+          deletionList[i].status === 'fulfilled' ? todo.id : null,
+        )
+        .filter((id): id is number => id !== null);
+
+      setTodos(prev => prev.filter(todo => !seccessIds.includes(todo.id)));
+      setTempClearAll(null);
+
+      if (hasErorr) {
+        onError('Unable to delete a todo');
+      }
+    } catch {
+      onError('Unable to delete a todo');
+    }
+  };
+
+  const handleDelete = async (todosId: number) => {
+    try {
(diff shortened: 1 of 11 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,60 @@
...
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      <TodoItem
+        filteredTodos={filteredTodos}
+        tempClearAll={tempClearAll}
+        tempDeleteTodo={tempDeleteTodo}
+        handleDelete={handleDelete}
+      />
+      {tempTodo !== null && (
```
TarasHoliuk: nit:
```suggestion
      {tempTodo && (
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+              <div
+                data-cy="TodoLoader"
+                className={`modal overlay ${tempDeleteTodo?.id === todo.id || tempClearAll?.some(t => t.id === todo.id) ? 'is-active' : ''}`}
+              >
+                <div className="modal-background has-background-white-ter" />
+                <div className="loader" />
+              </div>
+            </div>
+
+            {/* This todo is being edited */}
```
TarasHoliuk: delete comments

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+                onClick={() => {
+                  handleDelete(todo.id);
+                }}
+              >
+                ×
+              </button>
+
+              <div
+                data-cy="TodoLoader"
+                className={`modal overlay ${tempDeleteTodo?.id === todo.id || tempClearAll?.some(t => t.id === todo.id) ? 'is-active' : ''}`}
```
TarasHoliuk: You have a few separate variables to define if todo is in loading state. Instead, create one variable for it with an  array of ids (similar to tempClearAll but rename to something more descriptive, like `loadingTodoIds`, no need to store entire todos, ids is enough)
Use this array for all actions (updating todo, deleting todo)

Also, it would be nice to reuse the TodoItem component for tempTodo displaying. So it should accept just `isLoading` prop (will be always true for tempTodo)

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+export const TodoItem: React.FC<TodoItemProps> = ({
+  filteredTodos,
+  tempClearAll,
+  tempDeleteTodo,
+  handleDelete,
+}) => {
+  return (
+    <>
+      {filteredTodos.map(todo => {
```
TarasHoliuk: You shouldn't map in this component. Do it in `TodoList`. TodoItem component should be for one item

src/components/TodoList.tsx
```diff
@@ -0,0 +1,60 @@
...
+  tempClearAll,
+  tempTodo,
+  tempDeleteTodo,
+  handleDelete,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      <TodoItem
+        filteredTodos={filteredTodos}
+        tempClearAll={tempClearAll}
```
TarasHoliuk: map todos in this component and pass `isLoading` prop to each separate TodoItem, like `isLoading={loadingTodoIds.some(id => id === todo.id)}`

src/components/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+export const Header: React.FC<HeaderProps> = ({
+  todos,
+  inputRef,
+  tempTodo,
+  input,
+  setInput,
+  handleTodoSubmit,
+}) => {
+  return (
+    <>
```
TarasHoliuk: redundant fragment

src/components/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <header className="todoapp__header">
+        <button
+          type="button"
+          className={`todoapp__toggle-all ${todos.every(todo => todo.completed && 'active')}`}
```
TarasHoliuk: it's better for readability to define `areAllTodosCompleted` above and use here

src/components/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+            data-cy="NewTodoField"
+            ref={inputRef}
+            type="text"
+            disabled={tempTodo !== null}
+            value={input}
+            className="todoapp__new-todo"
+            placeholder="What needs to be done?"
+            onKeyDown={e => {
+              handleTodoSubmit(e);
+              if (
```
TarasHoliuk: I think that checking `if(inputRef?.current)` in enough

Also, you can do it inside of the `handleTodoSubmit` function - ref variable is available in App component

src/components/Footer.tsx
```diff
@@ -0,0 +1,82 @@
...
+  todos,
+  activeTodos,
+  completedTodos,
+  filter,
+  setFilter,
+  handlClearAll,
+}) => {
+  return (
+    <>
+      {todos.length > 0 && (
```
TarasHoliuk: Do this conditional render for the entire Footer component in the App.tsx instead of passing todos here

src/components/Footer.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { Todo } from '../types/Todo';
+
+interface FooterProps {
+  todos: Todo[];
+  activeTodos: Todo[];
+  completedTodos: Todo[];
+  filter: 'all' | 'active' | 'completed';
```
TarasHoliuk: Create union type for filter

src/App.tsx
```diff
@@ -1,26 +1,191 @@
...
+
+          if (newTodo) {
+            setTempTodo(null);
+            setTodos([...todos, newTodo]);
+            setInput('');
+            setErrorMessage('');
+          }
+        }
+
+        if (cleanInput === '') {
```
TarasHoliuk:
```suggestion
        if (!cleanInput) {
```
nit: empty string is a falsy value

TarasHoliuk (changes requested):
GJ! Let's improve your code a bit

maxim2310 (approved):
Well done!

PR #1869, opened 2025-07-31
solution

LinaDo1609:


src/components/Todo/todo.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { Todo } from '../../types/Todo';
+import { Loader } from '../loader/loader';
+import { useState } from 'react';
+
+type Props = {
+  todo: Todo;
+  deleteTodo: (postId: number) => Promise<void>;
+};
+
+export const TodoItem: React.FC<Props> = ({ todo, deleteTodo }) => {
```
TarasHoliuk: - All files with components should be in PascalCase
- No need to create a separate folder with only one .tsx file with a component. In this case you can just create file right inside of components folder
- File name should match the Component name

So, the current file name should be TodoItem.tsx, and should be placed right in components folder. Fix everywhere

src/App.tsx
```diff
@@ -1,26 +1,144 @@
...
+import { Header } from './components/header/header';
+import { TodoList } from './components/todoList/todoList';
+import { addTodo, deleteTodo, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const FILTERS = {
```
TarasHoliuk: Create enum instead of const, define it in a file with types, export it and use everywhere you deal with filter

src/App.tsx
```diff
@@ -1,26 +1,144 @@
...
+    all: 'all',
+    completed: 'completed',
+    active: 'active',
+  };
+
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [filteredList, setFilteredList] = useState<Todo[]>([]);
+  const [completedTodos, setCompletedTodos] = useState<Todo[]>([]);
```
TarasHoliuk: No need to create separate states if you can define their values based on other states (or props). Reason - once basic states/props change component will be rerendered and values will be recomputed.

So:

```
  return todos.filter(// filtering logic based on all todos and filter)
}, [todos, filterQuery]);
```

src/App.tsx
```diff
@@ -1,26 +1,144 @@
...
+  const [error, setError] = useState<string | null>(null);
+  const [unCompletedCount, setUnCompletedCount] = useState<number>(0);
```
TarasHoliuk: same - just create const instead of separate state

src/App.tsx
```diff
@@ -1,26 +1,144 @@
...
+  const [loading, setLoading] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [filterQuery, setFilterQuery] = useState('all');
```
TarasHoliuk: just filter, or filterStatus - better name

src/App.tsx
```diff
@@ -1,26 +1,144 @@
...
+  useEffect(() => {
+    const uncompleted = todoList.filter(todo => todo.completed === false);
+
+    setCompletedTodos(todoList.filter(todo => todo.completed === true)); //фільтруємо виконані todos
+    setUnCompletedCount(uncompleted.length); // лічільник активних todos
+    
+    handleFilter(filterQuery); // застосовуємо фільтр до списку todos
+
+
+  }, [todoList]);
```
TarasHoliuk: Знімок екрана 2025-07-31 о 20 54 54

Just add the correct dependencies in such cases

src/App.tsx
```diff
@@ -1,26 +1,144 @@
...
+      .then(newPost => {
+        setTodoList(prevList => [...prevList, newPost]);
+        setLoading(false);
+        setTempTodo(null);
+      })
+      .catch(() => {
+        setError('Unable to add a todo');
+        setTempTodo(null);
+        setLoading(false);
+        throw new Error('Unable to add a todo');
```
TarasHoliuk: Don't throw again - that will cause page crash if you don't catch it again somewhere

src/components/Todo/todo.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <input
+          data-cy="TodoStatus"
+          id={`todo-${todo.id}`}
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+        />
+      </label>
+
+      {/* <form>
```
TarasHoliuk: delete comments

src/components/todoList/todoList.tsx
```diff
@@ -0,0 +1,49 @@
...
+  deleteTodo,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todoList.map(todo => (
+        <TodoItem todo={todo} key={todo.id} deleteTodo={deleteTodo} />
+      ))}
+
+      {/* todoTemp з'являється, коли йде загрузка запиту на додавання todo до серверу */}
+      {todoTemp && (
```
TarasHoliuk: Try to reuse TodoItem component for tempTodo (at least you need to pass isLoading as a prop)

TarasHoliuk (changes requested):
You did a good job! Added a few comments to improve your code

src/App.tsx
LinaDo1609: i have to add throw an error, to not reset title in case if there is any error

because, catch return successful promise

vitalii-bondarenko-ua (changes requested):
Good job.
You need to pass all tests.
Now problem with imports. Check register for imports and file name.

Screenshot 2025-08-02 at 10 07 47
Screenshot 2025-08-02 at 10 09 14

Make sure all tests passed before next re-request.

LinaDo1609:
it's so strange, because in VS i have passed all test and linter test. but here it shows some fails
linter
image

LinaDo1609:
i have found the problem.
i updated names of tsx files, but Git didn't update it automatically, so i did it manually

vadiimvooo (approved):
Well done

PR #1884, opened 2025-08-14
React Todo App Add and Delete

vanyagg:
- [DEMO LINK](https://vanyagg.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,7 +10,6 @@ import { ErrorNotification } from './components/ErrorNotification';
 import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
-  const [allTodos, setAllTodos] = useState<Todo[]>([]);
   const [todos, setTodos] = useState<Todo[]>([]);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [todoTitle, setTodoTitle] = useState('');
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,201 @@
...
+          );
+          setIsDeletingId(todo.id);
+        })
+        .catch(() => setErrorMessage('Unable to delete a todo'))
+        .finally(() => {
+          setIsLoading(false);
+          setIsDeletingId(-1);
+        });
+    });
+  };
```
etojeDenys: it's better to use Promise.all here

src/App.tsx
```diff
@@ -1,26 +1,201 @@
...
+import { addTodos, deleteTodos, getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
+import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
+  const [allTodos, setAllTodos] = useState<Todo[]>([]);
+  const [todos, setTodos] = useState<Todo[]>([]);
```
etojeDenys: you don't need two of the same useState. if you're using it for filtering, you can just create a new variable for that

etojeDenys (changes requested):

vadiimvooo (approved):
Well done

PR #1885, opened 2025-08-14
Add task

iteamfox:
[DEMO LINK](https://iteamfox.github.io/react_todo-app-add-and-delete/)


danon321 (changes requested):
Brak linku do demo

Zibi95 (changes requested):
I’ve already suggested changes in another PR that contains the exact same code. Once you apply the fixes there and that PR gets accepted, you should copy those changes here. This one should only be opened after the first one is merged, otherwise it doesn’t really make sense.

danon321 (approved):
Wyglada bardzo dobrze :))

PR #1889, opened 2025-08-18
Solution

JPodgrodzki:


Zibi95 (changes requested):
Hi, could you please share the demo link so I can check if everything works?
Also, did you finish the previous part of this task, and was it accepted?

JPodgrodzki:
@Zibi95:
[DEMO LINK](https://JPodgrodzki.github.io/react_todo-app-add-and-delete/)
Previous part is finished and accepted

src/App.tsx
```diff
@@ -1,26 +1,313 @@
...
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
-
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<Errors | null>(null);
+  const [filter, setFilter] = useState<Filter>(Filter.All);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
```
zibi-dev: This state change causes the whole app to re-render, and it happens on every keystroke. Do you really need to update that state so aggressively on each keystroke (input can be uncontrolled and you would take value from input in submit handler)? Or would it make more sense to move this state somewhere else?

src/App.tsx
```diff
@@ -1,26 +1,313 @@
...
+    setTempTodo(newTempTodo);
+
+    try {
+      const createdTodo = await addTodo(trimmedTitle);
+
+      setTodos(current => [...current, createdTodo]);
+    } catch {
+      setError(Errors.Add);
+
+      return;
```
zibi-dev: This return is not needed

src/App.tsx
```diff
@@ -1,26 +1,313 @@
...
+        setDeleteTodoIds(current =>
+          current.filter(singleId => singleId !== id),
+        );
+      }
+    }
+
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  };
```
zibi-dev: This implementation deletes todos sequentially and updates state inside the loop, which causes unnecessary re-renders and slows things down. It also risks inconsistent state if some deletions fail. Consider batching state updates and running deletions in parallel (e.g. with Promise.allSettled) to improve performance

Zibi95 (changes requested):
Good job! Found some things that are concerning and should be thinked about when creating apps with react. Also it would be best if you would split JSX into separate components.

src/components/header/Header.tsx
```diff
@@ -0,0 +1,44 @@
...
+  newTodoTitle: string;
+  tempTodo: Todo | null;
+  inputRef: React.RefObject<HTMLInputElement>;
+};
+
+export const Header: React.FC<Props> = ({
+  handleSubmit,
+  setNewTodoTitle,
+  newTodoTitle,
+  tempTodo,
```
zibi-dev: It’s considered good practice to pass props as primitives, because they are easier to optimize — unlike objects or functions, they are not passed by reference. So just simple `disabled` prop passed to this component would be better

src/components/todoItem/TodoItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type Props = {
+  todo: Todo;
+  deleteTodoIds: number[];
+  handleDelete: (arg0: number) => {};
+};
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  deleteTodoIds,
```
zibi-dev: Here the same simple `isActive`

src/App.tsx
zibi-dev: Have you tought about this?

src/components/todoItem/TodoItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+        className="todo__remove"
+        data-cy="TodoDelete"
+        onClick={() => handleDelete(todo.id)}
+      >
+        ×
+      </button>
+
+      <div
+        data-cy="TodoLoader"
+        className={`modal overlay ${deleteTodoIds.includes(todo.id) && 'is-active'}`}
```
zibi-dev: There is cool lib installed in the projects called `classnames`, take a look into that

src/App.tsx
```diff
@@ -1,26 +1,273 @@
...
+
+    setDeleteTodoIds(current => [...current, ...completedIds]);
+
+    const results = await Promise.allSettled(
+      completedIds.map(id => deleteTodo(id)),
+    );
+
+    const successfullIds = completedIds.filter(
+      (_, i) => results[i].status === 'fulfilled',
+    );
```
zibi-dev: Its not bad for performance, bcs this list won't be that long, but you can to this in once go with `reduce`

```
  const { successfulIds, rejectedIds } = results.reduce(
    (acc, result, i) => {
      if (result.status === "fulfilled") {
        acc.successfulIds.push(completedIds[i]);
      } else {
        acc.rejectedIds.push(completedIds[i]);
      }
      return acc;
    },
    { successfulIds: [], rejectedIds: [] }
  );
```

Zibi95 (changes requested):

src/components/todoItem/TodoItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+        className="todo__remove"
+        data-cy="TodoDelete"
+        onClick={() => handleDelete(todo.id)}
+      >
+        ×
+      </button>
+
+      <div
+        data-cy="TodoLoader"
+        className={classNames('modal', 'overlay', isActive && 'is-active')}
```
zibi-dev:
```suggestion
        className={classNames('modal overlay', {'is-active': isActive})}
```

src/App.tsx
```diff
@@ -1,26 +1,273 @@
...
+        data-cy="ErrorNotification"
+        className={[
+          'notification',
+          'is-danger',
+          'is-light',
+          'has-text-weight-normal',
+          !error && 'hidden',
+        ]
+          .filter(Boolean)
+          .join(' ')}
```
zibi-dev: ??

zibi-dev:
```suggestion
  className={classNames([
    'notification is-danger is-light has-text-weight-normal',
    { hidden: !error },
  ])}
```

src/App.tsx
zibi-dev: ??

src/App.tsx
zibi-dev: ??

Zibi95 (changes requested):

src/components/header/Header.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <form
+        onSubmit={async event => {
+          event.preventDefault();
+
+          const success = await handleSubmit(query);
+
+          if (success) {
+            setQuery('');
+          }
+        }}
```
zibi-dev, line 35: move it into handler function

Zibi95 (approved):
Good job. One comment left but it can be changed in next step.

PR #1890, opened 2025-08-19
Solution

kamilbykow:
[DEMO LINK](https://kamilbykow.github.io/react_todo-app-add-and-delete/)


src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,53 @@
...
+}) => {
+  const [hide, setHide] = useState(true);
+
+  useEffect(() => {
+    if (hidden) {
+      setHide(false);
+      setTimeout(() => {
+        setHide(true);
+        setError(false);
+      }, 3000);
```
zibi-dev: It’s a good idea to extract such numbers into named constants (e.g. ERROR_CLEANUP_TIMEOUT). This makes the code more readable, self-documenting, and easier to maintain if the value needs to change later.

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,53 @@
...
+      {/* Unable to load todos */}
+      {errorMsg}
+      {/* <br />
+      Title should not be empty
+      <br />
+      Unable to add a todo
+      <br />
+      Unable to delete a todo
+      <br />
+      Unable to update a todo */}
```
zibi-dev: comments to remove

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,39 @@
...
+  setFilter: (name: Filters) => void;
+  activeFilter: Filters;
+};
+
+export const Filter: React.FC<Props> = ({ setFilter, activeFilter }) => {
+  return (
+    <nav className="filter" data-cy="Filter">
+      <a
+        href="#/"
+        className={`filter__link ${activeFilter === 'All' ? 'selected' : ''}`}
```
zibi-dev: use `classnames` library that you have installed in the project already.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    // {/* Hide the footer if there are no todos */}
+    <footer className={`todoapp__footer`} data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {count} items left
+      </span>
+
+      {/* Active link should have the 'selected' class */}
+
+      <Filter setFilter={setFilter} activeFilter={activeFilter} />
```
zibi-dev: looks like prp drilling. Do you know how to resolve such thing?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,45 @@
+import { Filter } from '../Filter/Filter';
+import { Filters } from '../../App';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  count: number;
+  setFilter: (name: Filters) => void;
+  activeFilter: Filters;
+  Completed: Todo[];
```
zibi-dev: prop starting with bigger case

zibi-dev: Its confusing even on github it is colored as components

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,129 @@
...
+            type="checkbox"
+            className="todo__status"
+            checked
+          />
+        </label>
+
+        <span data-cy="TodoTitle" className="todo__title">
+          Completed Todo
+        </span> */}
+
```
zibi-dev: remove comments

src/components/Main/Main.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { Todo as Todos } from '../../types/Todo';
+import { Todo } from '../Todo/Todo';
+
+type Props = {
+  todos: Todos[] | undefined | null;
+  handleDelete: (id: number[]) => void;
+  todosToDelete: number[];
+};
+
+export const Main: React.FC<Props> = ({
```
zibi-dev: This name for component is confusing. From the code for me it look like `TodosList`

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,56 @@
...
+  todo: TypeTodo;
+  handleDelete: (id: number[]) => void;
+  toDelete: boolean;
+};
+
+export const Todo: React.FC<Props> = ({ todo, handleDelete, toDelete }) => {
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    if (todo.id === 0) {
```
zibi-dev: it would be good to store this id as constant variable TEMP_TODO_ID if it is used in few places. This way it will be easier to make changes

zibi-dev: also what would be good. Bcs it might be confusing what is checked here. So soetimes its nice to create function for example here it could be `checkTempTodo`

src/App.tsx
```diff
@@ -1,26 +1,171 @@
...
 
-const USER_ID = 0;
+export type Filters = 'Completed' | 'Active' | 'All';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState<string>('');
+  const [disableSubmit, setDisableSubmit] = useState(false);
+  const [todos, setTodos] = useState<Todo[] | undefined>();
+  const [error, setError] = useState<boolean>(false);
+  const [errorMsg, setErrorMsg] = useState('');
```
zibi-dev: Hmm maybe just one state is enough. It can become cumbersome to manage this many states that do the same thing

src/App.tsx
```diff
@@ -1,26 +1,171 @@
...
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
 
-const USER_ID = 0;
+export type Filters = 'Completed' | 'Active' | 'All';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState<string>('');
+  const [disableSubmit, setDisableSubmit] = useState(false);
+  const [todos, setTodos] = useState<Todo[] | undefined>();
```
zibi-dev: You can start todos state with empty array instead of initializing it with `undefined`

src/App.tsx
```diff
@@ -1,26 +1,171 @@
...
+  const [error, setError] = useState<boolean>(false);
+  const [errorMsg, setErrorMsg] = useState('');
+  const [filter, setFilter] = useState<Filters>('All');
```
zibi-dev: create obj or enum for filters and use it instead of simple strings.

src/App.tsx
```diff
@@ -1,26 +1,171 @@
...
+import { Todo } from './types/Todo';
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
 
-const USER_ID = 0;
+export type Filters = 'Completed' | 'Active' | 'All';
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState<string>('');
+  const [disableSubmit, setDisableSubmit] = useState(false);
```
zibi-dev: This state probably can be renamed to isLoading. Which will be true on each action in app

zibi-dev: except writing in input

src/App.tsx
```diff
@@ -1,26 +1,171 @@
...
+        setErrorMsg('Unable to add a todo');
+      })
+      .finally(() => {
+        setDisableSubmit(false);
+      });
+  };
+
+  const handleDelete = async (id: number[]) => {
+    setTodosToDelete(id);
+    for (let i = 0; i < id.length; i++) {
```
zibi-dev: This implementation deletes todos sequentially and updates state inside the loop, which causes unnecessary re-renders and slows things down. It also risks inconsistent state if some deletions fail. Consider batching state updates and running deletions in parallel (e.g. with Promise.allSettled) to improve performance and reliability.

Zibi95 (changes requested):
Hey. Good job. Found some issues that should be fixed before this task can be accepted

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,44 @@
...
+  setError: (err: string) => {};
+  error: string;
+};
+
+export const ErrorNotification: React.FC<Props> = ({
+  hidden,
+  setError,
+  error,
+}) => {
+  const [hide, setHide] = useState(true);
```
zibi-dev: Are you sure you need this state. It is `true` when hidden is `true`.

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,43 @@
+import classNames from 'classnames';
+import { Filters, useFilterContext } from '../../App';
+
+export const Filter: React.FC = () => {
+  const { filter, changeFilter } = useFilterContext();
```
zibi-dev, line 6: I see that you went into context to get rid of prop drilling. That's okay. You could also use the `children` prop:

```jsx
function Card({ children }) {
  return <div className="card">{children}</div>;
}

function App() {
  return (
    <Card>
      <h2>Hello</h2>
      <p>This is inside the card</p>
    </Card>
  );
}

File: src/components/Header/Header.tsx [hunk 16]
```diff
@@ -0,0 +1,55 @@
...
+  setQuery,
+  handleSubmit,
+  disableInput,
+  inputRef,
+}) => {
+  useEffect(() => {
+    if (!disableInput) {
+      inputRef.current?.focus();
+    }
+  }, [disableInput, inputRef]);
```
zibi-dev: refs dont have to be added into deps array  https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+        case Filters.Active:
+          return filteredTodos?.filter(todo => !todo.completed);
+
+        case Filters.All:
+          return todos;
+      }
+    }
+
+    return undefined;
+  }, [todos, filteredTodos, filter]);
```
zibi-dev: You’re using filteredTodos inside its own useMemo. This creates a circular dependency and can cause stale or incorrect results. Use todos directly instead of filteredTodos

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+
+  return context;
+};
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState<string>('');
+  const [disableWritingInput, setDisableWritingInput] = useState(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo[] | null>();
```
zibi-dev: You didn’t provide an initial value, so tempTodo is `undefined | Todo[] | null`. Provide initial value of null

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,44 @@
+import { useEffect, useState } from 'react';
+
+type Props = {
+  hidden: string;
```
zibi-dev: this prop name sounds like it is boolean, but instead its string. It should be boolean. You can just pass Boolean `Boolean(error)` or `!!error`

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
-
-      <p className="subtitle">Styles are already copied</p>
-    </section>
+    <div className="todoapp">
+      <h1 className="todoapp__title">todos</h1>
+
+      <div className="todoapp__content">
+        <Header
+          query={query}
+          setQuery={item => setQuery(item)}
```
zibi-dev:
```suggestion
          setQuery={setQuery}
```

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+    }
+  };
+
+  const changeFilter = (name: Filters) => {
+    setFilter(name);
+  };
+
+  const value = {
+    filter: filter,
+    changeFilter: changeFilter,
```
zibi-dev:
```suggestion
    filter,
    changeFilter,
```

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+  const [disableWritingInput, setDisableWritingInput] = useState(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string>('');
+  const [tempTodo, setTempTodo] = useState<Todo[] | null>();
+  const [todosToDelete, setTodosToDelete] = useState<number[]>([]);
+  const [filter, setFilter] = useState<Filters>(Filters.All);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  let filteredTodos: Todo[] | undefined = todos;
```
zibi-dev: You dont reasign it anywhere. So just use const in line 137

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }, [todo.id]);
+
+  return (
+    <div data-cy="Todo" className={`todo ${todo.completed && 'completed'} `}>
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          defaultChecked={todo.completed}
```
zibi-dev: Using defaultChecked makes this an uncontrolled input, so if todo.completed changes later, the checkbox won’t update

src/components/Todo/Todo.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const TEMP_TODO_ID = 0;
+
+    return id === TEMP_TODO_ID;
+  };
+
+  useEffect(() => {
+    if (checkTempTodo(todo.id)) {
+      setLoading(true);
+    }
+  }, [todo.id]);
```
zibi-dev: I think that this might work alright without state. If todo is changed then this component will rerender.


```suggestion
 const checkTempTodo = (id: number) => {
    const TEMP_TODO_ID = 0;

    return id === TEMP_TODO_ID;
  };

  const isLoading = checkTempTodo(todo.id)
```

src/App.tsx
```diff
@@ -1,26 +1,214 @@
...
+
+export const useFilterContext = () => {
+  const context = useContext(Context);
+
+  if (context === undefined) {
+    throw new Error('undefined Context');
+  }
+
+  return context;
+};
```
zibi-dev: This context should be in separate component

Zibi95 (changes requested):

src/components/Context/Context.tsx
```diff
@@ -0,0 +1,19 @@
+import { createContext, useContext } from 'react';
+import { Filters } from '../../App';
+
+interface ContextType {
+  filter: Filters;
+  changeFilter: (name: Filters) => void;
+}
+
+export const Context = createContext<ContextType | undefined>(undefined);
```
zibi-dev, line 9: Context with what?. It should have some less generic name.

zibi-dev: example how to properly do Context in react 

```
// AuthContext.tsx
import { createContext, useContext, useState, ReactNode } from 'react';

type User = { id: string; name: string };
type AuthContextType = {
  user: User | null;
  login: (name: string) => void;
  logout: () => void;
};

const AuthContext = createContext<AuthContextType | null>(null);

export const AuthProvider = ({ children }: { children: ReactNode }) => {
  const [user, setUser] = useState<User | null>(null);

  const login = (name: string) =>
    setUser({ id: crypto.randomUUID(), name });
  const logout = () => setUser(null);

  return (
    <AuthContext.Provider value={{ user, login, logout }}>
      {children}
    </AuthContext.Provider>
  );
};

export const useAuth = () => {
  const ctx = useContext(AuthContext);
  if (!ctx) throw new Error('useAuth must be used inside AuthProvider');
  return ctx;
};
```

Usage 
```
// App.tsx
import { AuthProvider, useAuth } from './AuthContext';

function Dashboard() {
  const { user, login, logout } = useAuth();
  return user ? (
    <>
      <p>Hello {user.name}</p>
      <button onClick={logout}>Logout</button>
    </>
  ) : (
    <button onClick={() => login('Alice')}>Login</button>
  );
}

export default function App() {
  return (
    <AuthProvider>
      <Dashboard />
    </AuthProvider>
  );
}
```

zibi-dev: all the logic, state related to filters should be inside of here.

src/App.tsx
```diff
@@ -1,26 +1,189 @@
...
+        <TodosList
+          todos={filteredTodos}
+          handleDelete={handleDelete}
+          todosToDelete={todosToDelete}
+        />
+        <TodosList
+          todos={tempTodo}
+          handleDelete={handleDelete}
+          todosToDelete={todosToDelete}
+        />
```
zibi-dev, line 166: Ok this one is weird. Why you don't simply render TodoItem here

src/App.tsx
```diff
@@ -1,26 +1,189 @@
...
+
+        case Filters.Active:
+          return todos?.filter(todo => !todo.completed);
+
+        case Filters.All:
+          return todos;
+      }
+    }
+
+    return undefined;
```
zibi-dev, line 126: undefined is returned by default

src/App.tsx
```diff
@@ -1,26 +1,189 @@
...
+      }
+    });
+
+    if (failedIds.length > 0) {
+      setError('Unable to delete a todo');
+    }
+
+    if (successfulIds.length > 0) {
+      setTodos(prev => {
+        if (prev === undefined) {
```
zibi-dev, line 92: is there possibility that todos are undefined right now?

src/App.tsx
```diff
@@ -1,26 +1,189 @@
...
+      setError('Title should not be empty');
+      setTempTodo(null);
+      setDisableWritingInput(false);
+    } else {
+      postTodo(todo)
+        .then(res => {
+          setQuery('');
+          setTempTodo(null);
+          setTodos(prev => {
+            if (prev === undefined) {
```
zibi-dev, line 52: todos are initialized as []

src/App.tsx
```diff
@@ -1,26 +1,189 @@
...
-const USER_ID = 0;
+export enum Filters {
+  Completed,
+  Active,
+  All,
+}
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState<string>('');
+  const [disableWritingInput, setDisableWritingInput] = useState(false);
```
zibi-dev, line 22: I would say that this tate represents if app is processing something so name `isLoading` would fit better

Zibi95 (changes requested):

PR #1910, opened 2025-08-28
Develop

jeolinna:
[DEMO LINK](https://jeolinna.github.io/react_todo-app-add-and-delete/)


O1ympus (approved):

src/App.tsx
```diff
@@ -1,26 +1,225 @@
...
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+  const [isAdded, setIsAdded] = useState(false);
+  const [temporaryTodo, setTemporaryTodo] = useState<Todo | null>(null);
+  const [deleteId, setDeleteId] = useState<number[]>([]);
+
+  const newTodoFieldRef = useRef<HTMLInputElement>(null);
+
+  useLayoutEffect(() => {
+    newTodoFieldRef.current?.focus();
+  }, []);
```
alexander-ignatow, line 34: Why exactly you use layout effect here?

alexander-ignatow (approved):
looks good - lets proceed to a next stage

src/App.tsx
jeolinna: There was a problem with focus in the tests. At that moment, it was the only option that passed all the tests.

PR #1935, opened 2025-09-07
solution

isobol-js:
[DEMOLINK](https://isobol-js.github.io/react_todo-app-add-and-delete/)


bebelekele (approved):

alexander-ignatow (approved):
App component becomes too big

You need to separate it into smaller manageable components in the next version

Lets move forward, but next time such code won't pass

PR #1937, opened 2025-09-08
add task solution

Anna-0805:


src/App.tsx
```diff
@@ -1,26 +1,132 @@
...
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<null | string>(null);
+  const [filter, setFilter] = useState<SortType | string>('All');
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    const todoses = async () => {
```
FaiHamid: The function name should be the name of the action, such as loadingTodos, fetchTodos, or getTodos. This makes it clearer what the function does.

src/App.tsx
```diff
@@ -1,26 +1,132 @@
...
+      try {
+        setLoading(true);
+        const data = await getTodos();
+
+        setTodos(data);
+      } catch (err) {
+        setError('Unable to load todos');
```
FaiHamid: The error text can also be an enum.

src/App.tsx
```diff
@@ -1,26 +1,132 @@
...
 
+  let visibleToods = [...todos];
+
+  if (filter === SortType.Active) {
+    visibleToods = visibleToods.filter(good => !good.completed);
+  }
+
+  if (filter === SortType.Completed) {
+    visibleToods = visibleToods.filter(good => good.completed);
+  }
```
FaiHamid: This logic can be moved to a separate function.
`const getVisibleTodos = (sort, todos) => { .... }`
And in App.tsx, rewrite the visibleTodos variable like this:
`let visibleTodos = getVisibleTodos(filter, todos)`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+        data-cy="TodoDelete"
+      >
+        ×
+      </button>
+      {todo.id === 0 && (
+        <div data-cy="TodoLoader" className="modal overlay">
+          <div className="modal-background has-background-white-ter" />
+          <div className="loader" />
+        </div>
+      )}
```
FaiHamid: There should be no conditional rendering here, but only the ‘is-active’ class should be added conditionally.

FaiHamid (changes requested):
God job))
Let`s just fix some details:

1. Please, add DEMO LINK to the description of the request pool!
2. You also have only 12 tests to take, so you need to figure them out first.
image

Anna-0805:
https://Anna-0805.github.io/react_todo-app-add-and-delete

etojeDenys (approved):
good job

to improve:
after pressing the clear button on the to-do, the loader appears on every item, even those that will not be deleted

PR #1976, opened 2025-10-09
add solution

Aiiyuu:
[DEMO LINK](https://aiiyuu.github.io/react_todo-app-add-and-delete/)


src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,79 @@
...
+      .finally(() => {
+        setIsDisabled(false);
+        hideTempTodo();
+
+        // I need this timeout to not fail tests
+        setTimeout(() => {
+          focusInput();
+        }, 100);
+      });
+  };
```
781calmstate, line 62: use `async await` with `try...catch` instead.

this way you will have all functionality you need but much more readable and clear

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,79 @@
...
+  throwErr,
+  focusInput,
+}: Props) => {
+  const [isDisabled, setIsDisabled] = useState(false);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!input.current) {
+      throw new Error('Title input is not found');
```
781calmstate, line 28: make sure you process such errors so they don't crash your app if left unhandled

781calmstate (approved):
Should I be able to complete todos? in the working example I can, but in your demo I can't.

I remember explaining to you in one of your PRs about memoization and as I can see you decided to continue memoizing functionality in your code. It's your choice, but remember, trying to optimize simple and non-heavy operations by using memoization you can make it only harder for react, because memoization is an additional operations which react performs to compare

Aiiyuu:
> Should I be able to complete todos? in the working example I can, but in your demo I can't.
> 
> I remember explaining to you in one of your PRs about memoization and as I can see you decided to continue memoizing functionality in your code. It's your choice, but remember, trying to optimize simple and non-heavy operations by using memoization you can make it only harder for react, because memoization is an additional operations which react performs to compare

I completed this task before you explained memoization to me. I only pushed this task a week after implementing it because I encountered some issues with tests. So i didn't want to implement everything from the begining.

In the next tasks, I will consider your explanation and the effort you put into helping me understand the concept of memoization.

Also, it’s not supposed to have the ability to complete todos, since that functionality is implemented in the third part.

PR #1987, opened 2025-10-20
Develop

Romanov7788:
[DEMO LINK](https://Romanov7788.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -9,18 +9,10 @@ export const App: React.FC = () => {
   if (!USER_ID) {
     return <UserWarning />;
```
FaiHamid, line 10: Here USER_ID = 0, so !USER_ID always will be true, please change value for USER_ID.

src/components/Header.tsx
```diff
@@ -0,0 +1,22 @@
...
+        className="todoapp__toggle-all active"
+        data-cy="ToggleAllButton"
+      />
+
+      <form>
+        <input
+          data-cy="NewTodoField"
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
```
FaiHamid, line 17: No onClick handler has been added, so this code clearly does not add a todo.

FaiHamid (changes requested):
Good job!
Let`s just improve your code a little:

1.You have a footer and header in TodoList, although logically, all of this should be in App. TodoList should only have a section with todos mapping.
image

2.There is a discrepancy between how the code I copied from your repository works and the deployed code.

PR #1988, opened 2025-10-20
Develop

zhyliaieva:
[DEMO LINK](https://zhyliaieva.github.io/react_todo-app-add-and-delete/)


etojeDenys (changes requested):
add and delete logic is good, but the layout looks broken

1. fix the position of the "toggle all" button
image

2. there should not be any text
image

Anton-Kuchmasov (changes requested):
Hey, comments from previous review didn't fixed yet:
image


You can use this [link](https://mate-academy.github.io/react_todo-app-with-api/) as a reference. Just open a devTools in browser, explore stylenames & other attributes and compare them with your layout

Anton-Kuchmasov (changes requested):
Take a quick look here:
image


This block shouldn't appear if there are no notifications right now
Please double-check styles on the reference:
https://mate-academy.github.io/react_todo-app-with-api/

Get rid of this UI bug and your solution will be approved!

etojeDenys (approved):
good job

PR #1989, opened 2025-10-22
feat: implement adding and deleting todos

S-Khraban:
[DEMO LINK](https://S-Khraban.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -50,7 +50,7 @@ export const App: React.FC = () => {
         const data = await getTodos();
         setTodos(data);
       } catch {
-        setErrorMessage('Unable to load todos');
+        showError('Unable to load todos');
       }
     })();
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import cn from 'classnames';
 import { UserWarning } from './UserWarning';
 import { USER_ID, getTodos, createTodo, deleteTodo } from './api/todos';
@@ -6,16 +6,14 @@ import type { Todo } from './types/Todo';
 import { Filter, FilterStatus } from './components/Filter';
 import { Notification } from './components/Notification';
 import { TodoList } from './components/TodoList';
+import { getFilteredTodos } from './utils/getFilteredTodos';
 
 export const App: React.FC = () => {
   const shouldShowUserWarning = !USER_ID;
 
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [, setIsLoading] = useState(false);
   const [errorMessage, setErrorMessage] = useState<string | null>(null);
-  const [filterStatus, setFilterStatus] = useState<FilterStatus>(
-    FilterStatus.All,
-  );
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>(FilterStatus.All);
   const [title, setTitle] = useState('');
   const [isAdding, setIsAdding] = useState(false);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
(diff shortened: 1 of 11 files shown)


src/components/Filter.tsx
```diff
@@ -0,0 +1,51 @@
...
+type Props = {
+  value: FilterStatus;
+  onChange: (nextStatus: FilterStatus) => void;
+};
+
+const FILTERS: FilterStatus[] = [
+  FilterStatus.All,
+  FilterStatus.Active,
+  FilterStatus.Completed,
+];
```
Aiiyuu, line 18: The original implementation uses a `FILTERS` constant array to store the enum values, which is unnecessary because we can directly iterate over the `FilterStatus` enum itself. By using `Object.values(FilterStatus)`, we can dynamically generate the filter buttons without the need for an additional constant.


```javascript
{Object.values(FilterStatus).map(status => (
  <a
    key={status}
    href={getHref(status)}
    className={cn('filter__link', { selected: value === status })}
    data-cy={`FilterLink${toTitle(status)}`}
    onClick={event => handleLinkClick(status, event)}
    aria-current={value === status ? 'page' : undefined}
  >
    {toTitle(status)}
  </a>
))}
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <div
+              data-cy="TodoLoader"
+              className={cn('modal overlay', {
+                'is-active': isTemp || isPending,
+              })}
+            >
+              <div className="modal-background has-background-white-ter" />
+              <div className="loader" />
+            </div>
+          </div>
```
Aiiyuu, line 65: According to the checklist item №3
`3. Code should be split into small, reusable components if it possible (Filter, TodoList, Todo, NewTodo)`

it would be better to break `TodoList` into `TodoList` and `Todo`

src/components/TodoList.tsx
```diff
@@ -0,0 +1,70 @@
...
+            )}
+
+            <div
+              data-cy="TodoLoader"
+              className={cn('modal overlay', {
+                'is-active': isTemp || isPending,
+              })}
+            >
+              <div className="modal-background has-background-white-ter" />
+              <div className="loader" />
```
Aiiyuu, line 63: This is not required by the task, but considering that you're using the loader in many places across many components, I would move it into a dedicated component because, why not?

src/App.tsx
```diff
@@ -1,26 +1,253 @@
...
+    setErrorMessage(null);
+  }, [clearErrorTimer]);
+
+  const showError = useCallback(
+    (message: string) => {
+      setErrorMessage(message);
+      clearErrorTimer();
+      hideErrorTimeoutIdRef.current = window.setTimeout(() => {
+        setErrorMessage(null);
+      }, 3000);
```
Aiiyuu, line 51: You can move it into a dedicated constant

```javascript
const ERR_MESSAGE_DELETION_INTERVAL = 3000;
```

src/App.tsx
```diff
@@ -1,26 +1,253 @@
...
+
+    return () => clearErrorTimer();
+  }, [hideError, showError, clearErrorTimer]);
+
+  const filteredTodos =
+    filterStatus === FilterStatus.Active
+      ? todos.filter(todo => !todo.completed)
+      : filterStatus === FilterStatus.Completed
+        ? todos.filter(todo => todo.completed)
+        : todos;
```
Aiiyuu: That looks a little bit messy 😅😅

Aiiyuu (approved):
**Overall, well done!** ❤️❤️

src/components/Filter.tsx
S-Khraban: I prefer keeping the explicit FILTERS array since it provides a stable and controlled order of filters. Using Object.values() would make the order dependent on the enum definition and might unintentionally include future values not meant for display.

src/components/TodoList.tsx
S-Khraban: Thats good point. Thanks!

src/App.tsx
S-Khraban: Yes, you’re right. I’ll do it using a switch.

781calmstate:
I clicked enter and I got error "Title should not be empty" which is correct, but while the errors was fading away the text has changed to this
image


Should I be able to check completed todos, or this functionality isn't required in the task?

src/App.tsx
```diff
@@ -1,26 +1,261 @@
...
+      filteredTodos = todos.filter(todo => !todo.completed);
+      break;
+
+    case FilterStatus.Completed:
+      filteredTodos = todos.filter(todo => todo.completed);
+      break;
+
+    default:
+      filteredTodos = todos;
+  }
```
781calmstate: move it to helper

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+import cn from 'classnames';
+import type { Todo } from '../types/Todo';
+
+type Props = {
+  todo: Todo;
+  isPending: boolean;
+  onDelete: (id: number) => void;
+};
+
+export const TodoItem: React.FC<Props> = React.memo(
```
781calmstate: why do you need the memo here?

if you wanted to memoize your component, it seems like you failed, because your `TodoItem` accept `onDelete` as prop which is `handleDelete` from `App`. The `handleDelete` isn't memoize and since it's a non-primitive, it's reference changing every rerender of `App` component which breaks the memoization of `TodoItem` since every rerender your `onDelete` is a new reference which makes the `memo` think the props have changed

src/App.tsx
```diff
@@ -1,26 +1,261 @@
...
+import { Filter, FilterStatus } from './components/Filter';
+import { Notification } from './components/Notification';
+import { TodoList } from './components/TodoList';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const shouldShowUserWarning = !USER_ID;
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [, setIsLoading] = useState(false);
```
781calmstate: if you don't use the state value, then why do you need the useState at all?

781calmstate (changes requested):

src/App.tsx
```diff
@@ -1,26 +1,223 @@
...
+import { USER_ID, getTodos, createTodo, deleteTodo } from './api/todos';
+import type { Todo } from './types/Todo';
+import { Filter, FilterStatus } from './components/Filter';
+import { Notification } from './components/Notification';
+import { TodoList } from './components/TodoList';
+import { getFilteredTodos } from './utils/getFilteredTodos';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const shouldShowUserWarning = !USER_ID;
```
mkyloa: not sure it requires a specific variable

src/App.tsx
```diff
@@ -1,26 +1,223 @@
...
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>(FilterStatus.All);
+  const [title, setTitle] = useState('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [pendingIds, setPendingIds] = useState<Set<number>>(new Set());
```
mkyloa: whoa
can you explain what are the benefits of using set here?

src/App.tsx
```diff
@@ -1,26 +1,223 @@
...
+    hideError();
+    setTimeout(() => newTodoInputRef.current?.focus(), 0);
+  };
+
+  const handleNewTodoFormSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    const trimmed = title.trim();
+
+    if (!trimmed) {
+      showError('Title should not be empty');
```
mkyloa: let's enumerate the errors

src/App.tsx
```diff
@@ -1,26 +1,223 @@
...
+    }
+
+    try {
+      setIsAdding(true);
+      setTempTodo({ id: 0, title: trimmed, completed: false, userId: USER_ID });
+      const created = await createTodo(trimmed);
+      setTodos(curr => [...curr, created]);
+      setTitle('');
+    } catch {
+      showError('Unable to add a todo');
```
mkyloa: same here

mkyloa (approved):

PR #1992, opened 2025-10-26
Solution #1

michal-szwindowski:
[DEMO](https://n3di.github.io/react_todo-app-add-and-delete/)


Zibi95 (changes requested):
Missing DEMO link

src/components/FilterLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+export const FilterLink: React.FC<{
+  filter: Filter;
+  currentFilter: Filter;
+  setFilter: (filter: Filter) => void;
+  label: string;
+  dataCy: string;
+}> = ({ filter, currentFilter, setFilter, label, dataCy }) => (
+  <a
+    href={`#/${filter}`}
+    className={`filter__link ${currentFilter === filter ? 'selected' : ''}`}
```
zibi-dev: You have installed cool package called classnames check it out

src/App.tsx
```diff
@@ -1,26 +1,223 @@
...
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isError, setIsError] = useState<string | null>(null);
+  const [filter, setFilter] = useState<Filter>('all');
+
+  const [title, setTitle] = useState('');
```
zibi-dev: Are you sure we need state for input in this case?

zibi-dev: Remember that each keystroke causes this and all its children rerender

zibi-dev: You can simply use onSubmit actions and read data from input internal state not this one.

src/App.tsx
```diff
@@ -1,26 +1,223 @@
...
+
+    results.forEach((res, idx) => {
+      const id = completed[idx].id;
+
+      if (res.status === 'fulfilled') {
+        succeededIds.push(id);
+      } else {
+        hadError = true;
+      }
+    });
```
zibi-dev: Looks like good case for reducer method

src/App.tsx
```diff
@@ -1,26 +1,223 @@
...
+    );
+
+    if (hadError) {
+      setIsError(ErrorType.DELETE_TODO);
+    } else {
+      requestAnimationFrame(() => {
+        (
+          document.getElementById('newTodoField') as HTMLInputElement | null
+        )?.focus();
+      });
```
zibi-dev: maybe create helper function for this

Zibi95 (changes requested):
Nice

danon321 (approved):

PR #2007, opened 2025-10-29
add solution

dashasssss:
https://dashasssss.github.io/react_todo-app-add-and-delete/


Bohdan-Diatliuk (approved):
It's better to break it down into its components. Nevertheless, this is an excellent solution.

781calmstate (approved):
Next time make sure to split logic into components. React is all about reusable components and custom hooks, you should learn to stick to this approach

PR #2009, opened 2025-10-30
[Todo App add and delete React-TS]

Bohdan-Diatliuk:
[DEMO LINK](https://bohdan-diatliuk.github.io/react_todo-app-add-and-delete/)


vtmnkmarta (approved):
Goood job!😀

src/App.tsx
```diff
@@ -1,26 +1,241 @@
...
+
+  useEffect(() => {
+    setErrorMessage('');
+
+    focusedInput.current?.focus();
+
+    getTodos()
+      .then(setTodos)
+
+      .catch(() => {
```
Moroz-Dmytro, line 51:
```suggestion
      .then(setTodos)
      .catch(() => {
```

src/App.tsx
```diff
@@ -1,26 +1,241 @@
...
+          todos={filteredTodos}
+          tempTodo={tempTodo}
+          deletingTodoIds={deletingTodoIds}
+          updatingTodoIds={updatingTodoIds}
+          onToggle={handleToggleTodo}
+          onDelete={handleDeleteTodo}
+          onRename={handleRenameTodo}
+        />
+        {todos.length > 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
Moroz-Dmytro, line 218: Move footer and header to separate components

src/App.tsx
```diff
@@ -1,26 +1,241 @@
...
+
+    try {
+      const todo = await postTodos(newTodo);
+
+      setTodos(prev => [...prev, todo]);
+    } catch (error) {
+      setErrorMessage('Unable to add a todo');
+      focusedInput.current?.focus();
+      setTimeout(() => setErrorMessage(''), 3000);
+      throw error;
```
Moroz-Dmytro, line 93: Why do you need to throw new error?

Moroz-Dmytro (changes requested):
When I try to delete todo in offline mode I see infinite loader 
image


Also fix tests

src/App.tsx
Bohdan-Diatliuk: throw error is needed here so that:
The input is not cleared when an error occurs,
The NewTodo component can catch the error via a reject promise.

PR #2014, opened 2025-10-31
add task solution

Abdulahhh2005:
[DEMO LINK](https://Abdulahhh2005.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,199 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (error) {
+      const timeout = setTimeout(() => setError(null), 3000);
+
+      return () => clearTimeout(timeout);
+    }
+
+    return undefined; // <-- добавляем явный return для всех других путей
```
h-amster, line 56: 1. remove all comments before submitting a task
2. replace return undefined with return

src/App.tsx
```diff
@@ -1,26 +1,199 @@
...
+      })
+      .finally(() => setIsInputDisabled(false));
+  };
+
+  const deleteTodo = (id: number) => {
+    setIsInputDisabled(true);
+    setDeletedIds(prev => [...prev, id]);
+    deleteTodos(id)
+      .then(() => {
+        setTodos(current => current.filter(t => t.id !== id));
```
h-amster, line 90: make more declarative variable names. Replace t with todo. Do it in all similar cases

src/App.tsx
```diff
@@ -1,26 +1,199 @@
...
+          setError(Error.Delete_todo);
+        }
+      })
+      .catch(() => {
+        setError(Error.Delete_todo);
+      })
+      .finally(() => {
+        setDeletedIds([]);
+        setIsInputDisabled(false);
+      });
```
h-amster, line 130: can be simplified with idsToDelete.forEach(id => deleteTodo(id));

src/components/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <a
+          href="#/completed"
+          className={cn('filter__link', {
+            selected: filter === Filter.Completed,
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setFilter(Filter.Completed)}
+        >
+          Completed
+        </a>
```
h-amster, line 58: You can use Object.values(Filter) and map method to render these links

h-amster (approved):
Good job! Check my comments before moving to next tasks

PR #2023, opened 2025-11-05
solution add-delete functionality

andriiLink:
[DEMO LINK](https://andriiLink.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,8 +19,8 @@ export const App: React.FC = () => {
   const [status, setStatus] = useState(Status.All);
   const [errorType, setErrorType] = useState<ErrorType | null>(null);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
-  //const [deleteLoading, setDeleteLoading] = useState(false);
-  //const [todoToDeleteIds, setTodoToDeleteIds] = useState<Todo['id'][]>([]);
+  const [deleteLoading, setDeleteLoading] = useState(false);
+  const [todoToDeleteIds, setTodoToDeleteIds] = useState<Todo['id'][]>([]);
 
   useEffect(() => {
     setLoader(true);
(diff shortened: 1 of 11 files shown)


anastasia-kravchuk (approved):

src/App.tsx
```diff
@@ -1,26 +1,251 @@
...
+            return todo.completed;
+
+          default:
+            return new Error('unknown Status has been selected!');
+        }
+      });
+    }
+
+    return preparedTodos;
+  };
```
alexander-ignatow, line 80: This function should be outside of the component

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,61 @@
+import { forwardRef, Ref } from 'react';
+import { Todo } from '../types/Todo';
+import cn from 'classnames';
+
+type Props = {
+  todos: Todo[];
+  inputValue: string;
+  onChangeInput: (value: string) => void;
```
alexander-ignatow, line 8: Why not to store the input text inside this component and only trigger the requests for the TODO creation?

src/components/TodoMain.tsx
```diff
@@ -0,0 +1,111 @@
...
+  loader: boolean;
+  tempTodo: Todo | null;
+  onDelete: (todoId: Todo['id']) => void;
+};
+
+export const TodoMain: React.FC<Props> = ({
+  visibleTodos,
+  deleteTodoIds,
+  loader,
+  // deleteLoading,
```
alexander-ignatow, line 19: don't keep the commented out code like this

src/components/TodoMain.tsx
```diff
@@ -0,0 +1,111 @@
...
+type Props = {
+  visibleTodos: Todo[];
+  deleteLoading: boolean;
+  deleteTodoIds: number[];
+  loader: boolean;
+  tempTodo: Todo | null;
+  onDelete: (todoId: Todo['id']) => void;
+};
+
+export const TodoMain: React.FC<Props> = ({
```
alexander-ignatow, line 15: Consider renaming this component - TodoMain is an undescriptive name
Probably it makes sense to split this component into TodoList and TodoItem?

alexander-ignatow (approved):
Lets move on, but the mentioned errors are needed to be fixed

PR #2029, opened 2025-11-09
add task solution

webqarbon:
https://webqarbon.github.io/react_todo-app-add-and-delete/


src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (!error) {
+      return;
+    }
+
+    const timerId = setTimeout(() => {
+      onClose();
+    }, 3000);
+
+    return () => clearTimeout(timerId);
+  }, [error, onClose]);
```
alexander-ignatow, line 20: This component has a potential bug

1 non-empty error is passed
2 in 2,5 seconds different non-empty error is passed
3 in 0,5 seconds after the second error appeared it is cleared

It is rare but possible scenario

alexander-ignatow (approved):
Looks good, but think about the issue for the next iteration

PR #2039, opened 2025-11-22
make/tasks

YehorHeiko:
[DEMO LINK](https://YehorHeiko.github.io/react_todo-app-add-and-delete/)


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+        data-cy="TodoDelete"
+        onClick={() => deleteTodo(visibleGoods.id)}
+      >
+        ×
+      </button>
+
+      {/* overlay will cover the todo while it is being deleted or updated */}
+      <div data-cy="TodoLoader" className="modal overlay">
+        <div className="modal-background has-background-white-ter" />
+        <div className="loader" />
```
Anton-Kuchmasov: Here you should add class `is-active` as condition - if todo is not processing (f.e. deleting or adding) - there should be a visible spinner (loader)

Take a look here:
https://github.com/mate-academy/react_todo-app-add-and-delete/pull/2037/files#diff-12d531501e4f46a3bc96ffd1b7cd9c0a712f46f8730388219159d047ca94c26b

Anton-Kuchmasov (changes requested):
Hey, your solution aren't passing all tests. If you're stuck, feel free to ask in fe_chat

Also check out comment below - it might be helpful

etojeDenys (changes requested):
almost done

1. after trying to delete several todos in a short period of time, the loader only shows up on the last one. however, it should show up on every todo that is deleted

YehorHeiko:
Everything seems fine, but I can't track down this bug. According to the task specifications, everything was completed correctly.

etojeDenys (changes requested):
there is a problem with the 'isDeleting' value. currently, it is either true or false. however, to fix the problem mentioned in the last review, it should be an array with the id of the todos that should have a loader
image

PR #2043, opened 2025-12-02
add task solution

serhiy23471:
[DEMO LINK](https://serhiy23471.github.io/react_todo-app-add-and-delete/)


etojeDenys (changes requested):
it looks good, but the demo link does not work. also, it would be nice to remove redundant comments

etojeDenys (approved):
great job

PR #2045, opened 2025-12-08
Develop

ArtemYakhno:
(https://ArtemYakhno.github.io/react_todo-app-add-and-delete/)


ArtemYakhno:
Unfortunately, when I added the TransitionGroup animation, it worked very badly. Then I added framer-motion, which worked, but the tests could not pass (So I removed the animation for now, I will try to implement it in the next part of this work

etojeDenys (approved):
great job

PR #2047, opened 2025-12-11
solution

olhatrokhymchuk:
[DEMO LINK](https://olhatrokhymchuk.github.io/react_todo-app-add-and-delete/)


etojeDenys (changes requested):
1. after attempting to delete several tasks in a short period of time, the loader only appears on the last task but it should appear on every task that will be deleted

2. after pressing "clear completed," the loader does not show up on the todos

etojeDenys (approved):
great job

PR #2052, opened 2025-12-22
Develop

yahohulia:
[DEMO LINK](https://yahohulia.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,208 @@
...
+        .catch(() => {
+          setErrorMessage('Unable to delete a todo');
+        })
+        .finally(() => {
+          setLoadingIds(prev => prev.filter(id => id !== todo.id));
+          focusInput();
+        });
+    });
+  };
+
```
etojeDenys, line 159: it would be better to use promise.all() here

etojeDenys (approved):
good job

PR #2062, opened 2026-01-07
add solution

6ORSH:
[DEMO LINK](https://6ORSH.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,138 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { addTodo, deleteTodo, getTodos, USER_ID } from './api/todos';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { NewTodo } from './types/NewTodo';
+import { Todo } from './types/Todo';
+import { TodoErrors, TodosFilter } from './types/enums';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<TodosFilter>(TodosFilter.All);
+  const [error, setError] = useState<TodoErrors | null>(null);
+  const [isErrorShown, setIsErrorShown] = useState(false);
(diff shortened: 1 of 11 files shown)


src/types/enums.ts
```diff
@@ -0,0 +1,13 @@
...
+  AddError = 'AddError',
+  DeleteError = 'DeleteError',
+  UpdateError = 'UpdateError',
+}
+
+export enum TodosFilter {
+  All = 'All',
+  Active = 'Active',
+  Completed = 'Completed',
+}
```
Anton-Kuchmasov, line 13: Consider using UPPER_CASE for enum's fields naming

src/types/NewTodo.ts
```diff
@@ -0,0 +1,5 @@
+export interface NewTodo {
+  userId: number;
+  title: string;
+  completed: boolean;
+}
```
Anton-Kuchmasov, line 5:
```suggestion
`export type NewTodo = Omit<Todo, 'id'>;
````

Comment (Anton-Kuchmasov) [reply src/types/NewTodo.ts:5]: or, if you prefer interfaces:

`export interface NewTodo extends Omit<Todo, 'id'> {}
`

File: src/App.tsx [hunk 3]
```diff
@@ -1,26 +1,138 @@
...
+  };
+
+  const clearCompleted = () => {
+    todos
+      .filter(todo => todo.completed)
+      .forEach(todo => {
+        removeTodo(todo.id);
+      });
+  };
+
```
Anton-Kuchmasov, line 99: DON'T modify variable from State at any way except `setState`.
Never, it's anti-pattern, it violates React rules


`const newTodos = todos
      .filter(todo => todo.completed)
      .forEach(todo => {
        removeTodo(todo.id);
      }); 
      setTodos(newTodos);`

Anton-Kuchmasov (approved):
Good job!

Check out comments below - there are a few issues need to be fixed in next chapter

src/App.tsx
6ORSH: I don't get your point. Where do I modify state, as you said, if I used non-mutating methods?
I just got an array of todo IDs to remove, and for every ID perform deletion with removeTodo(), where it's done via state.
`  const removeTodo = (id: number) => {
    deleteTodo(id)
      .then(() => {
        setTodos(prevTodos => prevTodos.filter(todo => todo.id !== id));
      })
      .catch(() => {
        displayError(TodoErrors.DeleteError);
      })
      .finally(() => {
        titleInputRef.current?.focus();
      });
  };`

I did it this way to complete the `- the deletion should work as several individual deletions running at the same time;` requirement.

PR #2063, opened 2026-01-08
Completed Task

Augustwise:
[DEMO LINK](https://Augustwise.github.io/react_todo-app-add-and-delete/)

I will add animations in 3 part of the task

To meet the requirement that "the deletion should work as several individual deletions running at the same time," I used `Promise.allSettled`.
There may be an easier way, but I haven't figured out yet.


src/types/ErrorMessage.ts
```diff
@@ -0,0 +1,7 @@
+export enum ErrorMessage {
+  Load = 'Unable to load todos',
+  Update = 'Unable to update a todo.',
+  Delete = 'Unable to delete a todo',
+  Title = 'Title should not be empty',
+  Add = 'Unable to add a todo',
+}
```
Anton-Kuchmasov, line 7: Consider using UPPERCASE e.g. `LOAD, UPDATE` etc

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import classNames from 'classnames';
+import { ErrorMessage } from '../types/ErrorMessage';
+
+interface Props {
+  errorMessage: ErrorMessage | '';
```
Anton-Kuchmasov, line 6: better use`null` for empty Error state cases

src/components/Header.tsx
```diff
@@ -0,0 +1,63 @@
...
+      return;
+    }
+
+    const trimmedTitle = title.trim();
+
+    onAdd(trimmedTitle)
+      .then(() => {
+        setTitle('');
+      })
+      .catch(() => {});
```
Anton-Kuchmasov, line 35: looks redunant

src/components/Header.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const [title, setTitle] = useState('');
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (disabled) {
+      return;
+    }
+
+    const trimmedTitle = title.trim();
```
Anton-Kuchmasov, line 29: what if trimmedTitle = `''`?

better organize additional check here to prevent unnecessary API call

Anton-Kuchmasov (approved):
Well done!

You're absolutely right - `Promise.allSettled()` is the best (probably, the only?) way to do required bulk action.

Check out comments below - they might be helpful

PR #2069, opened 2026-01-16
add task solution

devsync25:
https://devsync25.github.io/react_todo-app-add-and-delete/


src/App.tsx
```diff
@@ -1,26 +1,191 @@
...
+        showError(ErrorMessage.Delete);
+        throw new Error(); // Важно для цепочки промисов
+      })
+      .finally(() => setLoadingIds(prev => prev.filter(curr => curr !== id)));
+  };
+
+  const clearCompleted = () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+    completedTodos.forEach(todo => handleDelete(todo.id));
+  };
```
etojeDenys, line 95: it would be better to use promise.all() here

etojeDenys (approved):
good job

PR #2083, opened 2026-01-26
solution

Nabeyko:
- [DEMO LINK](https://Nabeyko.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,164 @@
...
+
+  const handleError = (newError: ErrorType | null) => {
+    setIsError(newError);
+  };
+
+  const handleDeleteCompleted = () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    completedTodos.map(todo => handleDeleteTodo(todo.id));
+  };
```
etojeDenys, line 144: it would be better to use Promise.all() here

etojeDenys (changes requested):
almost done

1. after trying to delete several todos, the loader only shows up on the last pressed one instead of every task that is going to be deleted

src/App.tsx
```diff
@@ -1,26 +1,179 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosToDisplay, setTodosToDisplay] = useState<Todo[]>([]);
+
+  const [selectedValue, setSelectedValue] = useState<FilterStatus>(
+    FilterStatus.All,
+  );
+
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [isError, setIsError] = useState<ErrorType | null>(null);
+  const [deletingTodoId, setDeletingTodoId] = useState<number | null>(null);
```
etojeDenys: there is currently a problem with deleting. if a user tries to delete several tasks, the loader will only appear on the last one selected. To fix this, you should change the number to an array of numbers
```suggestion
  const [deletingTodoId, setDeletingTodoId] = useState<number[] | null>(null);
```

etojeDenys (changes requested):

etojeDenys (approved):
great job

PR #2089, opened 2026-01-28
Solution

dmytro-halieba:
[DEMO LINK](https://dmytro-halieba.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,22 +1,25 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useState, useEffect, useRef } from 'react';
-import { Todo } from './types/Todo';
 import cn from 'classnames';
 
-// import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
+import {
+  Todo,
+  FilterStatus,
+  ErrorMessage,
+  TODO_FILTER_NAV_CONFIG,
+} from './types';
+import { getVisibleTodos } from './utils/getVisibleTodos';
 
 import * as todoService from './api/todos';
-
-export type FilterStatus = 'All' | 'Active' | 'Completed';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState('');
-  const [filter, setFilter] = useState<FilterStatus>('All');
+  const [filter, setFilter] = useState<FilterStatus>(FilterStatus.All);
   const [query, setQuery] = useState('');
   const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
 
@@ -52,7 +55,7 @@ export const App: React.FC = () => {
         );
       })
       .catch(() => {
-        handleError('Unable to delete a todo');
+        handleError(ErrorMessage.DeleteTodoFailed);
       })
       .finally(() => {
         setDeletingTodoIds(current => current.filter(id => id !== todoId));
@@ -66,7 +69,7 @@ export const App: React.FC = () => {
     const trimmedTitle = query.trim();
 
     if (trimmedTitle.length === 0) {
-      handleError('Title should not be empty');
+      handleError(ErrorMessage.EmptyTitle);
 
       return;
     }
@@ -89,7 +92,7 @@ export const App: React.FC = () => {
         setQuery('');
       })
       .catch(() => {
-        handleError('Unable to add a todo');
+        handleError(ErrorMessage.AddTodoFailed);
       })
       .finally(() => {
         setLoading(false);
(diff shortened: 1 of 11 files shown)


dianakerekesha (approved):
Good job.

src/App.tsx
```diff
@@ -1,26 +1,312 @@
...
+                onClick={() => setFilter('Active')}
+                data-cy="FilterLinkActive"
+              >
+                Active
+              </a>
+
+              <a
+                href="#/completed"
+                className={cn('filter__link', {
+                  selected: filter === 'Completed',
```
dothearts: Create enum for all statuses. 
Then use it in all places on your app.

src/App.tsx
```diff
@@ -1,26 +1,312 @@
...
+          <footer
+            className={cn('todoapp__footer', { hidden: !haveTodos })}
+            data-cy="Footer"
+          >
+            <span className="todo-count" data-cy="TodosCounter">
+              {activeTodosCount} items left
+            </span>
+
+            <nav className="filter" data-cy="Filter">
+              <a
```
dothearts: You can create nav config and then use map  in this place (not critical)

src/App.tsx
```diff
@@ -1,26 +1,312 @@
...
+      });
+  }, []);
+
+  useEffect(() => {
+    if (!loading && !tempTodo && inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [loading, tempTodo]);
+
+  const visibleTodos = todos.filter(todo => {
```
dothearts: In this case, also use Enum. 
You can create helper (pure) func for visibleTodos, something like (getVisibleTodos)

src/App.tsx
```diff
@@ -1,26 +1,312 @@
...
+    setTempTodo(newTempTodo);
+
+    todoService
+      .createTodo({ title: trimmedTitle, userId: USER_ID, completed: false })
+      .then(newTodo => {
+        setTodos(currentTodos => [...currentTodos, newTodo]);
+        setQuery('');
+      })
+      .catch(() => {
+        handleError('Unable to add a todo');
```
dothearts: Can move all error in enum (not critical)

DanilWeda (changes requested):
In general, I liked your solution. 
I added some comment for improve it! 
Many thanks.

src/utils/getVisibleTodos.ts
```diff
@@ -0,0 +1,20 @@
...
+
+      case FilterStatus.Active:
+        return !todo.completed;
+
+      case FilterStatus.Completed:
+        return todo.completed;
+
+      default:
+        return true;
+    }
```
ArturErmolenko, line 18:
```suggestion
    switch (filter) {
      case FilterStatus.Active:
        return !todo.completed;

      case FilterStatus.Completed:
        return todo.completed;
        
      case FilterStatus.All:
      default:
        return true;
    }
```
you can combine cases

src/App.tsx
```diff
@@ -1,26 +1,290 @@
...
+
+import * as todoService from './api/todos';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState('');
```
ArturErmolenko, line 21:
```suggestion
  const [error, setError] = useState<ErrorMessage>();
```
where is type?

src/App.tsx
```diff
@@ -1,26 +1,290 @@
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  function handleStatusSwitch(todoId: number, todoStatus: boolean) {}
+
+  function handleErrorReset() {
+    setError('');
+  }
+
+  function handleDeleteTodo(todoId: number) {
```
ArturErmolenko, line 47: why you use function declaration?
use expression

src/App.tsx
```diff
@@ -1,26 +1,290 @@
...
+
+    setTimeout(() => {
+      setError('');
+    }, 3000);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  function handleStatusSwitch(todoId: number, todoStatus: boolean) {}
+
+  function handleErrorReset() {
```
ArturErmolenko, line 43: why we need this?

src/App.tsx
```diff
@@ -1,26 +1,290 @@
...
+  function handleError(message: string) {
+    setError(message);
+
+    setTimeout(() => {
+      setError('');
+    }, 3000);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  function handleStatusSwitch(todoId: number, todoStatus: boolean) {}
```
ArturErmolenko, line 41:
```suggestion
```
why?
KISS

src/App.tsx
```diff
@@ -1,26 +1,290 @@
...
+        handleError(ErrorMessage.AddTodoFailed);
+      })
+      .finally(() => {
+        setLoading(false);
+        setTempTodo(null);
+      });
+  }
+
+  useEffect(() => {
+    setError('');
```
ArturErmolenko, line 104:
```suggestion
```

ArturErmolenko (approved):
I approved this part but you should fix comments in next part
It's already a good time to split your code on components

PR #2119, opened 2026-02-09
Solution

Kawka06:
[DEMO LINK](https://Kawka06.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,455 @@
...
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<FilterStatus>('all');
+
+  const [error, setError] = useState<string>('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  const [newTitle, setNewTitle] = useState('');
```
zibi-dev, line 39: Are u sure that we need input value as state? I think that we can simply react to the form submit event and get value from that event argument. https://developer.mozilla.org/en-US/docs/Web/API/FormData

src/App.tsx
```diff
@@ -1,26 +1,455 @@
...
+  const [isAdding, setIsAdding] = useState(false);
```
zibi-dev, line 40: We have isAdding, isLoading states it becomes cumbersome and easy to introduce bugs./ https://kentcdodds.com/blog/stop-using-isloading-booleans

src/App.tsx
```diff
@@ -1,26 +1,455 @@
...
+          return;
+        }
+
+        setIsLoading(false);
+      });
+
+    return () => {
+      cancelled = true;
+    };
+  }, []);
```
zibi-dev, line 116: Extra: do some research on when this (useEffect) runs/triggers and whether there’s a potentially better way to load async data.

Here’s a starting point. I know it isn’t in the Mate lessons (as far as I remember), but that’s how it is with libraries—they release new versions with new APIs, so it’s hard to follow and keep the lessons updated. As developers, we should keep our hands on the pulse and read the docs:

https://react.dev/reference/react/Suspense

src/App.tsx
zibi-dev: also its easier to introduce new states

src/App.tsx
```diff
@@ -1,26 +1,455 @@
...
+  const [error, setError] = useState<string>('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  const [newTitle, setNewTitle] = useState('');
+  const [isAdding, setIsAdding] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+
+  const [loadingTodoIds, setLoadingTodoIds] = useState<Set<number>>(
+    () => new Set(),
+  );
```
zibi-dev, line 45: nc use of set

Zibi95 (changes requested):
Nice. I have left a comment about splitting this whole logic and JSX into separate components in PR with previous part.

PR #2131, opened 2026-02-16
solution

nastia7408:
demo: https://nastia7408.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -87,23 +87,33 @@ export const App: React.FC = () => {
   };
 
   const onDeleteAll = () => {
-    const todoComplete = todos.filter(todo => todo.completed);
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,26 +1,258 @@
...
+          setTodos(currentTodos => currentTodos.filter(t => t.id !== todo.id));
+        })
+        .catch(() => {
+          setErrorMessage('Unable to delete a todo');
+        })
+        .finally(() => {
+          setLoadingIds(prev => prev.filter(id => id !== todo.id));
+        });
+    });
+  };
```
etojeDenys, line 117: it is better to use promise.all here

etojeDenys (approved):
good job. just fix this problem before you start the next part

PR #2138, opened 2026-02-21
Develop

maksymivivan2000-dotcom:
https://maksymivivan2000-dotcom.github.io/react_todo-app-add-and-delete/   

Split the App into reusable components (Header, TodoList, TodoItem, Footer, ErrorNotification), moved state and business logic into a custom hook (useTodos.ts) that handles loading, adding, deleting, clearing completed, filtering, error handling, and input focus behavior, kept App.tsx as a composition layer that renders components, reads data and actions from useTodos, and handles only the USER_ID guard, and implemented both the Add Todo and Delete Todo flows.

### DIFF
--- a/src/hooks/useTodos.ts
+++ b/src/hooks/useTodos.ts
@@ -1,0 +1,185 @@
+﻿import { FormEvent, useEffect, useRef, useState } from 'react';
+import { addTodo, deleteTodo, getTodos, USER_ID } from '../api/todos';
+import { Todo } from '../types/Todo';
+
+type FilterStatus = 'all' | 'active' | 'completed';
+
+export const useTodos = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isAdding, setIsAdding] = useState(false);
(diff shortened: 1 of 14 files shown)


Anton-Kuchmasov (changes requested):
DEMO link is not working, please re-deploy your solution and re-request review again

Image

src/hooks/useTodos.ts
```diff
@@ -0,0 +1,185 @@
...
+    setErrorMessage(message);
+  };
+
+  const handleDeleteTodo = (id: number) => {
+    setErrorMessage('');
+    startTodoLoading(id);
+
+    deleteTodo(id)
+      .then(() => setTodos(current => current.filter(todo => todo.id !== id)))
+      .catch(() => showError('Unable to delete a todo'))
```
Anton-Kuchmasov: Would you mind to create custom type / enum `ErrorMessage`, where will be placed all possible error messages?
It will be such a best practice, cause you'll be protected from all possible typos

Anton-Kuchmasov (changes requested):
Good job!

Just one tiny fix and your solution will be approved

etojeDenys (approved):
good job

PR #2144, opened 2026-02-23
Develop

George-Buzhenitsa:
https://George-Buzhenitsa.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,497 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import * as todosServices from './api/todos';
+import { Todo } from './types/Todo';
+import { ErrorType } from './types/Error';
+import classNames from 'classnames';
+import { handleError } from './services/ErrorHandling';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [title, setTitle] = useState('');
+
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,26 +1,504 @@
...
+    const completedTodos = [...todosList].filter(current => current.completed);
+
+    setActiveTodo([...completedTodos]);
+
+    for (const current of todosList) {
+      if (current.completed) {
+        deleteTodo(current.id);
+      }
+    }
+  };
```
etojeDenys: you should use Promise.all() here

etojeDenys (changes requested):
almost done

1. after trying to delete several todos in a short period of time, the loader only shows up on the last todo that was pressed (the same problem occurs with status changes)

George-Buzhenitsa:
> almost done
> 
> 1. after trying to delete several todos in a short period of time, the loader only shows up on the last todo that was pressed (the same problem occurs with status changes)

I did not qiute get the problem( please explain in more detailes

src/App.tsx
```diff
@@ -1,26 +1,516 @@
...
+import { handleError } from './services/ErrorHandling';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [title, setTitle] = useState('');
+
+  const [activeTodo, setActiveTodo] = useState<Todo[]>([]);
+  const [errorType, setErrorType] = useState<ErrorType | null>(null);
```
Anton-Kuchmasov, line 19:
```suggestion
  const [errorMessage, setErrorMessage] = useState<ErrorType | null>(null);
```

or even use just `error, setError` instead of this

Anton-Kuchmasov (approved):
Good job!


In the final chap of this task please do those improvements:
1. Split whole `App` into several components (Header, Footer, TodoList, TodoItem etc.)
2. Fix linter errors

PR #2159, opened 2026-03-09
add task solution

adadiada:


brespect (changes requested):
Good progress, but lets add DEM O LINK and pass all tests before PR submit

etojeDenys (changes requested):
good job, last fixes:

1. there should not be that white space here

Image

2. after trying to delete several todos in a short period of time, the loader only appears on the last todo that was pressed

Anton-Kuchmasov (approved):
Good job!


In the last chap of this task, try to get rid of this UI bug:

Image

PR #2165, opened 2026-03-12
solution

Mykolyan:
https://mykolyan.github.io/react_todo-app-add-and-delete/


brespect (changes requested):
Good progress, let`s check next:
1. Fix all tests, before submitting the PR.
2. There is no ability to select item for multiple delete, add this.

Mykolyan:
>     2. There is no ability to select item for multiple delete, add this.

I don't really get, what i supposed to do with that point, because in the [the working example](https://mate-academy.github.io/react_todo-app-with-api/) from task, there is only ability to delete task either each one, by clicking x, either clear all completed. So despite that, in working example there is only two ways of deleting items, i should add ability for select item for multiple delete?

Denys-Kravchuk9988 (approved):
Looks good

PR #2169, opened 2026-03-16
Solution

Igor-hrm:
[DEMO LINK](https://Igor-hrm.github.io/react_todo-app-add-and-delete/)


IanLuan (changes requested):
Good job, your solution is working very well. However, the component structure could be improved. Instead of writing the todo logic inside the `Section` component, you could split it into separate components like `TodoList` and `TodoItem`. This would make your code more readable and easier to maintain.

pedro-ruas (approved):

PR #2172, opened 2026-03-17
sol1

7uckingbad:
[DEMO LINK](https://7uckingbad.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,231 @@
...
+    }
+  };
+
+  const handleClearCompleted = () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    completedTodos.forEach(todo => {
+      handleDelete(todo.id);
+    });
+  };
```
etojeDenys, line 43: you should use promise.all() here

etojeDenys (approved):
good job, jsut fix handleClearCompleted and you can move on to the next step

PR #2176, opened 2026-03-20
solution

despair849:
[DEMO LINK](https://despair849.github.io/react_todo-app-add-and-delete/)


etojeDenys (changes requested):
1. make sure u pass all the tests

Image

2. after attempting to delete a todo, a loader should appear on each todo that will be deleted

3. 'clear completed' button should work (use Promise.all here)

4. split the app into several components (for example: header, main, footer, and error notification)

despair849:
Can u help me , Denys? Because I can't figure this out

etojeDenys (changes requested):
sure, but it would be better if you asked for help in the chat or texted me because it's uncomfortable to communicate here

2pasha (approved):
lgtm 🔥

PR #2179, opened 2026-03-22
Task solution

Igrik74:
[DEMO LINK](https://Igrik74.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,339 @@
...
+                className={classNames('filter__link', {
+                  selected: filter === FilterStatus.Completed,
+                })}
+                onClick={event => {
+                  event.preventDefault();
+                  setFilter(FilterStatus.Completed);
+                }}
+              >
+                Completed
+              </a>
```
etojeDenys, line 302: you can create an array with the necessary parameters and then map through it to create links and avoid code repetition

src/App.tsx
```diff
@@ -1,26 +1,339 @@
...
+      ...newTodo,
+    });
+
+    createTodo(newTodo)
+      .then(createdTodo => {
+        setTodos(currentTodos => [...currentTodos, createdTodo]);
+        setQuery('');
+      })
+      .catch(() => {
+        showError('Unable to add a todo');
```
etojeDenys, line 130: create enum for errors as well

etojeDenys (changes requested):
good job
to improve:
1. split the app into several components (for example: header, main, footer, errorNotification)

PR #2188, opened 2026-04-02
add task solution

Lazarin123:
[DEMO LINK](https://Lazarin123.github.io/react_todo-app-add-and-delete/)


IanLuan (changes requested):
Great progress! However, the tests are currently failing. Also, please refactor your code by splitting it into smaller components. Currently, all elements are contained within App.tsx.

Lazarin123:
Oi ian, fiz com o Pedro hoje e ele falou que era isso e estava tudo ok, pode validar a possível aprovação no projeto?

IanLuan (changes requested):
Hi Samuel! Great refactor, the code structure looks much cleaner. However, the tests are still failing. If you check the test feedback, the issue happens when clearing completed todos.

It seems that when one of the delete requests fails, the UI is not updated correctly. The todos that were successfully deleted are still being displayed. This is likely happening because you're using Promise.all, which rejects immediately if any request fails and prevents handling partial successes. Take a look at that part and consider how you could handle successful and failed deletions separately so the UI stays in sync with the actual results.

Lazarin123:
foi feito tudo com os outros professores e falaram que estava tudo ok

pedro-ruas (approved):

PR #2192, opened 2026-04-13
add task solution

AndrewPrszn:
[DEMO LINK](https://AndrewPrszn.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,25 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
+
 import React, { useEffect, useMemo, useRef } from 'react';
+
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
-import { getTodos } from './api/todos';
-import { addTodo } from './api/todos';
-import { deleteTodo } from './api/todos';
+import { getTodos, addTodo, deleteTodo } from './api/todos';
+
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Filter } from './types/Filter';
 
 enum ErrorMessage {
   Load = 'Unable to load todos',
   Add = 'Unable to add a todo',
   Delete = 'Unable to delete a todo',
   EmptyTitle = 'Title should not be empty',
-}
-
-enum Filter {
-  All = 'all',
-  Active = 'active',
-  Completed = 'completed',
 }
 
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,323 @@
...
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { addTodo } from './api/todos';
+import { deleteTodo } from './api/todos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = React.useState<Todo[]>([]);
+  const [error, setError] = React.useState<string | null>(null);
+  const [filter, setFilter] = React.useState<'all' | 'active' | 'completed'>(
+    'all',
```
etojeDenys: create an enum for the filter parameters

src/App.tsx
```diff
@@ -1,26 +1,323 @@
...
+  const inputRef = useRef<HTMLInputElement>(null);
+  const loadTodos = async () => {
+    setError(null);
+
+    try {
+      const data = await getTodos();
+
+      setTodos(data);
+    } catch (e) {
+      setError('Unable to load todos');
```

src/App.tsx
```diff
@@ -1,26 +1,323 @@
...
+                href="#/completed"
+                className={`filter__link ${filter === 'completed' ? 'selected' : ''}`}
+                onClick={e => {
+                  e.preventDefault();
+                  setFilter('completed');
+                }}
+                data-cy="FilterLinkCompleted"
+              >
+                Completed
+              </a>
```
etojeDenys: you can create an array with the required parameters and then map through it to create links and avoid code repetition

etojeDenys (changes requested):
1. there is a problem with the server

Image

2. split the app into several components. for example, header, main, footer and errorNotification

etojeDenys (changes requested):
the same problems remain. make sure you deploy your changes

etojeDenys (changes requested):
the same problems. feel free to ask for help if you get stuck on something

AndrewPrszn:
In my last change, I split the app into different files. That seems to have fixed the server issue.

etojeDenys (commented):
server problems are still here
Image

Image

AndrewPrszn:
I fixed the issue with Cypress not catching the request.
The request was triggered too early, so I added a small delay to ensure intercept works correctly.
The app behavior remains unchanged for users.

Anton-Kuchmasov (changes requested):
Take a look onto link that your App is trying to fetch:

Image

Image


Here you should use your unique digit `USER_ID` to get exactly your Todos from Mate API. Not `andrew-todo-app`, but only your unique ID

if you don't have unique ID, you can retreive it [here](https://mate-academy.github.io/react_student-registration/)

etojeDenys (approved):
good job

PR #2193, opened 2026-04-15
Implement add and delete todos functionality - Part 2

Dianaeva:
[DEMO LINK](https://dianaeva.github.io/react_todo-app-add-and-delete/)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,103 @@
...
+      completed: false,
+    });
+
+    client
+      .addTodo(todoTitle.trim())
+      .then(res => {
+        setTempTodo(null);
+        if (appliedFilter !== 'completed') {
+          addTodo(res);
+        }
```
polosanya: In that case the new item won't be saved. Fix it

https://github.com/user-attachments/assets/996cd155-48e8-4875-aab7-c421188bf2e1

src/App.tsx
```diff
@@ -1,26 +1,159 @@
...
+  }, [processTodoData]);
+
+  const visibleTodos = useMemo(() => {
+    switch (appliedFilter) {
+      case 'active':
+        return notCompletedTodos;
+      case 'completed':
+        return completedTodos;
+      case 'all':
+        return todos;
```
polosanya: Use Enum instead of hardcoded values. Also, don't forget about the default case

polosanya (changes requested):

polosanya (approved):

PR #2207, opened 2026-05-03
adding and deleting implemented

YuliaSabitova:
[DEMO LINK](https://YuliaSabitova.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,203 @@
 /* eslint-disable max-len */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+/* eslint-disable jsx-a11y/label-has-associated-control */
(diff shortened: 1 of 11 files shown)


brespect (changes requested):
Good progress, almost done, but for now, I can't select several todos for multiple delete

src/App.tsx
```diff
@@ -1,26 +1,217 @@
...
+    setLoading(false);
+  }
+
+    const clearCompleted = () => {
+    const completedIds = todos.filter(todo => todo.completed);
+
+    completedIds.forEach(todo => {
+      deleteItem(todo.id);
+    });
+  };
```
etojeDenys, line 133: it would be better to use promise.all() here

etojeDenys (approved):
good job

PR #2210, opened 2026-05-05
Solution

aholubko:
[DEMO LINK](https://aholubko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,15 +4,15 @@ import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { addTodo, deleteTodo, getTodos, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
-import classNames from 'classnames';
 
-enum FilterBy {
-  All = 'all',
-  Active = 'active',
-  Completed = 'completed',
-}
+import { FilterBy } from './types/FilterBy';
+import { TodoFooter } from './components/TodoFooter';
+import { ErrorNotification } from './components/ErrorNotification';
+import { TodoHeader } from './components/TodoHeader';
+import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+      });
+  };
+
+  const handleClearCompleted = () => {
+    todos
+      .filter(todo => todo.completed)
+      .forEach(todo => {
+        handleDelete(todo.id);
+      });
+  };
```
etojeDenys, line 112: it would be better to use promise.all() here

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+                  selected: filterBy === FilterBy.Completed,
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={e => {
+                  e.preventDefault();
+                  setFilterBy(FilterBy.Completed);
+                }}
+              >
+                Completed
+              </a>
```
etojeDenys: you can make an array with the necessary parameters and then map through it to create lint and avoid code repetition

etojeDenys (changes requested):
good job, to improve:
1. you should split the app into several components (for example: header, main, footer and error notification)

brespect (changes requested):
Good progress, but check previous comment about Promise.all(), also I can't select several todos to test multiple delete

aholubko:
Thanks! I updated the already submitted files and refactored handleClearCompleted to use Promise.all(). I also made handleDelete return the delete promise so the multiple deletion flow is handled correctly.

Regarding selecting several todos: according to the task split, toggling todos is part 3 (“Toggle and Rename”), so in part 2 I implemented only adding and deleting. Multiple delete can currently be tested with todos that are already completed in the API/demo data. Toggle functionality will be implemented in the next part.

etojeDenys (approved):
great job

PR #2211, opened 2026-05-05
Develop

DmytroYakovchuk:
[DEMO LINK](https://DmytroYakovchuk.github.io/react_todo-app-add-and-delete/)


brespect (changes requested):
Good progress, but for now I can't select several todos to test multiple delete. Also, there is strange behaviour on adding - letters stack to column and after that appear normally, check it.

brespect (changes requested):
Good progress, but you need to pass all CI checks (including linter) before requesting the review

brespect (approved):
Well done!

PR #2213, opened 2026-05-06
completed the task

SerhiyShimko:
[DEMO LINK](https://SerhiyShimko.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/components/Filter.tsx
+++ b/src/components/Filter.tsx
@@ -6,7 +6,7 @@ type Props = {
   todosFormServer: Todo[] | null;
   setTodos: (todos: Todo[]) => void;
   updateList: () => void;
-  deleteTodo: (id: number) => void;
+  deleteTodo: (id: number, todo: Todo) => void;
 };
 
 enum TypeFilter {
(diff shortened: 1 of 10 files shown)


src/components/Filter.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const clearCompleted = useCallback(() => {
+    if (todosFormServer) {
+      todosFormServer.map((todo: Todo) => {
+        if (todo.completed === true) {
+          deleteTodo(todo.id);
+        }
+      });
+    }
+  }, [todosFormServer, deleteTodo]);
```
etojeDenys, line 102: it would be better to use promise.all here

src/components/Filter.tsx
```diff
@@ -0,0 +1,136 @@
...
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={e => {
+            e.preventDefault();
+            setSelectedFilter(TypeFilter.Completed);
+            onlyCompleted();
+          }}
+        >
+          Completed
+        </a>
```
etojeDenys: you can create an array with the required parameters and then map through it to create links and avoid code repetition

etojeDenys (changes requested):
almost done

1. right now, when trying to delete several tasks in a short period of time, the loader only appears on the last task that was pressed instead of on every task that will be deleted

brespect (approved):
Lgtm 🚀

PR #2215, opened 2026-05-07
add task solution

timurradkevic:
- [DEMO LINK](https://timurradkevic.github.io/react_todo-app-add-and-delete/)


src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,31 @@
...
+  errorMessage,
+  setErrorMessage,
+}) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={
+        errorMessage === ''
+          ? 'notification is-danger is-light has-text-weight-normal hidden'
+          : 'notification is-danger is-light has-text-weight-normal'
```
dothearts, line 18: You can use clsx in this case, or add/remove only hidden class

src/components/TodoList.tsx
```diff
@@ -0,0 +1,70 @@
...
+            ×
+          </button>
+
+          {/* overlay will cover the todo while it is being deleted or updated */}
+          <div
+            data-cy="TodoLoader"
+            className={
+              deletingTodos.includes(todo.id)
+                ? 'modal overlay is-active'
+                : 'modal overlay'
```
dothearts, line 58: the same, you can add/remove only is-active or use clsx

DanilWeda (approved):

PR #2224, opened 2026-05-16
add/remove todos

petrushinsky-kirill:
[DEMO LINK](https://petrushinsky-kirill.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,22 +8,23 @@ import { Error } from './components/error';
 import { AppFooter } from './components/app-footer';
 import { ToDo } from './components/todo';
 import { Errors } from './types/errors';
+import { FilterKeys } from './types/filters';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[] | null>(null);
   const [error, setError] = useState<string | null>(null);
   const [isLoading, setIsLoading] = useState(false);
-  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [filter, setFilter] = useState<FilterKeys>(FilterKeys.all);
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,135 @@
...
+import { Errors } from './types/errors';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [error, setError] = useState<string | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
etojeDenys: it would be better to use an enum here

src/App.tsx
```diff
@@ -1,26 +1,135 @@
...
+      })
+      .finally(() => {
+        setLoadingIds(prev => prev.filter(itemId => itemId !== id));
+        setTimeout(() => inputRef.current?.focus(), 0);
+      });
+  };
+
+  const onClearCompleted = () => {
+    todos?.filter(todo => todo.completed).forEach(todo => removeTodo(todo.id));
+  };
```
etojeDenys: you should use the Promise.all() here

etojeDenys (changes requested):
almost done

brespect (changes requested):
Almost done, but for now I can't select several todos for multiple delete:
Image

Anton-Kuchmasov (approved):
Good job, but your deploy was made 2 days ago, so there's still no ability to select todos. 

I'm approving your solution, but pay attention - all required functional should be present in the final chap of this task

PR #2227, opened 2026-05-17
Develop

MykolaBa:
[DEMO LINK](https://MykolaBa.github.io/react_todo-app-add-and-delete/)


src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+      const createdTodo = await postServise.createTodo({
+        userId: USER_ID,
+        title: trimmedTitle,
+        completed: false,
+      });
+
+      setTodos(currentTodos => [...currentTodos, createdTodo]);
+      setTitle('');
+    } catch {
+      setError('Unable to add a todo');
```
Anton-Kuchmasov, line 62: Best approach is to create custom type or enum (same as you made with `FilterStatus` and `Filter`), name it `ErrorMessage.ts` and move all possible error messages here to re-use and avoid hard-coded typos

src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+        todo.id === id ? { ...todo, completed: !todo.completed } : todo,
+      ),
+    );
+  };
+
+  const filters: readonly Filter[] = [
+    { label: 'All', value: FilterStatus.All },
+    { label: 'Active', value: FilterStatus.Active },
+    { label: 'Completed', value: FilterStatus.Completed },
+  ];
```
Anton-Kuchmasov, line 85: you can move it to `/constants` subdir if it's readonly

src/App.tsx
```diff
@@ -1,26 +1,218 @@
...
+  const filteredTodos = todos.filter(todo => {
+    if (filterStatus === FilterStatus.Active) {
+      return !todo.completed;
+    }
+
+    if (filterStatus === FilterStatus.Completed) {
+      return todo.completed;
+    }
+
+    return true;
```
Anton-Kuchmasov, line 96: you can use ternary operator here (or better `switch/case` statement) here

Anton-Kuchmasov (approved):
Good work!

Check out comments below - they can be helpful

PR #2237, opened 2026-05-20
Sol

sergeisvidlo4:
[DEMO](https://sergeisvidlo4.github.io/react_todo-app-add-and-delete/)


DmitriyS2k (approved):
The functionality looks solid overall.

One thing I would improve is component decomposition. Right now App.tsx is getting pretty large, and according to the checklist it could be split into smaller reusable components like TodoItem, TodoList, Footer, Filter, NewTodo, and ErrorNotification. I think you may have forgotten to check checklist.md :)

Also, using classnames for conditional classes could make some JSX parts cleaner.

Other are implemented nicely.

dothearts (approved):

PR #2238, opened 2026-05-20
add task solution

nikassafronovs:


src/App.tsx
```diff
@@ -1,26 +1,199 @@
...
+    }
   }
 
+  const clearCompleted = () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    completedTodos.forEach(todo => {
+      deleteTodo(todo.id);
+    });
+  };
```
etojeDenys: use Promise.all() here

src/App.tsx
```diff
@@ -1,26 +1,199 @@
...
+
+  const loadTodos = () => {
+    setLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```
etojeDenys: create an enum for errors

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,82 @@
...
+            selected: filter === 'completed',
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={e => {
+            e.preventDefault();
+            setFilter('completed');
+          }}
+        >
+          Completed
+        </a>
```
etojeDenys: you can create an array with the required parameters and then map through it to create links and avoid code repetition

etojeDenys (changes requested):
good job, to improve:

1. after trying to remove several todos in a short period of time, the loader only appears on the last one that was pressed

brespect (changes requested):
Good progress, but you need to pass all tests

brespect (changes requested):
Also, you need to pass working DEMO LINk to the PR description

brespect (approved):
Lgtm

PR #2248, opened 2026-05-24
add task solution

chlodav:
https://chlodav.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,454 +1,566 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-/* eslint-disable jsx-a11y/label-has-associated-control */
-import React, { useState, useEffect, useRef, useCallback } from 'react';
-import { flushSync } from 'react-dom';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import classNames from 'classnames';
-import { Todo } from './types/Todo';
-import { getTodos, addTodo, deleteTodo } from './api/todos';
+import { TransitionGroup, CSSTransition } from 'react-transition-group';
+import { Header } from './components/Header';
+import { TodoItem } from './components/TodoItem';
+import { Footer } from './components/Footer';
+import { Todo, FilterType } from './types/Todo';
+import { ErrorMessage } from './enums/ErrorMessage';
+import { UserWarning } from './UserWarning';
 
-type FilterType = 'all' | 'active' | 'completed';
-type User = {
-  id: number;
-  email: string;
-};
+const USER_ID = 1853;
+const API_URL = 'https://mate.academy/students-api/todos';
+const NATIVE_SET_TIMEOUT =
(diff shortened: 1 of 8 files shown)


brespect (changes requested):
Your DEMO LIN is broken. Also, you need to separate your App into smaller components:
Image

src/App.tsx
2pasha, line 1: separate your App to components (header, footer, todo, etc)

src/App.tsx
```diff
@@ -1,26 +1,454 @@
...
+  useEffect(() => {
+    if (!user) {
+      loginInputRef.current?.focus();
+
+      return;
+    }
+
+    getTodos(user.id)
+      .then(setTodos)
+      .catch(() => showError('Unable to load todos'));
```
2pasha: it's better to move all your error messages to `enum`

src/App.tsx
```diff
@@ -1,26 +1,454 @@
...
+              onChange={e => setNewTodoTitle(e.target.value)}
+              onKeyDown={e => {
+                if (e.key === 'Enter') {
+                  const trimmedTitle = e.currentTarget.value.trim();
+
+                  if (trimmedTitle) {
+                    beginAddTodo(trimmedTitle);
+                  }
+                }
+              }}
```
2pasha: move to separate fn

src/App.tsx
```diff
@@ -1,26 +1,454 @@
...
+                className={classNames('filter__link', {
+                  selected: filter === 'completed',
+                })}
+                onClick={e => {
+                  e.preventDefault();
+                  setFilter('completed');
+                }}
+              >
+                Completed
+              </a>
```
2pasha: you may add all your filter values in kind of array and map through it to avoid code duplication

src/App.tsx
```diff
@@ -1,26 +1,454 @@
...
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => {
+            if (errorTimerRef.current) {
+              clearTimeout(errorTimerRef.current);
+            }
+
+            setErrorMessage('');
+          }}
```
2pasha: move to separate fn

src/App.tsx
```diff
@@ -1,26 +1,454 @@
...
+              <a
+                href="#/completed"
+                data-cy="FilterLinkCompleted"
+                className={classNames('filter__link', {
+                  selected: filter === 'completed',
+                })}
+                onClick={e => {
+                  e.preventDefault();
+                  setFilter('completed');
+                }}
```
2pasha: move to separate fn

2pasha (changes requested):
you don't need to implement login logic, your app should just look like [this](https://mate-academy.github.io/react_todo-app-with-api/)

right now, even if i logged in App doesn't work

Image

vadiimvooo (approved):
Well done

PR #2249, opened 2026-05-24
Solution

Palfinger2002:
[DEMO LINK](https://Palfinger2002.github.io/react_todo-app-add-and-delete/)


Evgeniy45 (approved):
Good job! But in file App it`s better to name functions like handleAddTodo or handleRemoveTodo. Think about it
Many thanks)

src/App.tsx
```diff
@@ -1,26 +1,138 @@
...
+        setErrorMessage(Errors.UnableToLoad);
+        setTimeout(() => setErrorMessage(null), 3000);
+      });
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const visibleTodos = todos?.filter(todo => {
```
dothearts, line 79: You can move it as pure function in separate file (helpers.ts for example)

src/App.tsx
```diff
@@ -1,26 +1,138 @@
...
+    }
+  });
+
+  const countActiveTodos = todos?.filter(
+    todo => todo.completed === false,
+  ).length;
+
+  const countCompletedTodos = todos?.filter(
+    todo => todo.completed === true,
+  ).length;
```
dothearts, line 98: this also can be helper funct (not critical)

dothearts (approved):

PR #2255, opened 2026-05-28
add task solution

Ivan7457:


src/App.tsx
```diff
@@ -1,26 +1,370 @@
...
+          });
+      }
+    });
+  };
+
+  const clearCompleted = () => {
+    const completedTodos = todos.filter(tod => tod.completed);
+
+    completedTodos.forEach(to => deleteTodo(to.id));
+  };
```
etojeDenys, line 207: you should use the 'Promise.all' here

src/App.tsx
```diff
@@ -1,26 +1,370 @@
...
+  useEffect(() => {
+    setErrorMessage(null);
+
+    postService
+      .getTodos()
+      .then(loadedTodos => {
+        setTodos(loadedTodos);
+      })
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```

etojeDenys (changes requested):
good job, but there are two things to improve

brespect (changes requested):
You need to pass your DEMO LINK to the PR description

brespect (approved):
Well done!

PR #2257, opened 2026-05-30
todo-add-and-delete1

TaniaP26:


vadiimvooo (approved):
Well done. Consider to split your app to smaller components in the next part. (like Footer, Header, TodoItem, Todolist, Error etc)

PR #2259, opened 2026-05-31
second part of todo app

Dazavrrr:
- [DEMO LINK](https://Dazavrrr.github.io/react_todo-app-add-and-delete/)


Tysiachnyk (approved):
Hello! 
Everything works well from my side 👍

I ran the code through an AI review tool to get an additional perspective. These suggestions looked reasonable and might be worth checking:

_src/api/todos.ts_
In the add method, userID is sent instead of userId. It would be worth verifying this against the API contract to make sure the field name matches what the server expects.

_src/App.tsx_
handleDeleteTodo calls handleToggleTodoLoading both before the request and again in the finally block. Since the function toggles state, this may lead to less predictable behavior. Using dedicated helpers such as addLoadingId / removeLoadingId would make the logic more explicit and easier to follow.

The useEffect responsible for loading todos contains setTimeout(() => setErrorMessage(''), 3000), but similar error-reset logic already appears to be handled inside useErrorMessage. This looks like duplicated functionality and could potentially be removed.

_src/components/TodoItem.tsx_
key={todo.id} on the root element inside the component has no effect. React only uses key when rendering a list in the parent component (map), so this prop can be removed from TodoItem.

Overall, the solution looks good. 👍

src/App.tsx
```diff
@@ -1,26 +1,255 @@
...
+        setErrorMessage(getTodoError(TodosServiceError.UnableToLoadTodos));
+        setTimeout(() => setErrorMessage(''), 3000);
+      });
+  }, [setErrorMessage]);
+
+  const showTodosAndFooter = todos.length > 0;
+  const showToggleAllButton = todos.length > 0;
+  const hasCompletedTodos = todos.some(todo => todo.completed);
+
+  const filteredTodos = todos.filter(todo => {
```
dothearts, line 127: You can create pure function for it, and move it from component in separate file (helpers.ts for example)

dothearts (approved):

PR #2262, opened 2026-06-01
second part

YYarik:
https://YYarik.github.io/react_todo-app-add-and-delete/


src/components/Filter.tsx
```diff
@@ -0,0 +1,63 @@
...
+  selectedFilter: FilterType;
+  onFilterChange: (filter: FilterType) => void;
+}
+
+export const Filter: React.FC<FilterProps> = ({
+  selectedFilter,
+  onFilterChange,
+}) => {
+  const handleFilterClick = (
+    e: React.MouseEvent<HTMLAnchorElement>,
```
Viqsol, line 20: I think we shouldn't use single-letter names. You can use event next time)

Viqsol (approved):
Great job! My congrats! 
Тext time you can even remove the comments - the code will be cleaner in App.tsx)

dothearts (approved):
Better fix one letter naming. 

Also, don't forget about DEMO LINK in README.md file.

Many thanks!

PR #2263, opened 2026-06-02
task 2

lilp3ka:
https://lilp3ka.github.io/react_todo-app-add-and-delete/


vaynexx (changes requested):
React Hook "useEffect" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks

PR #2273, opened 2026-06-13
Develop

AntonNimachuk:
https://antonnimachuk.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,6 +7,7 @@ import { NewTodoForm } from './components/NewTodoForm';
 import { TodoList } from './components/TodoList';
 import { ErrorType } from './types/ErrorType';
 import { FilterType } from './types/FilterType';
+import { TodoItem } from './components/TodoItem';
 
 export const App: React.FC = () => {
   if (!clientMethods.USER_ID) {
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,26 +1,195 @@
...
+      setError(ErrorType.Delete);
+    } finally {
+      setDeletingIds(current => current.filter(id => id !== deleteId));
+      inputFocusRef.current?.focus();
+    }
+  };
+
+  const handleClearCompleted = () => {
+    completedTodos.forEach(todo => handleDelete(todo.id));
+  };
```
etojeDenys, line 65: it would be better to use Promise.all() here

src/App.tsx
```diff
@@ -1,26 +1,195 @@
...
+              Clear completed
+            </button>
+          </footer>
+        )}
+      </div>
+
+      <div
+        data-cy="ErrorNotification"
+        className={`notification is-danger is-light has-text-weight-normal ${!error ? `hidden` : ''}`}
+      >
```
etojeDenys: you can use the class name library here

etojeDenys (changes requested):
almost done

etojeDenys (approved):
lgtm

PR #2279, opened 2026-06-26
feat: adding and deletints feats are added

TarasPidkuimukha:
[DEMO LINK](https://TarasPidkuimukha.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,8 +4,12 @@ import React, { useState, useEffect, useRef } from 'react';
 import { UserWarning } from './UserWarning';
 import { USER_ID, deleteTodo, getTodos, postTodos } from './api/todos';
 import { Todo, TodoWithoutId } from './types/Todo';
-import classNames from 'classnames';
 import { TodoFilter } from './enums/TodoFilter.enum';
+import { ErrorMessages } from './enums/ErrorMessages.enum';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { TempTods } from './components/TempTods';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
   const [title, setTitle] = useState('');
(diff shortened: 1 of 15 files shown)


brespect (changes requested):
Good progress, but you need pass linter before requesting the review

src/App.tsx
```diff
@@ -1,26 +1,298 @@
...
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
 
 export const App: React.FC = () => {
+  const [title, setTitle] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterStatus, setFilterStatus] = useState<
+    'all' | 'active' | 'completed'
+  >('all');
```
etojeDenys: create an enum for the filter parameters

src/App.tsx
```diff
@@ -1,26 +1,298 @@
...
+  // const [isLoading, setIsLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const addTodos = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    const trimmed = title.trim();
+
+    if (trimmed === '') {
+      setErrorMessage('Title should not be empty');
```

src/App.tsx
```diff
@@ -1,26 +1,298 @@
...
+                    selected: filterStatus === 'completed',
+                  })}
+                  data-cy="FilterLinkCompleted"
+                  onClick={event => {
+                    event.preventDefault();
+                    setFilterStatus('completed');
+                  }}
+                >
+                  Completed
+                </a>
```
etojeDenys: you can create an array with the necessary parameters and then map through it to create links and avoid code repetition

etojeDenys (changes requested):
good job, to improve:

1. this part is redundant
Image

2. it would be better to split the app into several components

etojeDenys (changes requested):
there is still space in the footer
Image

etojeDenys (changes requested):
there is still a problem with the margin for error. here is an example of what it should look like:

Image

feel free to ask for help in the chat if you get stuck

etojeDenys (approved):
good job

PR #2296, opened 2026-07-16
add task solution

nkoropka:
- [DEMO LINK](https://<your_account>.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,9 +2,14 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { getTodos, USER_ID, createTodo, deleteTodo } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
-import classNames from 'classnames';
+import { useAddTodo } from './hooks/useAddTodo';
+import { useDeleteTodo } from './hooks/useDeleteTodo';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
 
 type FilterType = 'all' | 'active' | 'completed';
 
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+                      has-background-white-ter"
+                      />
+                      <div className="loader" />
+                    </div>
+                  </div>
+                );
+              })}
+
+              {tempTodo && (
+                <div
```
Zoolgrand: This condition now ignores filtering, so if you add a new todo while on the completed tab, you will see the new element flash.
You can add this condition
`{tempTodo && filter !== 'completed' && (`

Zoolgrand (changes requested):
Looks good overall, but requires a useEffect dependency fix and a minor fix for filtering.

src/App.tsx
```diff
@@ -1,26 +1,336 @@
...
+    }, 3000);
+
+    return () => clearTimeout(timer);
+  }, [errorMessage]);
+
+  useEffect(() => {
+    if (newTodoFieldRef.current) {
+      newTodoFieldRef.current.focus();
+    }
+  });
```
Zoolgrand: Missing dependancy array, this will trigger after each rerender

vitalii-bondarenko-ua (changes requested):
Good job.

Split App component into separate smaller components (Header, Footer, TodoList, TodoItem, ErrorNotification)

src/App.tsx
```diff
@@ -1,26 +1,104 @@
...
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState<FilterType>('all');
+
+  const newTodoFieldRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```
etojeDenys, line 27: create an enum for errors

src/components/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: filter === 'completed',
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => onFilterChange('completed')}
+        >
+          Completed
+        </a>
```
etojeDenys, line 59: you can create an array with the required parameters and then map through it to create links and avoid code repetition

etojeDenys (approved):
good job, but please fix the comment before you start the next part

PR #2299, opened 2026-07-20
add task solution

A-Melnykk:
[DEMO LINK](https://A-Melnykk.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,6 +5,10 @@ import { UserWarning } from './UserWarning';
 import { getTodos, createTodo, deleteTodo, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
 import { ErrorMessage } from './types/ErrorMessage';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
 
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,26 +1,298 @@
...
+      .finally(() => {
+        setLoadingTodoIds(prevIds => prevIds.filter(id => id !== todoId));
+      });
+  };
+
+  const handleClearCompleted = () => {
+    const completedTodos = todos.filter(todo => todo.completed);
+
+    completedTodos.forEach(todo => handleDeleteTodo(todo.id));
+  };
```
etojeDenys, line 74: you have to use Promise.all() here

etojeDenys (changes requested):
visual part looks broken

Image

also split the app into several components

roman-mirzoian (changes requested):
Good job! 
Let's also add loaders to display while we're waiting for todos to be added or removed:
1. temp TodoItem should have the loader (check the original markup);
2. covered the todo with the loader while waiting for the API response;

There is the example of result loading:
https://gyazo.com/9a89923195a4a2edf01bcaad5324d432

Anton-Kuchmasov (approved):
Good job!

Let's proceed with final chap of this task

PR #2308, opened 2026-07-28
react-todo-app-add-and-delete 1.0

FRANCISCOPELEGRINI:
https://franciscopelegrini.github.io/react_todo-app-add-and-delete/


src/App.tsx
```diff
@@ -1,26 +1,410 @@
-/* eslint-disable max-len */
+/* eslint-disable jsx-a11y/label-has-associated-control */
```
DanielCamposDevX, line 1: dont use lint disablers on tasks you are suposed to remove then

cypress/integration/page.spec.js
```diff
@@ -519,7 +519,7 @@ describe('', () => {
         });
 
         // this test may be flaky
-        it.skip('should replace loader with a created todo', () => {
```
DanielCamposDevX, line 522: You cannot change test files. On github the original test is not passing.

DanielCamposDevX (changes requested):
Do not change teste flies.
Remember to not leave the whole code on App.tsx create components and utils for clean code.
While adding the todo after the loading happens the loading todo disapears and then the new todo appears, the todo cannot dissapear and appear, the transition between load and creation should be smooth and the todo cannot dissapear and appear.
## No demo link was provided i used the default deploy link so i can provide this suggested changes (remember to place demo link on the description of the pull request)

DanielCamposDevX (changes requested):
no changes were applied since the last review.
Only request a new review when changes are applied

DanielCamposDevX (approved):

PR #2318, opened 2026-08-08
Add Solution

CodableBase:


src/App.tsx
```diff
@@ -1,26 +1,424 @@
...
+
+        if (isMounted) {
+          setTodos(loadedTodos);
+          setError(null);
+          setShouldFocusInput(true);
+        }
+      } catch {
+        if (isMounted) {
+          setTodos([]);
+          setError('Unable to load todos');
```
etojeDenys, line 134: create an enum for errors

src/App.tsx
```diff
@@ -1,26 +1,424 @@
...
+                  type="button"
+                  className={classNames('todoapp__filter', {
+                    selected: filter === 'completed',
+                  })}
+                  data-cy="FilterLinkCompleted"
+                  onClick={() => setFilter('completed')}
+                >
+                  Completed
+                </button>
+              </div>
```
etojeDenys, line 392: you can create an array with the necessary parameters and then map through it to create links and avoid code repetition

src/App.tsx
```diff
@@ -1,26 +1,424 @@
...
 
-const USER_ID = 0;
+type Todo = {
+  id: number;
+  userId: number;
+  title: string;
+  completed: boolean;
+};
+
+type FilterType = 'all' | 'active' | 'completed';
```
etojeDenys, line 14: it would be better to use an enum here instead of a type

etojeDenys (changes requested):
almsot done, to improve:

1. it is bad practice to have several components in one file

vadiimvooo (changes requested):
Looks like there are no changes since last review. Please feel free to ask for help in fe_chat.

brespect (changes requested):
Good progress, but you need to share your working DEMO LINK to the PR description

PR #2319, opened 2026-08-10
ready

MarianaKlochnyk:
https://MarianaKlochnyk.github.io/react_todo-app-add-and-delete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -31,7 +31,7 @@ export const App: React.FC = () => {
   const [error, setError] = useState('');
   const [filter, setFilter] = useState(Filter.All);
   const [title, setTitle] = useState('');
-  const [deletingTodoId, setDeletingTodoId] = useState<number | null>(null);
+  const [deletingTodoIds, setDeletingTodoIds] = useState<number[]>([]);
   const [tempTodo, setTempTodo] = useState<Todo | null>(null);
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [isClearing, setIsClearing] = useState(false);
@@ -105,7 +105,7 @@ export const App: React.FC = () => {
   }, [tempTodo]);
 
   const removeTodo = (todoId: number) => {
-    setDeletingTodoId(todoId);
+    setDeletingTodoIds(currentIds => [...currentIds, todoId]);
 
     deleteTodo(todoId)
       .then(() => {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,26 +1,372 @@
...
+  const [title, setTitle] = useState('');
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const [deletingTodoId, setDeletingTodoId] = useState<number | null>(null);
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  // eslint-disable-next-line react-hooks/rules-of-hooks, @typescript-eslint/no-unused-vars
+  const [isClearing, setIsClearing] = useState(false);
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const inputRef = useRef<HTMLInputElement>(null);
```
roman-mirzoian: Move the check for the existence of `USER_ID` to after all the hooks have been declared and then you won't need to disable the linter on every line

roman-mirzoian (changes requested):
Great start!
I have a small comment for improve below

src/App.tsx
```diff
@@ -1,26 +1,365 @@
...
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  };
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  useEffect(() => {
+    loadTodos();
+  }, []);
```
2pasha, line 99: check linter error

src/App.tsx
```diff
@@ -1,26 +1,365 @@
...
+      });
+  };
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    const trimmedTitle = title.trim();
+
+    if (!trimmedTitle) {
+      showError('Title should not be empty');
```
2pasha: move all error messages to enum

2pasha (changes requested):
good job! 👏

take a look on comments below ⬇️

and make sure that you have pushed all your files

src/App.tsx
2pasha: error is still here

Image

src/App.tsx
2pasha: add this error to your `enum ErrorMessage`

2pasha (changes requested):
comments from previous review are still actual

etojeDenys (changes requested):
almost done
1. after trying to delete several todos, the loader only appears on the last todo that was pressed, instead of on all of them

roman-mirzoian (approved):
Well done 👍

PR #2321, opened 2026-08-12
12/08

goshulyakvitalik-sudo:
[DEMO LINK](https://goshulyakvitalik-sudo.github.io/react_todo-app-add-and-delete/)


brespect (changes requested):
Good progress, but you need to pass all CI checks (incl. tests & linter) before requesting mentor's review

vitalii-bondarenko-ua (changes requested):
Compare your work with [working example](https://mate-academy.github.io/react_todo-app-with-api/).
For example button `Clear completed` should be visible in case if exists some finished tasks

Image
Image


And you need to pass all linter and test checks

etojeDenys (approved):
lgtm

PR #2325, opened 2026-08-19
Solution

Vadym-fe:
[DEMO LINK](https://Vadym-fe.github.io/react_todo-app-add-and-delete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,280 @@
-/* eslint-disable max-len */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import { UserWarning } from './UserWarning';
+import {
+  USER_ID,
+  addTodo,
+  deleteTodo,
+  getTodos,
+  updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Selected } from './types/Selected';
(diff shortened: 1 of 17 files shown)


brespect (changes requested):
Good progress, but logic seems to be broken: when I select one todo it select all available that is not completed instead, after that when I try to delete only one it will remove everything presented

src/components/Header.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const newTitle = title.trim();
+
+  useEffect(() => {
+    if (blockedInput) {
+      return;
+    }
+
+    focusElement.current?.focus();
+  }, [blockedInput]);
```
Zoolgrand, line 30: Please extract logic into custom hook

src/App.tsx
```diff
@@ -1,26 +1,280 @@
...
+    }, 3000);
+
+    return () => {
+      if (timerError.current === null) {
+        return;
+      }
+
+      clearTimeout(timerError.current);
+    };
+  }, [error]);
```
Zoolgrand: Please extract this logic into custom hook

Zoolgrand (changes requested):
You have fixed functionality and code looks good overall, I'd suggest to make small refactoring and cleane the code comments across project

etojeDenys (changes requested):
almost done
1. after trying to remove several todos in a short period of time, the loader only appears on the last  that was pressed

2. make sure that the loader appears on the todos that will be deleted after you press 'clear completed'

roman-mirzoian (approved):
Well done 👍

